PR #13, opened 2018-08-19
various fixes

FFroehlich:


pesto/optimize/optimize.py
```diff
@@ -50,8 +50,11 @@ def minimize(problem, optimizer,
     # do multistart optimization
     for j in range(0, n_starts):
         startpoint = startpoints[j, :]
-        optimizer_result = optimizer.minimize(problem, startpoint)
-        result.append_optimizer_result(optimizer_result=optimizer_result)
+        try:
+            optimizer_result = optimizer.minimize(problem, startpoint)
+            result.append_optimizer_result(optimizer_result=optimizer_result)
+        except Exception as err:
```
yannikschaelte: what errors would one expect here? ok for the moment, but just printing looks not like the best option.

FFroehlich: none of the optimizers that I have tried so far exit gracefully when amici fails. `ls_dogbox` for example fails with "SVD did not converge in Linear Least Squares".

yannikschaelte: you mean amici fails to integrate the problem (but does not throw an error itself, right?), and then the optimizer throws an error?

FFroehlich: yes

PR #26, opened 2018-08-23
Fixes and qol features

FFroehlich:


pypesto/objective.py
```diff
@@ -17,8 +17,12 @@
 except ImportError:
     amici = None
     pass
-    # we only pass import errors, if the initialization of the module itself
-    # fails this should not be caught right?
+
+try:
+    import pandas as pd
```
yannikschaelte: I would rather put pandas in setup.py -> install_requires (might have missed that), and avoid try here, since it will be needed in many places anyway

pypesto/objective.py
```diff
@@ -188,38 +213,195 @@ def call_mode_res(self, x, sensi_orders):
...
+        res = self.call_mode_res(x, (0,))
         return res
 
     def get_sres(self, x):
-        sres = self.__call__(x, (1,), Objective.MODE_RES)
+        sres = self.call_mode_res(x, (1,))
         return sres
 
+    """
+    Gradient checker function
```
yannikschaelte: could you describe in more detail what the function does?

pypesto/objective.py
```diff
@@ -81,6 +80,18 @@ def __init__(self, fun,
         self.res = res
         self.sres = sres
 
+        self.dim = dim
```
yannikschaelte: we had discussed with Jan to have dim only in the problem description, because sometimes an objective can be called with different argument lengths, and because we have this whole fixed parameters stuff. i am currently also working on sth related, and introduced in Problem a full_dim and a dim field which can be filled in diverse ways.
would there be a reason to have dim in objective with amici?

pypesto/objective.py
```diff
@@ -188,38 +213,195 @@ def call_mode_res(self, x, sensi_orders):
...
+                + parameter_names
+            self.trace = pd.DataFrame(columns=cols)
+            self.temp_file = temp_file
+            self.temp_save_iter = temp_save_iter
+        else:
+            self.trace = None
+            self.temp_file = None
+            self.temp_save_iter = None
+
+    def update_eval_counts(self, sensi_orders):
```
yannikschaelte: nice solution. i was wondering whether it would be possible to move this "counting stuff" to a wrapper class, so as to disentangle the objective and gradient computation from the counting part. the wrapping could then occur before every local optimization, also helping avoid parallelization and history issues.
but that can still be done later.

FFroehlich: Sure, we could push all the newly introduced fields to a different class but I do not see a way around attaching that class to the objective. But maybe there is some python magic that I do not know of.

Regarding parallelization, I think really the best way to do this is have pypesto being started from separate jobs. This is the only way to effectively parallelize across nodes on a cluster and also works on a local machine.
It would not work with cooperative scatter search strategies, but there we probably want to introduce an additional level of hierarchy anyways.

FFroehlich:
> what do you mean by separate jobs?

submitting individual batch jobs for every multistart. e.g. using snakemake:

```
rule parameter_estimation:
	input:
		model=rules.build_amici.output,
		data=rules.process_data.output,
		script='parameter_estimation.py',
	output:
		os.path.join('results', 'multistart', '{modelname}', 'result_{dataset}_{repeat}.pickle')
	wildcard_constraints:
		modelname='[\w\d_]+',
		dataset='[\w\d]+',
		repeat='\d+'
	shell:
		expand('python3 {{input.script}} {{input.model}} {{wildcards.dataset}} {{wildcards.repeat}} {n_starts}', n_starts=MS_PER_JOB)[0]
```

pypesto/objective.py
yannikschaelte: Jan also mentioned that we would like to have transparent parallelization engines in optimization and sampling à-la pyabc using e.g. redis-servers, much of which one might be able to re-use. in optimization first of all I guess only starts would be parallelized, but also startpoint selection.

FFroehlich: > what do you mean by separate jobs?

submitting individual batch jobs for every individual (subbatch of) optimization(s). e.g. using snakemake:

```
rule parameter_estimation:
	input:
		model=rules.build_amici.output,
		data=rules.process_data.output,
		script='parameter_estimation.py',
	output:
		os.path.join('results', 'multistart', '{modelname}', 'result_{dataset}_{repeat}.pickle')
	wildcard_constraints:
		modelname='[\w\d_]+',
		dataset='[\w\d]+',
		repeat='\d+'
	shell:
		expand('python3 {{input.script}} {{input.model}} {{wildcards.dataset}} {{wildcards.repeat}} {n_starts}', n_starts=MS_PER_JOB)[0]
```

This should fit with the use of redis servers.

PR #30, opened 2018-08-29
Feature plotwaterfall

yannikschaelte:
first functioning version of waterfall plot


pypesto/visualize/optimize.py
```diff
@@ -0,0 +1,70 @@
...
+    # axes
+    if ax is None:
+        fig, ax = plt.subplots()
+
+    # extract cost function values from result
+    result_fval = result.optimize_result.get_for_key('fval')
+    result_fval = np.reshape(result_fval, [len(result_fval), 1])
+    startpoints = range(1, len(result_fval) + 1)
+
+    # cluster
```
dweindl: As discussed with @DantongWang we want to have two plotting layers here, one taking only the relevant data as "basic" types and doing the actual plotting, and another one taking `pypesto.Result` extracting the relevant data and passing it to the former.

pypesto/visualize/optimize.py
```diff
@@ -0,0 +1,70 @@
...
+    clust = cluster.hierarchy.fcluster(
```
dweindl: Move clustering to separate function for readability and reuse. This will be used for parameter plots as well.

pypesto/visualize/optimize.py
```diff
@@ -0,0 +1,70 @@
...
+        cluster.hierarchy.linkage(result_fval),
+        0.1, criterion='distance')
+    uclust, ind_clust = np.unique(clust, return_inverse=True)
+    clustsize = np.zeros(len(uclust))
+    for iclustsize in range(len(uclust)):
+        clustsize[iclustsize] = sum(clust == uclust[iclustsize])
+
+    # assign colors
+    jet = plt.get_cmap('jet')
```
dweindl: Don't select color maps, keep styling at the bare minimum. The user should take care of that in a centralized way. We can provide a default style though (https://matplotlib.org/users/customizing.html).

pypesto/visualize/optimize.py
```diff
@@ -0,0 +1,70 @@
...
+        if clustsize[ind_clust[iind]] == 1:
+            ax.plot(startpoints[iind], result_fval[iind],
+                    color='lightgrey', marker='o')
+            sum_grey = sum_grey
+        else:
+            Col = scalarMap.to_rgba(uclust[ind_clust[iind]] - sum_grey)
+            ax.plot(startpoints[iind], result_fval[iind],
+                    color=Col, marker='o')
+
+    ax.set_xlabel('Ordered multi-starts')
```
dweindl: Remove "multi"
"Ordered optimizer run"?

pypesto/visualize/optimize.py
dweindl: Consider adding a test for the clustering, including vectors containing NaN and Inf.

pypesto/visualize/optimize.py
dweindl: Not sure about the ideal naming and location, but something like:
```
def waterfall_lowlevel(optimizerResults=[1,2,3], exitCode=None, ax=None,....):
         ...
def waterfall(result=pypesto.Result,...):
    return waterfall_lowlevel(result.finalLlh, result.exitCode, ...)
```

DantongWang: I try to do it with two files, 'pypesto.visualize.extract_results' and '.waterfall'. The first one currently includes two functions, 'extract_fval' and 'extract_x', to get values needed for both waterfall plot and parameters plot. The second file contains only the 'waterfall' function, which takes a list of fval as input, i.e. it can be used for any list of numbers. How do you think?

yannikschaelte: if i understand correctly, this does not solve the problem that we want a waterfall() function that can take both a list and a pypesto.Result object.

yannikschaelte: i think daniel's solution is good, since usually the user will call waterfall then, so the not-so-nice name waterfall_lowlevel is usually not used. then we would always have _lowlevel functions that take  numpy arrays instead of result objects?

dweindl: @DantongWang 

>The first one currently includes two functions, 'extract_fval' and 'extract_x', to get values needed for both waterfall plot and parameters plot. 

This functionality should be inside the function taking the pypesto.Result, and passing the on to to the function below.

> The second file contains only the 'waterfall' function, which takes a list of fval as input, i.e. it can be used for any list of numbers.

:+1:

dweindl: @yannikschaelte 
> then we would always have _lowlevel functions that take numpy arrays instead of result objects?

sounds good. feel free to change the naming though. we can also put them inside pypesto.visualize.lowlevel, so they can be imported with nicer names more easily.

pypesto/visualize/clust_color.py
```diff
@@ -0,0 +1,69 @@
...
+
+        Parameters
+        ----------
+
+        result_fval: list of cost function values
+
+        Returns
+        -------
+
+        clust: clusters of cost function values
```
dweindl: what exactly? nested list?

pypesto/visualize/clust_color.py
```diff
@@ -0,0 +1,69 @@
...
+    for iclustsize, value_uclust in enumerate(uclust):
+        clustsize[iclustsize] = sum(clust == value_uclust)
+
+    return clust, clustsize, ind_clust
+
+
+def assigncolor(result_fval):
+
+    """
+        Assign color to each cluster
```
dweindl: Reduce indentation

pypesto/visualize/waterfall.py
```diff
@@ -0,0 +1,74 @@
...
+
+def waterfall_lowlevel(result_fval, ax=None):
+
+    """
+    Plot waterfall plot using list of cost function values.
+
+    Parameters
+    ----------
+
+    result_fval: list of cost function values
```
dweindl: We assume here that they are ordered already, right? Should be mentioned or just sort them here.

pypesto/visualize/waterfall.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+
+    # axes
+    if ax is None:
+        fig, ax = plt.subplots()
+
+    # extract cost function values from result
+
+    result_fval = np.reshape(result_fval, [len(result_fval), 1])
+    startpoints = range(1, len(result_fval) + 1)
```
dweindl: not really "startpoints" but rather optimization_index or something

pypesto/visualize/waterfall.py
yannikschaelte: valid point. currently, the multistart results are always automatically sorted after the optimization. i think we should keep doing that, but maybe remember in the result object the index of the optimizer run.

dweindl: Not sure I would appreciate the automated sorting in cases where i provide starting points myself.

DantongWang: maybe describe like "numeric list, ordered by 'sorted'"?

yannikschaelte: true, so maybe i delete sorting by default. in the result extraction functions we can maybe add a flag sorted=True, which then does the sorting on a copy list on-the-fly.

yannikschaelte: I mean not now but later, so important for here @DantongWang is just that you do not assume the list to be sorted, and do not do sorting yourself, right?

dweindl: @yannikschaelte sounds good

@DantongWang I think I'd prefer just sorting them

yannikschaelte: @dweindl so what do you mean with "not sure ... see how trajectories diverge", then you would want the sorting routines also to not automatically sort but present the results in the given order?

dweindl: well, sorting routines should sort.. :)

By default I would expect to obtain the optimization results in the order of provided/drawn starting points (In case of parallelization this might become non-trivial for the latter). For convenience, there should additionally be support for sorting by final objective (as implemented, but this should not be the default in my opinion).

DantongWang: @yannikschaelte by "do not do sorting yourself" you mean that the list is sorted in the extraction function? And a question: would it still be called a "waterfall plot" if plotted without sorting the cost function values?

dweindl: "do not assume the list to be sorted, and ~do not~ do sorting yourself", i think ;)

yannikschaelte: ups, meant "visualization routines should not automatically sort"

yannikschaelte: but on second thoughts, an unsorted list probably makes not much sense to visualize, so maybe better sort the list.

DantongWang: sorry, I did not get it so well. So the visualization routine does not 'automatically' sort the list, but the list can be sorted 'manually' by a flag?

dweindl: > ups, meant "visualization routines should not automatically sort"

since the function is called "waterfall" and not "scatter", i think it should sort automatically.

DantongWang: @dweindl Agreed, so I think we can simply move the sorting routine from 'optimization' to 'plotwaterfall'.

yannikschaelte: concerning plotting: should work now (it's sorted in waterfall), but I want to raise awareness that one has to make sure the same coloring is applied in the diverse plotting functions).

atm, in optimizer(), the results are still sorted, but that will be dealt with separately.

yannikschaelte:
@dweindl @DantongWang wanna have another look?

pypesto/visualize/clust_color.py
```diff
@@ -0,0 +1,73 @@
+from scipy import cluster
+import matplotlib.colors as colors
+import matplotlib.cm as cm
+import numpy as np
+
+
+def get_cluster(result_fval):
```
DantongWang: I found that 'result_fval' is changed to 'fvals' later in the 'parameters' function, so maybe also need to be changed here?

pypesto/visualize/clust_color.py
```diff
@@ -0,0 +1,73 @@
...
+        0.1, criterion='distance')
+    uclust, ind_clust = np.unique(clust, return_inverse=True)
+    clustsize = np.zeros(len(uclust))
+    for iclustsize, value_uclust in enumerate(uclust):
+        clustsize[iclustsize] = sum(clust == value_uclust)
+
+    return clust, clustsize, ind_clust
+
+
+def assign_color(result_fval):
```
DantongWang: result_fval->fvals?

pypesto/visualize/parameters.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Returns
+    -------
+
+    ax: matplotlib.Axes
+        The plot axes.
+    """
+
+    result_fval = result.optimize_result.get_for_key('fval')
+    result_x = result.optimize_result.get_for_key('x')
```
DantongWang: result_fval->fvals, result_x->xs

.travis.yml
```diff
@@ -26,7 +26,10 @@ install:
   - pip3 install --upgrade setuptools wheel pkgconfig
   - pip3 install --upgrade -r ./.travis_pip_reqs.txt
   - pip3 install .
-
+before_script:
```
dweindl: Might be helpful to add a comment why xvfb is required

pypesto/visualize/clust_color.py
dweindl: I'd simply go for `x`. We don't need any interpretation here.

pypesto/visualize/parameters.py
```diff
@@ -0,0 +1,90 @@
...
+    lb = result.problem.lb
+    ub = result.problem.ub
+
+    return parameters_lowlevel(result_x, result_fval, lb, ub, ax,)
+
+
+def parameters_lowlevel(xs, fvals, lb=None, ub=None, ax=None):
+
+    """
+    Plot waterfall plot using list of cost function values.
```
dweindl: no waterfall here

pypesto/visualize/parameters.py
```diff
@@ -0,0 +1,90 @@
...
+
+    result_fval = result.optimize_result.get_for_key('fval')
+    result_x = result.optimize_result.get_for_key('x')
+    lb = result.problem.lb
+    ub = result.problem.ub
+
+    return parameters_lowlevel(result_x, result_fval, lb, ub, ax,)
+
+
+def parameters_lowlevel(xs, fvals, lb=None, ub=None, ax=None):
```
dweindl: Should (later?) allow for passing parameter names as axis labels

test/test_visualize.py
```diff
@@ -0,0 +1,65 @@
...
+    fvals = np.array(fvals)
+    pypesto.visualize.waterfall_lowlevel(fvals)
+
+
+def test_cluster():
+    fvals = [0.01, 0.02, 1.01, 2.02, 2.03, 2.04, 3, 4, 4.1, 4.11]
+    pypesto.visualize.get_cluster(fvals)
+    fvals = np.array(fvals)
+    pypesto.visualize.get_cluster(fvals)
+    # TODO: evaluate calculated clusters
```
dweindl: TODO

pypesto/visualize/parameters.py
```diff
@@ -0,0 +1,90 @@
...
+    colors = assign_color(fvals)
+
+    # parameter indices
+    parameters_ind = range(1, len(xs[0]) + 1)
+
+    # plot parameters
+    ax.xaxis.set_major_locator(MaxNLocator(integer=True))
+    for j_x, x in reversed(list(enumerate(xs))):
+        ax.plot(x, parameters_ind, color=colors[j_x], marker='o')
+
```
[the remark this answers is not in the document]
yannikschaelte: that is so that the "best" parameter is plotted last and therefore on top of the others. this way it will be better visible

yannikschaelte:
@dweindl : What do you mean by "In either example there are two colored clusters. How come they get different colors assigned?" sounds like something that should not happen.

PR #37, opened 2018-08-31
Feature fixedpars

yannikschaelte:
Enable fixing of parameters:
* In pypesto.Problem, fixed parameters can be specified via x_fixed_indices and x_fixed_vals. This information is then propagated via the callback method objective.handle_x_fixed() to the pypesto.Objective (which is copied) which can then take reduced vectors and internally map them back to the full input vector including fixed dimensions.
* In this way, the Optimizer only sees the reduced vectors, which is desirable.
* After the optimization, in the results object all vectors are mapped back to include fixed dimensions, using comfort function in pypesto.Problem.
* The AmiciObjective one might (later) want to change to use amici's ability to fix parameters, this has not been done yet because the AmiciObjective will undergo diverse changes anyway.
* Question: E.g. problem.lb, ub are maintained in the reduced dimension. One might want to do this differently. Atm, if one want the full vectors, one would need to call problem.get_full_vector(problem.lb, problem.x_fixed_vals)


dweindl:
I'd prefer if somebody who was using this feature in matlab before could review this. Who?

doc/contribute.rst
```diff
@@ -21,5 +21,20 @@ Tests can be written with `pytest <https://docs.pytest.org/en/latest/>`_
 or the `unittest <https://docs.python.org/3/library/unittest.html>`_ module.
```
dweindl: Is contribute.rst recognized by github? Otherwise change to CONTRIBUTING.md to [show](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) this information in pull requests?

pypesto/problem.py
```diff
@@ -17,27 +18,194 @@ class Problem:
...
+
+        self.dim_full = dim_full if dim_full is not None else self.lb.size
+
+        if x_fixed_indices is None:
+            x_fixed_indices = np.array([])
+        self.x_fixed_indices = np.array(x_fixed_indices, dtype=int)
+
+        if x_fixed_vals is None:
+            x_fixed_vals = np.array([])
+        self.x_fixed_vals = np.array(x_fixed_vals)
```
dweindl: Check that len(x_fixed_vals) == len(x_fixed_indices) ?

FFroehlich:
Would be great to implement a testcase for the added functionality ;)

yannikschaelte:
SORRY FOR THE BIG MERGE!!!

Merging with another pull request where, among other things, the objective was enabled to record a history, a couple of more changes became necessary:

* An ObjectiveHistory object was introduced which separated the history functionality from the objective class. It would be of interest to further this separation. There are different update/reset methods (one might see these as external configurators) in the objective which enable the problem and the optimization routine to make changes to these values.
* objective.has_fval etc. properties were introduced which compute the logic to see what the objective function supports regarding gradients, residuals and so on. These properties are now used in the Optimizers to feed their routines with available function values. It would be desirable to be a bit more flexible here with regard to finite differences. Beforehand there must have been a logic error somewhere, see also #41.
* Certain parameters were reshuffled and put into options objects, which enables easier tossing around.
* The optimize.minimize function was decomposed to make it better readable.

* One design point: Now everywhere ``x`` instead of ``par`` (or ``pars`` or whatever) is used. Just because it's shorter.

One remark on the underlying idea: Objective should be rather free from dimensions, because in particular different parameter fixings can be used, and the dimensions can be read automatically from the data (bounds etc) passed to a Problem. Therefore, the problem deals with everything dimension-related, and just updates the objective accordingly. It also provides functions to switch between representations, as well as the dimension used in the optimization etc.

Please indicate if you would do sth different, e.g. the options constructs or whatever.

yannikschaelte:
@dweindl @FFroehlich maybe you can have a short look bc due to the merge many changes were required. I will merge to master tomorrow then hopefully. the problems with the notebooks should be resolved.

pypesto/objective/objective.py
```diff
@@ -0,0 +1,910 @@
...
+        # create table row
+        values = [
+            used_time,
+            self.n_fval, self.n_grad, self.n_hess, self.n_res, self.n_sres,
+            fval, grad, hess, res, sres,
+            x
+        ]
+
+        # check whether to append to trace
+        if not self.options.trace_all and fval >= self.fval_min:
```
FFroehlich: we could check for that a bit earlier, right?

pypesto/optimize/optimize.py
```diff
@@ -25,71 +74,67 @@ def minimize(problem, optimizer,
...
+        err) -> OptimizerResult:
+    """
+    Handle exceptions. Raise exception if allow_faile_starts is False,
+    otherwise return a dummy pypesto.OptimizerResult.
+    """
+    if allow_failed_starts:
+        print(('start ' + str(j_start) + ' failed: {0}').format(err))
+        optimizer_result = recover_result(objective, startpoint, err)
+        return optimizer_result
+    raise err
```
FFroehlich: does this properly rethrow the error with full traceback? If not, I would prefer the inline version.

yannikschaelte: true, the stacktrace is wrong, starting in the new method. will change that.

pypesto/objective/objective.py
yannikschaelte: At this point one remark: I found that pd.DataFrame.toc appending is really slow compared to list.append() (like 10s to <1s), but of course has better readability. Usually won't matter for our ODE problems.

pypesto/objective/amici_objective.py
```diff
@@ -0,0 +1,266 @@
+import numpy as np
+import copy
+from .objective import Objective
+
+try:
+    import amici
```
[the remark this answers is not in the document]
yannikschaelte: requires https://github.com/ICB-DCM/AMICI/issues/426, then we should test that here using ``amici.__version__`` and see here: https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python

pypesto/objective/amici_objective.py
```diff
@@ -0,0 +1,266 @@
...
+                rdata = amici.runAmiciSimulation(
+                    self.amici_model,
+                    self.amici_solver,
+                    self.preequilibration_edata[fixedParameters]['edata'])
+
+                if rdata['status'] < 0.0:
+                    return self.get_error_output(sensi_orders, mode)
+
+                self.preequilibration_edata[fixedParameters]['x0'] = \
+                    rdata['x0']
```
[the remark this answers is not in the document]
yannikschaelte: @FFroehlich ?

Issue #41, opened 2018-09-03
optimization seems to fail


DantongWang:
Visualization fails when running the 'rosenbrock' example because the optimized cost function values are all 'Inf'. Might be something wrong in the optimize function.

yannikschaelte:
I can't completely tell why that occurs, but I did a few changes in a soon-to-be-merged pull request, where the problem no longer occurs. So, should be solved like tomorrow.
Upvoting again to have the notebooks as travis tests :) .

dweindl:
After having fixes this, please re-enable optimization in doc/example/conversion_reaction.ipynb. Disabled in #44.

PR #44, opened 2018-09-04
Run example notebooks on travisCI (Closes #34)

### LINKED ISSUE #34: Run example notebooks on travis
Example notebooks should be run on travis as additional integration tests.

dweindl:


.travis.yml
```diff
@@ -20,24 +20,29 @@ addons:
...
   - "sh -e /etc/init.d/xvfb start"
   - sleep 3 # give xvfb some time to start
 
 # run tests
-
 script:  
   - python3 -m flake8 --ignore=F401 --exclude=build,doc,example
   - python3 -m pytest --cov=pypesto ./test
   - coverage xml
+  - test/runNotebook.sh doc/example/rosenbrock.ipynb
```
yannikschaelte: would it be possible to change the script to only pass the directory, and then run all notebooks (*.ipynb) therein? would make it easier to catch up when notebooks are added.

dweindl: Can do, if you prefer. But I won't add an exclusion filter, in case you decide to not run all... ;)

yannikschaelte: think will be better to test all. on that occasion, in conversion_reaciont.ipynb the profiles and sampling things need to be commented out, otherwise they won't pass.

test/runNotebook.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 # Run jupyter notebook, show output only on error
 tempfile=$(tempfile)
-jupyter nbconvert --debug --stdout --execute --to markdown $@ &> $tempfile
+ipython3 nbconvert --debug --stdout --execute --to markdown $@ &> $tempfile
```
FFroehlich: 💩

dweindl: Back in the days... :)

dweindl:
@yannikschaelte Can you please take care of https://travis-ci.com/ICB-DCM/pyPESTO/builds/83771943#L1023

yannikschaelte:
@dweindl sure, that seems to be related to #41. I can look into that tomorrow. there are some issues with how the amici objective gets read in the scipy optimizer.

yannikschaelte:
note there were currently issues also with the non-amici notebooks (i.e. rosenbrock), where the function values were all inf (i think bc the optimization broke off right after the start, and was then filled in from the objective_wrapper), so that the visualization errors

dweindl:
Thanks for the hint. Ran doc/example/rosenbrock.ipynb 10 times locally, didn't error.

PR #53, opened 2018-09-18
Various small bugfixes

FFroehlich:
1) fixes an issue with startpoint resampling
2) removes redundant computation of lower order derivatives for AmiciObjective
3) adds some debug logging to amici objective

### DIFF
--- a/pypesto/objective/amici_objective.py
+++ b/pypesto/objective/amici_objective.py
@@ -163,7 +163,10 @@ class AmiciObjective(Objective):
             logger.debug('=== DATASET %d ===' % data_index)
             logger.debug('status: ' + str(rdata['status']))
             logger.debug('llh: ' + str(rdata['llh']))
-            logger.debug('y:\n' + str(rdata['y']))
+            if not rdata['t_steadystate'] == np.nan:
+                logger.debug('t_steadystate: ' + str(rdata['t_steadystate']))
+            logger.debug('res:\n' + str(rdata['res']))
+
 
             # check if the computation failed
             if rdata['status'] < 0.0:
@@ -176,7 +179,6 @@ class AmiciObjective(Objective):
                     snllh -= rdata['sllh']
                     # TODO: Compute the full Hessian, and check here
                     ssnllh -= rdata['FIM']
-                return nllh, snllh, ssnllh
 
             elif mode == Objective.MODE_RES:
                 res = np.hstack([res, rdata['res']]) \
@@ -184,7 +186,12 @@ class AmiciObjective(Objective):
                 if sensi_order > 0:
                     sres = np.vstack([sres, rdata['sres']]) \
                         if sres.size else rdata['sres']
-                return res, sres
+
+        if mode == Objective.MODE_FUN:
+            return nllh, snllh, ssnllh
+        elif mode == Objective.MODE_RES:
+            return res, sres
+
 
     def preprocess_preequilibration(self, data):
         original_fixed_parameters_preequilibration = None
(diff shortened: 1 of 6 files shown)


FFroehlich:
@yannikschaelte could you please have a look at this?

pypesto/objective/amici_objective.py
```diff
@@ -84,17 +79,31 @@ def sres(x):
         # extract parameter names from model
         self.x_names = list(self.amici_model.getParameterNames())
 
+    def fun(self, x):
+        return self.call_amici(
+            x,
+            Objective.MODE_FUN
+        )
+
+    def res(self, x):
```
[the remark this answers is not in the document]
yannikschaelte: is this overwriting of fun, res necessary? codacy complains and it might be error-prone and is also not so nice to have ``overwritefun`` in the constructor. i think it is valid to save the sensi_orders temporarily, as you did in the Objective. But maybe one can just have in ``__init__``
```
def fun(x):
    return self._call_amici(x, Objective.MODE_FUN)
...
super().__init__(...fun=fun,...)
```
and similarly for res?

pypesto/objective/amici_objective.py
```diff
@@ -162,20 +179,18 @@ def call_amici(
...
-        return Objective.map_to_output(
-            sensi_orders,
-            mode,
-            fval=nllh, grad=snllh, hess=ssnllh,
-            res=res, sres=sres
-        )
+        if mode == Objective.MODE_FUN:
+            return nllh, snllh, ssnllh
+        elif mode == Objective.MODE_RES:
+            return res, sres
```
yannikschaelte: why is this here? i think fun, i.e. here _call_amici, should return just the values as a tuple that were asked for in tuple form, so e.g. fval if sensi_orders is (0,), or (fval, grad) if (0,1). now always 3 resp. 2 values are returned, looks like this migt lead to errors when using e.g. derivative-free optimizers. does this really work? haven't checked, can do that when i have time.

FFroehlich: `_call_amici` implements `fun` and `res` which are supposed to return the outputs exactly in that order. The subindexing according to the specified `sensi_orders` is applied in `_call_mode_fun` and `_call_mode_res` and `map_to_output`.

pypesto/objective/objective.py
```diff
@@ -244,21 +264,34 @@ def _update_trace(self, x, mode, result):
...
+                else res_result
+            sres = None if not self.options.trace_record_res \
+                else sres_result
+            chi2 = None if not self.options.trace_record_chi2 \
+                           or res_result is None \
+                else np.inner(res_result,res_result)/2
+            schi2 = None if not self.options.trace_record_chi2 \
+                            or sres_result is None \
+                            or res_result is None \
+                else res_result * res_result
```
yannikschaelte: for chi2, can't we use the res_to_fval function? then the normalization is always the same. is the formula for schi2 correct? res_result * res_sresult maybe?

pypesto/objective/objective.py
```diff
@@ -646,8 +734,10 @@ def map_to_output(sensi_orders, mode, **kwargs):
                 output += (kwargs[Objective.RES],)
             if 1 in sensi_orders:
                 output += (kwargs[Objective.SRES],)
-        if len(output) == 1:
-            # return a single value not as tuple
+        if len(output) == 1 and (
```
yannikschaelte: are these 2 additional checks really needed?

FFroehlich: yep :(

PR #57, opened 2018-10-18
Several little updates

yannikschaelte:
* disentangle the objective.py file which became too long
* add Boehme example to notebooks (as an example of how the benchmark models can be used)
* add some tests for the output of the objective
* update version from 0.0.1 to 0.0.2


doc/example/boehm_JProteomeRes2014/benchmark_import.py
```diff
@@ -0,0 +1,94 @@
...
+            e.g. "Fujita_SciSignal2010", file name relative to current
+            working directory
+
+        Returns
+        -------
+
+        dp: DataProvider
+            a DataProvider to the generated h5_file
+        """
+
```
dweindl: Something's missing here

PR #59, opened 2018-11-07
Feature fixed par plist for AmiciObjective

FFroehlich:
Modifies AmiciObjective to make use of Model.setParameterList() to reduce computational cost if a large number of parameters is fixed.

### DIFF
--- a/pypesto/objective/amici_objective.py
+++ b/pypesto/objective/amici_objective.py
@@ -2,7 +2,7 @@ import numpy as np
 import copy
 import logging
 from .objective import Objective
-from .constants import MODE_FUN, MODE_RES
+from .constants import MODE_FUN, MODE_RES, HESS
 
 try:
     import amici
@@ -125,6 +125,7 @@ class AmiciObjective(Objective):
         plist = self.amici_model.getParameterList()
         plist = [plist[idx] for idx in x_free_indices]
         self.amici_model.setParameterList(plist)
+
         def postprocess(result):
             if HESS in result:
                 hess = result[HESS]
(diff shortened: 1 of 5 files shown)


FFroehlich:
@yannikschaelte

pypesto/objective/amici_objective.py
```diff
@@ -86,6 +84,108 @@ def res(x, sensi_orders):
         # extract parameter names from model
         self.x_names = list(self.amici_model.getParameterNames())
 
+    def get_bound_fun(self):
```
yannikschaelte: Looks good to me. I think that beforehand amici always had to compute sensis for all parameters, and using model.setParameterList() now, only those that are not fixed are computed?

Can you shortly document why get_bound_fun and rebind_fun are necessary? Just so we don't run into similar problems in the future.

FFroehlich: Correct. Will add documentation

yannikschaelte: @FFroehlich : In the doc you say for get_bound_fun(self) that
> Generate a fun function that calls _call_amici with MODE_FUN. Defining
        a non-class function that references self as a local variable will bind
        the function to a copy of the current self object and will
        accordingly not take future changes to self into account.

I currently need to work on related stuff here. I do not completely understand what the problem is. Could you shortly explain please? If I do the following:
```python
class A:
    def __init__(self):
        self.x = 1
        def nested_fun(y):
            print(self.x, y)
        self.fun = nested_fun

a = A()
a.fun(2)  # gives [1 2]
a.x = 42
a.fun(2)  # gives [42 2]
```
I get the output I would expect, i.e. the references inside self are updated correctly.

PR #64, opened 2018-11-13
fixed problem in visualization of parameters for fixed parameters

DantongWang:
Fixed the problem and added a flag to plot (or not plot) fixed parameters. Additionally I added this parameter visualization part in the 'fixed_parameter' example.


pypesto/problem.py
```diff
@@ -92,19 +92,19 @@ def __init__(self,
         self.dim_full = dim_full if dim_full is not None else self.lb.size
 
         if x_fixed_indices is None:
-            x_fixed_indices = np.array([])
-        self.x_fixed_indices = np.array(x_fixed_indices, dtype=int)
+            x_fixed_indices = []
+        self.x_fixed_indices = [int(i) for i in x_fixed_indices]
```
yannikschaelte: needed to do that because numpy integers cannot be taken as indices for lists for some reasons apparently.

PR #71, opened 2018-11-25
Add codecov badge

FFroehlich:


yannikschaelte:
@FFroehlich I had done some badgies also in branch feature_badges (https://github.com/ICB-DCM/pyPESTO/tree/feature_badges). Could you reconcile both? I think in particular it would be good to have them also on the readthedocs homepage.

PR #74, opened 2018-11-26
Feature profiles

paulstapor:
This is a first, very raw implementation for profiling.

Profiling results are stored as a nested list (maybe you want to run and save more than one profiling run in your results, in order to merge the runs later on).

I included a very simple visualization tool as well, which expects a list of numpy-arrays.

I have to finish up the documentation and the sanity checks the next days.

Currently, only optimization-based profiling with fixed step-sized are allowed, but the interface is flexible enough to allow adaptive step sizes. This should anyhow be not complicated to add...


dweindl:
Cool. Could you add a short usage example to one of the example notebooks?

paulstapor:
Added it to the Rosenbrock example

dweindl:
> Added it to the Rosenbrock example

Thanks

pypesto/problem.py
```diff
@@ -149,15 +148,112 @@ def normalize_input(self):
...
             raise AssertionError(
                 "x_fixed_indices and x_fixed_vals musti have the same length."
             )
 
+    def fix_parameters(self, parameter_indices, parameter_vals):
+        """
+        Fix specified parameters to specified values
+        """
+
+        if not parameter_indices is list:
```
yannikschaelte: you mean ``if not isinstance(parameter_indices, list)``?

paulstapor: yep...
for some reason, that worked (or was never hit...)

pypesto/problem.py
```diff
@@ -149,15 +148,117 @@ def normalize_input(self):
...
             )
 
+    def fix_parameters(self, parameter_indices, parameter_vals):
+        """
+        Fix specified parameters to specified values
+        """
+        if not isinstance(parameter_indices, list):
+            parameter_indices = [parameter_indices]
+
+        if not parameter_vals is list:
```
yannikschaelte: same :)

pypesto/problem.py
```diff
@@ -149,15 +148,117 @@ def normalize_input(self):
...
+        lb = []
+        ub = []
+
+        # first clean to be fixed indices to avoid redundancies
+        for i_index, i_parameter in enumerate(parameter_indices):
+            # check if parameter was already fixed, otherwise add it to the
+            # fixed parameters
+            if i_parameter in self.x_fixed_indices:
+                self.x_fixed_vals[self.x_fixed_indices.index(i_parameter)] = \
+                    parameter_vals.pop(i_index)
```
yannikschaelte: hm. does this work when there are multiple cases where i_parameter in self.x_fixed_indices? when you pop a value, it means that parameter_vals length decreases, such that i_index does not point to the correct element between parameter_indices and parameter_vals?

pypesto/profile/profile.py
```diff
@@ -0,0 +1,254 @@
...
+    optimizer: pypesto.Optimizer
+        The optimizer to be used along each profile.
+
+    profile_index: integer array, optional
+        array with parameter indices, whether a profile should
+        be computed (1) or not (0)
+
+    result_index: integer, optional
+        index, from which optimization result profiling should be started
+        (default: global optimum, i.e., index = 1)
```
yannikschaelte: note: in python, indexing is 0-based.

pypesto/profile/profile.py
```diff
@@ -0,0 +1,254 @@
...
+        Returns a valid options object.
+
+        Parameters
+        ----------
+
+        maybe_options: OptimizeOptions or dict
+        """
+        if isinstance(maybe_options, ProfileOptions):
+            return maybe_options
+        options = ProfileOptions(**maybe_options)
```
yannikschaelte: i think that to be able to use this, the constructor ``__init__`` above must accept default_step_size and ratio_min as arguments (I think that would make sense anyway).

pypesto/profile/profile.py
```diff
@@ -0,0 +1,254 @@
...
+        profile_index = np.ones(problem.dim_full)
+
+    # profile startpoint method
+    if create_profile_startpoint is None:
+        def create_next_startpoint(x, par_index, par_direction):
+            return fixed_step(x, par_index, par_direction,
+                              step_size=profile_options.default_step_size)
+
+    # check profiling options
+    if profile_options is None:
```
yannikschaelte: profile_options needs to be instantiated before the previous block

pypesto/profile/profile_startpoint.py
```diff
@@ -0,0 +1,46 @@
...
+            The index of the current profile
+
+        par_direction: int
+            The direction, in which the profiling is done (1 or -1)
+       """
+    delta_x = np.zeros(len(x))
+    step_size = 0.01
+    delta_x[par_index] = par_direction * step_size
+
+    return x + delta_x
```
yannikschaelte: what exactly is the difference between these 2?

pypesto/visualize/profiles.py
```diff
@@ -0,0 +1,122 @@
+import matplotlib.pyplot as plt
+from matplotlib.ticker import MaxNLocator
+import numpy as np
+
+
+def profiles(result, ax=None, profile_indices=None):
+    """
+    Plot waterfall plot.
```
yannikschaelte: i guess nope waterfall.

pypesto/profile/profile.py
paulstapor: Done

pypesto/profile/profile.py
paulstapor: Done

pypesto/profile/profile.py
```diff
@@ -0,0 +1,254 @@
...
+    profile_options: pypesto.ProfileOptions, optional
+        Various options applied to the profile optimization.
+
+    optimize_options: pypesto.OptimizeOptions, optional
+        Various options applied to the optimizer.
+    """
+
+    # profiling indices
+    if profile_index is None:
+        profile_index = np.ones(problem.dim_full)
```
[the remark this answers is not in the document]
paulstapor: The idea is to loop over all parameters and then abort profiling if a parameter is fixed (as done later in the code)

pypesto/profile/profile.py
```diff
@@ -0,0 +1,254 @@
...
+    optimize_options = OptimizeOptions.assert_instance(optimize_options)
+
+    # create the profile result object and retrieve global optimum
+    global_opt = initialize_profile(problem, result, result_index)
+
+    # loop over parameters for profiling
+    for i_parameter in range(0, problem.dim_full):
+        if (profile_index[i_parameter] == 0) | \
+                (i_parameter in problem.x_fixed_indices):
+            continue
```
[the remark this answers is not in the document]
paulstapor: right

pypesto/visualize/profiles.py
paulstapor: corrected

pypesto/profile/profile_startpoint.py
paulstapor: There's not yet one. At some point, simple_step should be the 1D (so 0th-order) update with adaptive step size. But for now, I should delete that routine.

paulstapor:
Okay, I think now everything is adressed, new unit tests are in place. Please somebody feel free to review the code :)

pypesto/problem.py
```diff
@@ -89,15 +88,23 @@ def __init__(self,
...
             x_fixed_indices = []
         self.x_fixed_indices = [int(i) for i in x_fixed_indices]
 
+        # We want the fixed values to be a list, since we might need to add
+        # or remove values during profile computation
         if x_fixed_vals is None:
-            x_fixed_vals = np.array([])
-        self.x_fixed_vals = np.array(x_fixed_vals)
+            x_fixed_vals = []
+        if not isinstance(x_fixed_vals, list):
```
yannikschaelte: I assume a numpy array would not be changed to a list here. maybe ``x_fixed_vals = list(x_fixed_vals)`` is the thing to do here?

pypesto/problem.py
```diff
@@ -89,15 +88,23 @@ def __init__(self,
         self.lb = np.array(lb).flatten()
         self.ub = np.array(ub).flatten()
 
+        self.lb_full = np.array(lb).flatten()
+        self.ub_full = np.array(ub).flatten()
```
yannikschaelte: can you document lb_full, ub_full in the Parameters above?

PR #86, opened 2019-01-09
fix(problem) fix problem.get_reduced_vector

FFroehlich:
How could this not be caught by tests?


dweindl:
> I guess we internally only use numpy arrays such that this is possible?

This is maybe something to discuss / document. Whether we want to support only numpy.ndarray or also python lists. Everywhere or only in the public API (which is what exactly?)?

yannikschaelte:
Internally, I everywhere used np arrays. When something comes from public, it is converted to ndarray first. In this sense, public API is every interface where the user puts in data.

Sometimes I also used methods like .size (instead of len), which are only available for numpy. On second thoughts, this should be revised I think, and list-based methods used. When a ndarray is explicitly needed in a method, it should maybe be converted first, also to ensure e.g. the dimensions are correct.

paulstapor:
I would vote for using lists for free and fixed indices, since I also want to use/am using fixing and "unfixing" of parameters in profiling. Lists are more comfortable to use with .pop and .append for this.

PR #92, opened 2019-01-16
add xvfb to services

paulstapor:
Trying to fix graphical interface tests in travis


paulstapor:
`services -xvfb` is not necessary, as it seems. But a change in the xvfb call and choosing a display is.

PR #96, opened 2019-01-21
feature(CI) automatic travis deploy to pypi

FFroehlich:


paulstapor:
Do I get it correctly that this would directly update the version on pypi from the pypesto master branch after creating a new release?
If so: Cool! :+1:

FFroehlich:
Not after every pull request, only when a version is tagged.

I did not use encryption here but set the password as token, probably a bit less secure but yes ` $TWINE_USERNAME` is effectively doing the same job.

PR #110, opened 2019-02-07
parameter summary for Problem

FFroehlich:


pypesto/problem.py
```diff
@@ -322,3 +322,23 @@ def get_reduced_matrix(self, x_full):
...
+        Prints a summary of what parameters are being optimized and
+        what parameter boundaries are
+        """
+        
+        print(
+            pd.DataFrame(
+                index=problem.x_names, 
+                data={
+                    'free': [
+                        idx not in problem.x_fixed_indices
```
yannikschaelte: there exists also problem.x_free_indices: https://github.com/ICB-DCM/pyPESTO/blob/834ba3ba556b00c376609459c7416961beaf0da9/pypesto/problem.py#L111

FFroehlich: True, I assumed that `problem.x_fixed_indices` has less elements than `self.x_free_indices`, but with `in` we can terminate as soon as we have found the correct element while for `not in` needs to check all elements. Not that efficiency is likely to be ever relevant here, but checking against `problem.x_free_indices` is probably better readable code.

PR #118, opened 2019-02-11
Create copies of amici.Model and amici.Solver (Closes #67)

### LINKED ISSUE #67: AmiciObjective.__init__ should create copies of model & solver

dweindl:


pypesto/objective/amici_objective.py
```diff
@@ -106,8 +106,8 @@ def __init__(self,
             options=options
         )
 
-        self.amici_model = amici_model
-        self.amici_solver = amici_solver
+        self.amici_model = amici.ModelPtr(amici_model.clone())
```
yannikschaelte: can't hurt, but I wouldn't do too much cloning (since we deepcopy the full objective already when creating a problem)

dweindl: I'd rather copy one time too often. Modifying one of those pointers somewhere else in this session is really bad to debug.

PR #121, opened 2019-02-13
Test parameter mapping

yannikschaelte:
tidy up some code by using petab.core.get_simulation_conditions by default. Corresponds to https://github.com/ICB-DCM/PEtab/pull/57.

 And add function that captures behavior that was wrong in a previous petab version. should be fixed now, but maybe it is good to have this check in there for future versions (?).


pypesto/objective/petab_import.py
```diff
@@ -341,58 +357,61 @@ def rdatas_to_measurement_df(self, rdatas, model=None):
...
-    # of those rows only once (and an additional counting row)
-    simulation_conditions = measurement_df.groupby(
-        grouping_cols).size().reset_index()
-
-    return grouping_cols, simulation_conditions
-
-
-def _get_rows_for_condition(measurement_df, grouping_cols, condition):
+    Currently checks whether nan values in the parameter mapping table
+    corresponds to nan columns in the edatas, corresponding to missing
```
dweindl:
```suggestion
    correspond to nan columns in the edatas, corresponding to missing
```

pypesto/objective/petab_import.py
```diff
@@ -341,58 +357,61 @@ def rdatas_to_measurement_df(self, rdatas, model=None):
...
+    data points.
     """
-    Extract rows in `measurement_df` according to `grouping_cols`
-    for `condition`.
-
-    Returns
-    -------
+    # regular expression for noise and observable parameters
+    rex = "(noise|observable)Parameter[0-9]+_"
```
dweindl: Feel free to re.compile

yannikschaelte: yes, it's better re.compile than to recompile every time. thanks!

Issue #122, opened 2019-02-13
PetabImporter: Allow replicates


yannikschaelte:
Currently (see https://github.com/ICB-DCM/pyPESTO/blob/develop/pypesto/objective/petab_import.py#L201), if there are multiple replicates for an observable in a simulation condition, I guess only the last one is passed to the amici.ExpData. I guess that it should be possible to pass timepoints multiple times to the ExpData, i.e. to create a larger table, in which all data points can be filled, having nans for non-existing replicates.

Some of the Benchmark models, e.g. Schwen, require this.

dweindl:
... or at least fail if anything is overriden.

yannikschaelte:
Implemented it now in #123 such that failures are not necessary, but the data table is just enlarged accordingly and all data points are filled in iteratively

PR #123, opened 2019-02-13
allow arbitrary number of replicates in measurement_df when creating edatas

yannikschaelte:
here, replicate means that time point, observableId, preequilibrationConditionId, simulationConditionId coincide. This is allowed by the petab format. Solves #122

### DIFF
--- a/pypesto/objective/petab_import.py
+++ b/pypesto/objective/petab_import.py
@@ -170,16 +170,29 @@ class PetabImporter:
             # amici.ExpData for each simulation
 
             # extract rows for condition
-            cur_measurement_df = petab.core.get_rows_for_condition(
+            df_for_condition = petab.core.get_rows_for_condition(
                 measurement_df, condition)
 
             # make list of all timepoints for which measurements exist
             timepoints = sorted(
-                cur_measurement_df.time.unique().astype(float))
+                df_for_condition.time.unique().astype(float))
 
             # init edata object
             edata = amici.ExpData(model.get())
-            edata.setTimepoints(timepoints)
+
+            # find rep numbers of time points
+            timepoints_w_reps = []
+            for time in timepoints:
+                # subselect for time
+                df_for_time = df_for_condition[df_for_condition.time == time]
+                # rep number is maximum over rep numbers for observables
+                n_reps = max(df_for_time.groupby(
+                    ['observableId', 'time']).size())
+                # append time point n_rep times
+                timepoints_w_reps.extend([time] * n_reps)
+
+            # set time points in edata
+            edata.setTimepoints(timepoints_w_reps)
 
             # handle fixed parameters
             _handle_fixed_parameters(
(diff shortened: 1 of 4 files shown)


yannikschaelte:
äh, I had done sth wrong w/ the replicates filling into the amici.ExpData. Now I added a test that checks whether all values are **actually** contained. @dweindl Would you have a second look?

test/test_petab_import.py
```diff
@@ -68,7 +69,47 @@ def test_3_optimize(self):
...
+                             and not col.endswith("_std")]]
+
+        meas_df = importer.petab_problem.measurement_df
+
+        # amici_df subset measurement_df
+        for _, row in meas_df.iterrows():
+            val = row.measurement
+            # test if amici_df contains this value somewhere
+            # this will test up to np.isclose closeness
+            self.assertTrue(np.isclose(amici_df, val).any(axis=1).any())
```
dweindl: This will not detect a lost data point if there is another one with the same value, will it?

Could flatten and sort the array and check if each entry is close to corresponding the sorted measurement column for that condition.

pypesto/objective/petab_import.py
```diff
@@ -207,19 +207,32 @@ def create_edatas(self, model=None, simulation_conditions=None):
...
+
+                # remember used time indices for each observable
+                time_ix_for_obs_ix = {}
+
                 # iterate over measurements
                 for _, measurement in df_for_time.iterrows():
-                    # get observable index
+                    # extract observable index
                     observable_ix = observable_ids.index(
                         f'observable_{measurement.observableId}')
```
dweindl: I'm really annoyed by all those `f'observable_{...}'`in the code and the ambiguity of observable id... :expressionless:

yannikschaelte: you mean that it's ``observable_`` in the sbml, but not in the measurement file? yeah, kind of not optimal, and it would be good to have a unique id used everywhere.
but it's good to have the id in the sbml to directly see what's an observable, and have it not in the measurement_df, because there we really don't need it as it does not help readability. so this solution also makes sense ...

test/test_petab_import.py
yannikschaelte: No, it will not detect such a point. But the points in the Schwen model are quite diverse such that we definitely have many cases where both values need to end up in the file.

I was not sure how stable the amici dataframe format is. If one can assume that the time column will always be denoted ``time`` and the observable columns ``observable_{OBS_ID}`` (hej, there the observable_ is again ;) ), then I could indeed do that.

yannikschaelte: funnily, in the measurement table, the lines https://github.com/LoosC/Benchmark-Models/blob/hackathon/hackathon_contributions_new_data_format/Schwen_PONE2014/measurementData_Schwen_PONE2014.tsv#L27 and following are exactly the same though (don't know if that makes sense).

pypesto/objective/petab_import.py
dweindl: Yes. And that we call both observable_id, which regularly confuses me. General issue, not specific to this PR.

test/test_petab_import.py
dweindl: Very reproducible measurement, yeah, wow.

yannikschaelte: > Could flatten and sort the array and check if each entry is close to corresponding the sorted measurement column for that condition.

Done that now. The only thing that could be missed is when the wrong observable gets assigned the data.

yannikschaelte: now recovering the full details, including which data for which observable id (and time point) :)

dweindl: Great

PR #132, opened 2019-02-24
Feature aggregate objective

FFroehlich:


pypesto/objective/aggregate.py
```diff
@@ -0,0 +1,175 @@
+import numpy as np
+
+from .objective import Objective
+
+
+class AggregateObjective(Objective):
+    """
+    This class allows to create an aggregateObjective from a list of
```
yannikschaelte: aggregate(d) objective

pypesto/objective/aggregate.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # check if all objectives consistently accept sensi orders in fun/res
+        # and adopt the same behaviour in aggregate
+        for attr in [
+            'fun', 'grad', 'hess', 'hessp', 'res', 'sres'
+        ]:
+            if any(
+                getattr(objective, attr) is None
+                for objective in objectives
+            ):
```
yannikschaelte: if i understand correctly, attr is set to None as soon as any objective.attr is None? Maybe we want to check that all are None, to have consistency, just as done for boolean values?

pypesto/objective/aggregate.py
```diff
@@ -0,0 +1,175 @@
...
+                            f'was {type(objectives)}.')
+
+        if not all(
+                isinstance(objective, Objective)
+                for objective in objectives
+        ):
+            raise TypeError('Objectives must only contain elements of type'
+                            'pypesto.Objective')
+
+        self.objectives = objectives
```
yannikschaelte: picky, but one should also check that ``len(objectives) > 0`` for various reasons.

pypesto/objective/aggregate.py
```diff
@@ -0,0 +1,175 @@
...
+        else:
+            sres = None
+
+        # skip iobj=0 after initialization, stack matrices
+        for iobj in range(1, len(self.objectives)):
+            rval = self.objectives[iobj].res(x, sensi_orders)
+            if res is not None:
+                res = np.hstack([res, np.asarray(rval['res'])])
+            if sres is not None:
+                sres = np.vstack([sres, np.asarray(rval['sres'])])
```
yannikschaelte: a problem might occur when one of the objectives errors, I'm not sure what happens then. maybe just wait and see if anything breaks then.

pypesto/objective/aggregate.py
```diff
@@ -0,0 +1,175 @@
...
+    def aggregate_fun_sensi_orders(self, x, sensi_orders):
+        rvals = [
+            objective.fun(x, sensi_orders)
+            for objective in self.objectives
+        ]
+        return {
+            key: sum(rval[key] for rval in rvals)
+            for key in ['fval', 'grad', 'hess']
+            if key in rvals[0]
+        }
```
yannikschaelte: one issue: it is possible to call an objective with ``return_dict = True``. In that case the whole dict that is created is returned. This was introduced to have an easy way to get the ``RDATAS`` from AmiciObjectives. Problem now is that these get lost in the aggregation step. Maybe one could check whether an RDATAS field exists (pypesto.constants.RDATAS), and then just concatenate them here?

pypesto/objective/aggregate.py
FFroehlich: Should be fine, although we currently do not do any early termination ...

PR #114, opened 2019-02-10
Feature plotting

paulstapor:
The main idea of this branch is to provide a set of additional features for plotting. It is branched out of "feature figure sizes", so we will have to update it. The idea of this branch is to provide
 - [x] the possibility of plotting reference points/parameters like in PESTO
 - [x] a function that handles options like axis specifications and things that are identical for all visualization routines
 - [x] a routine for plotting the optimizer trajectory
 - [x] plotting of many different optimization results at once, in order to compare waterfall plots, for example (could be realized via, e.g., a list of result objects)
 - [x] unit tests for the new features
This pr will fix #108, #115, #111, #98

PR #133, opened 2019-02-25
Develop

yannikschaelte:
* Multi process parallelization
* Fix error for multiple conditions
* Fix documentation build
* Greatly extend visualization routines
* multiple minor things

Version 0.0.3 -> 0.0.4


yannikschaelte:
didn't know squash-merging was possible. will look into that next time.

paulstapor:
As we get PEtab import here: Please make sure, that we have an example for this, which documents the use... Maybe add a list of models, for which this is working, as a comment in a corresponding notebook.

yannikschaelte:
good point. not completely clear yet if the files should be included in the examples folder, or if we just continue git-downloading the benchmark repo (which hopefully might also get smaller whence the old dependencies are removed), but we should definitely maintain a list of models for which pypesto works (i.e. everything from import, compilation to parameter estimation).

paulstapor:
As this would go perfectly in line with #114, do we maybe want to wait for  #106 being merged?

yannikschaelte:
currently, there were some updates to petab and amici which still need to be integrated (and both toolboxes updated as well). this is why the tests still fail. hopefully fixed in a few days.

PR #137, opened 2019-03-02
Fix petab amici

yannikschaelte:
Make pypesto ready for the lates pypesto 0.0.0a11. Will need amici 0.10.1 for the tests to pass.


yannikschaelte:
everything passing except amici-related stuff, which worked locally, so I'll just merge this now and wait for amici to be updated to 0.10.1, and then run the travis checks again before merging to master (where tests MUST pass first).

PR #139, opened 2019-03-03
fix amici tolerances in sbml_conversion test

yannikschaelte:
the sundials version used in amici changed, so there probably is a difference in behavior nothing can be done anything about. let's just be a bit more relaxed about the tolerances in the test, since the obtained values are definitely still close.


yannikschaelte:
yes, that's definitely right. for the MODE_FUN, they have always been that way already, apparently now also for MODE_RES needed. The actual difference are indeed also at 1e-3 to 1e-2, so that is actually needed.

PR #140, opened 2019-03-04
New examples

paulstapor:
Very most of the new functionality is implemented into the rosenbrock example, but some things can also be found in the petab example...


paulstapor:
Now, I think I really double checked all the stuff or at least 95% of it...
I hope, I found most of the obvious errors...

pypesto/visualize/optimizer_history.py
```diff
@@ -308,6 +308,9 @@ def get_vals(vals, scale_y, offset_y, start_indices):
 
     vals: list
         list of numpy arrays of size 2 x len(start_indices)
+
+    offset_y:
+        offset for the y-axis, as this is supposed to be in log10-scale
```
yannikschaelte:
```suggestion
        offset for the y-axis, if this is supposed to be in log10-scale
```
?

pypesto/visualize/parameters.py
```diff
@@ -49,6 +49,10 @@ def parameters(results, ax=None, free_indices_only=True, lb=None, ub=None,
         Flag indicating whether alpha for large clusters should be reduced to
         avoid overplotting (default: True)
 
+    start_indices: list or int
+        list of integers specifying the multistart to be plotted or
```
yannikschaelte:
```suggestion
        list of integers specifying the multistarts to be plotted or
```

pypesto/visualize/parameters.py
```diff
@@ -198,6 +218,10 @@ def handle_inputs(result, free_indices_only, lb=None, ub=None):
         If not None, override result.problem.lb, problem.problem.ub.
         Dimension either result.problem.dim or result.problem.dim_full.
 
+    start_indices: list or int
+        list of integers specifying the multistart to be plotted or
```
yannikschaelte:
```suggestion
        list of integers specifying the multistarts to be plotted or
```

pypesto/visualize/profiles.py
```diff
@@ -6,8 +6,8 @@
 from .misc import process_result_list
 
 
-def profiles(results, fig=None, profile_indices=None, size=(18.5, 6.5),
-             reference=None, colors=None, legends=None):
+def profiles(results, ax=None, profile_indices=None, size=(18.5, 6.5),
+             reference=None, colors=None, legends=None, profile_list=0):
```
yannikschaelte: ``profile_list`` sounds like one would pass a list. ``profile_list_ix/index``?

pypesto/visualize/profiles.py
```diff
@@ -248,10 +252,14 @@ def handle_reference_points(ref, ax, fvals):
...
 
+            # create legend for reference points
+            if i_ref.legend is not None:
+                i_ax.legend()
+
     return ax
 
 
-def handle_inputs(result, profile_indices):
+def handle_inputs(result, profile_indices, profile_list=0):
```
yannikschaelte: see above

pypesto/visualize/profiles.py
```diff
@@ -276,9 +287,11 @@ def handle_inputs(result, profile_indices):
     fvals = []
     for i_par in range(0, len(result.profile_result.list[0])):
         if i_par in profile_indices:
-            tmp = np.array(
-                [result.profile_result.list[0][i_par].x_path[i_par, :],
-                 result.profile_result.list[0][i_par].ratio_path[:]])
+            tmp = np.array([result.profile_result.list[profile_list][
```
yannikschaelte: sorry, but this code looks sooo autoformatted :)

pypesto/visualize/waterfall.py
```diff
@@ -208,6 +208,9 @@ def get_fvals(result, scale_y, offset_y, start_indices):
 
     fvals: ndarray
         function values
+
+    offset_y:
+        offset for the y-axis, if this is supposed to be in log10-scale
```
yannikschaelte:
```suggestion
        offset for the y-axis, if it is supposed to be in log10-scale
```

pypesto/visualize/waterfall.py
```diff
@@ -265,6 +268,9 @@ def handle_options(ax, max_len_fvals, ref, y_limits):
     y_limits: float or ndarray, optional
         maximum value to be plotted on the y-axis, or y-limits
 
+    offset_y:
+        offset for the y-axis, if this is supposed to be in log10-scale
```
yannikschaelte:
```suggestion
        offset for the y-axis, if it is supposed to be in log10-scale
```

pypesto/visualize/profiles.py
paulstapor: puh... right, that's not finished, I decided to give up on this... 😅 
The idea is to have a list of axis handles for a profiling plot, which on can pass to another call of the profiling plot... There are arbitrarily many problems with that, however... E.g. what, if you want to plot for another number of profiles than you have entries in the handle list? How to make sure that these profiles, which are plotted in the same axes, really belong to the same parameter? Etc. etc...

pypesto/visualize/profiles.py
paulstapor: yes, arbitrarily ugly...
will change that...

pypesto/visualize/profiles.py
yannikschaelte: if it's not used yet and maybe too difficult to implement, I'd just remove it?

paulstapor: Ah, totally got your comment wrong, now I see!
Well, okay, that'll be done in 5 minutes

PR #142, opened 2019-03-04
Fix readme

yannikschaelte:


yannikschaelte:
looks like codacy was just very very slow. will redraw this one. hope it works.

PR #138, opened 2019-03-03
correctly translate parameter mapping to amici plist

FFroehlich:
probably not properly implemented yet, waiting for petab fixes for easier debugging/testing

PR #145, opened 2019-03-07
Feature multithreading

FFroehlich:
enables multithreading in amici_objective, will require amici 0.10.2 to work


pypesto/objective/amici_objective.py
```diff
@@ -25,6 +25,7 @@ def __init__(self,
                  mapping_par_opt_to_par_sim=None,
                  mapping_scale_opt_to_scale_sim=None,
                  guess_steadystate=True,
+                 threads=1,
```
yannikschaelte:
```suggestion
                 n_threads: int = 1,
```
because otherwise in the class it will look like this holds a list of thread objects

pypesto/objective/amici_objective.py
```diff
@@ -68,6 +69,10 @@ def __init__(self,
             Whether to guess steadystates based on previous steadystates and
             respective derivatives.
 
+        threads: int, optional (default = 1(
```
yannikschaelte:
```suggestion
        threads: int, optional (default = 1)
```

pypesto/objective/amici_objective.py
```diff
@@ -68,6 +69,10 @@ def __init__(self,
...
+            Number of threads that are used for parallelization over
+            experimental conditions
```
yannikschaelte:
```suggestion
            experimental conditions.
```

FFroehlich:
waiting for #138 as that code will also need to be adapted

PR #146, opened 2019-03-07
update pypesto version

FFroehlich:
would like to push for a new realease including #138 and #145, ideally also #140 and #145


paulstapor:
Hope finishing my stuff this weekend...

PR #149, opened 2019-03-08
Improvements steadystate guesses

FFroehlich:
- resets steadystate guesses for each minimize call (propagates through aggregated objective)
- rejects linear updates according to sensitivities if they exceed 20% of the steadystates (which inherently prevents negative initial states)


FFroehlich:
One thing that I just realised which might be problematic in the current implementation is that steadystate guesses do not account for parameter dependent total abundances. If parameters are updated, the actual total abundances of conserved species might be different compared to the previous parameter values. This means that we would probably not simulate the ODE that we were intending to solve.

FFroehlich:
> One thing that I just realised which might be problematic in the current implementation is that steadystate guesses do not account for parameter dependent total abundances. If parameters are updated, the actual total abundances of conserved species might be different compared to the previous parameter values. This means that we would probably not simulate the ODE that we were intending to solve.

This issue needs to be solved in AMICI, not pypesto, so think we are good to merge this now and do the 0.0.5 release.

.gitignore
```diff
@@ -100,3 +100,4 @@ venv/*
 tmp/*
 */amici_models/*
 *.txt
+test/doc/example/tmp/benchmark-models/
```
yannikschaelte: this directory should not actually be created when the tests are run from the base directory, because the menchmark-models are downloaded do ``pypesto/doc/example/tmp/benchmark-models``, not the test subfolder. but can't be wrong to catch this one as well.

pypesto/objective/aggregated.py
```diff
@@ -189,6 +189,11 @@ def aggregate_hess(self, x):
     def aggregate_hessp(self, x):
         return sum(objective.hessp(x) for objective in self.objectives)
 
+    def reset_steadystate_guesses(self):
```
yannikschaelte: not nice but should work. steady states are only for AmiciObjective, right? could you add a comment here?

pypesto/objective/amici_objective.py
```diff
@@ -429,9 +427,12 @@ def apply_steadystate_guess(self, condition_ix, x):
             if guess_data['x_ss'] is not None:
                 x_ss_guess = guess_data['x_ss']
             if guess_data['sx_ss'] is not None:
-                x_ss_guess += guess_data['sx_ss'].transpose().dot(
+                linear_update = guess_data['sx_ss'].transpose().dot(
                     (x_sim - guess_data['x'])
                 )
+                # limit linear updates to max 20 % elementwise change
+                if (x_ss_guess/linear_update).max() < 0.2:
```
yannikschaelte: is there a reason why 20%? that looks rather arbitrary, but if it works better then ok.

.travis.yml
```diff
@@ -31,7 +31,7 @@ install:
 # run tests
 script:
   - python3 -m flake8 --exclude=build,doc,example,tmp,amici_models
-  - xvfb-run -a python3 -m pytest --cov=pypesto ./test
+  - travis_wait 20 xvfb-run -a python3 -m pytest --cov=pypesto ./test
```
paulstapor: This might save some rerunning... :+1:

pypesto/objective/amici_objective.py
```diff
@@ -152,11 +155,15 @@ def __init__(self,
...
+            if self.amici_model.ncl() > 0:
+                raise ValueError('Steadystate prediciton is not supported for'
+                                 'models with conservation laws!')
+
             if self.amici_model.getSteadyStateSensitivityMode() == \
                     amici.SteadyStateSensitivityMode_simulationFSA:
                 raise ValueError('Steadystate guesses cannot be enabled when'
                                  ' `simulationFSA` as '
-                                 'SteadyStateSensitivityMode')
+                                 'SteadyStateSensitivityMode!')
```
paulstapor: I'm sorry, I don't quite get the exception, which should be raised here... When should this occur? Maybe a little comment might help...

.gitignore
FFroehlich: I consistently see this directory ...

pypesto/objective/amici_objective.py
FFroehlich: its arbitrary, we do not have access to second order terms so we cannot judge the approximation quality. This definitely also ensures that initial states remain positive ;)

pypesto/objective/amici_objective.py
FFroehlich: simulationFSA relies on correct initial state sensitivities, which we do not provide here.

.gitignore
yannikschaelte: might be that you run tests differently (maybe ``python -m pytest`` from the test subdirectory? Anyway, then let' include it here.

pypesto/objective/aggregated.py
FFroehlich: done

PR #151, opened 2019-03-08
implement startpoint ordering, fixes #76

### LINKED ISSUE #76: Sorting initial parameter by objective function value
Should we sort startpoints according to their initial objective function value and start running estimation beginning at startpoint with lowest value?

Could improve results when working with a fixed computational budget.

FFroehlich:


test/test_sbml_conversion.py
```diff
@@ -92,10 +92,14 @@ def parameter_estimation(
...
 
-    optimize_options = pypesto.OptimizeOptions(allow_failed_starts=False)
+    optimize_options = pypesto.OptimizeOptions(
+        allow_failed_starts=False,
+        startpoint_resample=True,
+    )
 
     results = pypesto.minimize(
-        problem, optimizer, n_starts, options=optimize_options)
+        problem, optimizer, n_starts, options=optimize_options,
```
paulstapor: The comma here looks weird... Is it really supposed to be?

yannikschaelte: that's totally allowed in python

FFroehlich: I would actually argue that it's good style because it makes things like reordering arguments and extension of the arguments a little bit easier.

PR #166, opened 2019-04-10
Fix mapping in add_sim_hess_to_opt_hess

LeonardSchmiester:
Tried to optimize the Boehm model from the benchmark collection and got an out of index error in line 677 in `amici_objective`, because the `par_opt_id` did not skip over numeric values ('sim_hess' has dimensions 9x9 and `par_sim_idx` was out of bounds). 

Tested this on the Boehm model and seems to work

### DIFF
--- a/pypesto/objective/amici_objective.py
+++ b/pypesto/objective/amici_objective.py
@@ -651,7 +651,6 @@ def add_sim_hess_to_opt_hess(par_opt_ids,
     Same as for add_sim_grad_to_opt_grad, replacing the gradients by hessians.
     """
 
-    # https://github.com/ICB-DCM/AMICI/issues/274
     par_sim_idx = 0
     for par_opt_id in mapping_par_opt_to_par_sim:
         if not isinstance(par_opt_id, str):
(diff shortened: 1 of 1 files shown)


yannikschaelte:
@FFroehlich You documented that the dimension reduction is only applied to the second dimension, but now it's also applied to the first, or is it something different?

pypesto/objective/amici_objective.py
```diff
@@ -653,7 +653,8 @@ def add_sim_hess_to_opt_hess(par_opt_ids,
 
     # use enumerate for first axis as plist is not applied
```
FFroehlich:
```suggestion
```

pypesto/objective/amici_objective.py
```diff
@@ -653,7 +653,8 @@ def add_sim_hess_to_opt_hess(par_opt_ids,
...
     # https://github.com/ICB-DCM/AMICI/issues/274
```
FFroehlich:
```suggestion

```

PR #178, opened 2019-08-22
Feature logo

### LINKED ISSUE #173: Logo
create a logo for (py)PESTO

yannikschaelte:
Add logo in various variants

fixes #173


README.md
```diff
@@ -1,5 +1,7 @@
 # pyPESTO - Parameter EStimation TOolbox for python
 
+<image src="https://raw.githubusercontent.com/ICB-DCM/pyPESTO/master/doc/logo/logo_wordmark.svg" width="50%" alt="pyPESTO logo"/>
```
dweindl:
```suggestion
<img src="https://raw.githubusercontent.com/ICB-DCM/pyPESTO/master/doc/logo/logo_wordmark.svg" width="50%" alt="pyPESTO logo"/>
```

Check https://github.com/ICB-DCM/pyPESTO/tree/feature_logo. that it is shown properly

yannikschaelte: unfortunately, in the feature branch it cannot be shown correctly yet, because it references a file on the master branch (absolute path). I inserted the absolute path here (to raw.githubusercontent), because with relative links the pypi rendering would fail.

so once the changes are on master, it will show correctly. if there's a better way, welcome :)

dweindl: I see. Yeah, it's annoying to get it working locally offline, on github, pypi, etc... okay then

yannikschaelte:
> > what's the difference between logo and logo_quadratic? if they are different, one svg is missing
> 
> true. will add svg. quadratic is a quadratic version (nerd face). that's because often the logo is required to be in that shape, with even side lengths.

deleted quadratic version now since no need atm, and margins would need to be adapted

PR #179, opened 2019-08-28
Fix petab

yannikschaelte:
Adapt to changes of petab, in particular the parameter and scale mapping, and preequilibration.

Currently, the amici objective was not touched. That means that there everything still works based on indices, assuming we know the order of amici simulation parameters fixed (which is the case). The new mapping from dicts to lists happens in the petab_importer already. Dont' know if there is a "nicer" way, but this seems the easiest one to me.

### DIFF
--- a/pypesto/objective/petab_import.py
+++ b/pypesto/objective/petab_import.py
@@ -638,14 +638,12 @@ def _merge_preeq_and_sim_pars(parameter_mappings, scale_mappings):
 
     Parameters
     ----------
-
     parameter_mappings, scale_mappings: list of tuple of dict
         As returned by petab.get_optimization_to_simulation_parameter_mapping
         and petab.get_optimization_to_simulation_scale_mapping.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


pypesto/objective/petab_import.py
```diff
@@ -621,6 +630,66 @@ def _find_model_name(output_folder):
     return os.path.split(os.path.normpath(output_folder))[-1]
 
 
+def _merge_preeq_and_sim_pars(parameter_mappings, scale_mappings):
```
dweindl: Could be implemented in PEtab, but okay for now.

dweindl: Do we have a policy on type hints in pyPESTO? Should be formulated at some point...

yannikschaelte: yes, so far we have "ok" type hints often in the docstring, but these do not have to be exact. we should use type hints more, definitely.

PR #185, opened 2019-10-07
Feature constant species

paulstapor:
In the long run, pyPESTO's petab-importer should be replaced more and more by the one from Amici...
This is the first (very barebones) step for it. Now pyPESTO should be able to import PEtab models with constant species. I don't add a unit test for this, as this functionality is tested within AMICI.

### DIFF
--- a/.travis_pip_reqs.txt
+++ b/.travis_pip_reqs.txt
@@ -4,5 +4,5 @@ coverage
 flake8
 dlib
 gitpython
-petab>=0.0.0a6
-amici>=0.9.2
+petab>=0.0.0a12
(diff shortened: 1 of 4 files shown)


paulstapor:
Stupid question: How do we want to do it for merging?
It can currently not pass unit tests, as those would require amici 0.10.13, which isn't out yet...
Merge anyway? Wait for it?

FFroehlich:
> Should wait for that to happen.

dweindl:
@paulstapor : AMICI 0.10.13 should be available on pypi within the next hour.

PR #186, opened 2019-10-11
prepare release 0.0.9

yannikschaelte:


yannikschaelte:
@paulstapor could you shortly check/edit the releasenotes to what you changed? then we are fit to create a new release.

PR #190, opened 2019-10-20
pypesto.PetabImporter: Enable passing extra AMICI arguments for model…

### LINKED ISSUE #188: PEtab import add option allow_reinit_fixpar_initcond
Currently, for SBML/PEtab import the AMICI SBML import is called with `allow_reinit_fixpar_initcond=True` (=default). There should be an option to change that, as for large models, this takes a huge amount of time. Potentially this could be switched off by default in pyPESTO.

dweindl:
… generation (Closes #188)

### DIFF
--- a/pypesto/objective/petab_import.py
+++ b/pypesto/objective/petab_import.py
@@ -80,7 +80,7 @@ class PetabImporter:
             output_folder=output_folder,
             model_name=model_name)
 
-    def create_model(self, force_compile=False):
+    def create_model(self, force_compile=False, *args, **kwargs):
         """
         Import amici model. If necessary or force_compile is True, compile
         first.
@@ -93,10 +93,11 @@ class PetabImporter:
             exist yet. If True, the output folder is deleted and the model
             (re-)compiled in either case.
 
+            .. warning::
+                If `force_compile`, then an existing folder of that name will
+                be deleted.
 
-        .. warning::
-            If `force_compile`, then an existing folder of that name will be
-            deleted.
+        args, kwargs: Extra arguments passed to amici.SbmlImporter.sbml2amici
         """
         # courtesy check if target not folder
         if os.path.exists(self.output_folder) \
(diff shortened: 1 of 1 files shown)


pypesto/objective/petab_import.py
```diff
@@ -93,10 +93,11 @@ def create_model(self, force_compile=False):
             exist yet. If True, the output folder is deleted and the model
             (re-)compiled in either case.
 
-
         .. warning::
             If `force_compile`, then an existing folder of that name will be
             deleted.
+
+        args, kwargs: Extra arguments passed to amici.SbmlImporter.sbml2amici
```
yannikschaelte: These should belong in the parameters section.

pypesto/objective/petab_import.py
```diff
@@ -197,7 +203,8 @@ def compile_model(self):
             observables=observables,
             constantParameters=constant_parameter_ids,
             sigmas=sigmas,
-            noise_distributions=noise_distrs
+            noise_distributions=noise_distrs,
+            *args, **kwargs
```
yannikschaelte: one thing we would need to keep in mind: it may happen that there are other optional parameters which are not supposed to be forwarded to amici. but fine for the moment. whata is amici's behavior about arguments: will it raise if it gets an argument it does not understand?

pypesto/objective/petab_import.py
dweindl: Hmm, shouldn't the warning be part of the respective parameter section?

yannikschaelte: not sure if this can be interpreted correctly. maybe try to build the documentation using ``cd doc; make html; firefox _build/html/index.html`` and see how it looks.

pypesto/objective/petab_import.py
dweindl: Yes, any additional arguments will have to be put there explicitly. In that call tree, I didn't see any function with a large number of arguments where a catch-all might be more appropriate than here. Can be adapted once that changes.

Yes, it will raise if there are non-conforming arguments.

pypesto/objective/petab_import.py
dweindl: Thanks. Properly indented, but kept order.

PR #194, opened 2019-10-22
Only call add_sim_hess_to_opt_hess, if sensi_order>1. Fixes #193

### LINKED ISSUE #193: Calling pyPESTO objective with sensi_orders=(0,1) fails when using adjoints
When running:
```
importer = pypesto.PetabImporter(petab_problem)
obj = importer.create_objective()
obj.amici_solver.setSensitivityMethod(2)
obj(petab_problem.x_nominal, sensi_orders=(0,1))
```
I get the error message:
```
  File "<stdin>", line 1, in <module>
  File "/home/icb/leonard.schmiester/anaconda3/envs/speedy5/lib/python3.7/site-packages/pypesto/objective/objective.py", line 219, in __call__
    result = self._call_unprocessed(x, sensi_orders, mode)
  File "/home/icb/leonard.schmiester/anaconda3/envs/speedy5/lib/python3.7/site-packages/pypesto/objective/objective.py", line 239, in _call_unprocessed
    result = self._call_mode_fun(x, sensi_orders)
  File "/home/icb/leonard.schmiester/anaconda3/envs/speedy5/lib/python3.7/site-packages/pypesto/objective/objective.py", line 251, in _call_mode_fun
    result = self.fun(x, sensi_orders)
  File "/home/icb/leonard.schmiester/anaconda3/envs/speedy5/lib/python3.7/site-packages/pypesto/objective/amici_objective.py", line 193, in fun
    return self._call_amici(x, sensi_orders, MODE_FUN)
  File "/home/icb/leonard.schmiester/anaconda3/envs/speedy5/lib/python3.7/site-packages/pypesto/objective/amici_objective.py", line 316, in _call_amici
    coefficient=-1.0
  File "/home/icb/leonard.schmiester/anaconda3/envs/speedy5/lib/python3.7/site-packages/pypesto/objective/amici_objective.py", line 675, in add_sim_hess_to_opt_hess
    coefficient * sim_hess[par_sim_idx, par_sim_idx_2]
TypeError: 'NoneType' object is not subscriptable
```

Adding `if sensi_order > 1` in line 309 of the `amici_objective` should fix this issue I think

LeonardSchmiester:
Tested the fix with Boehm model from the benchmark collection


FFroehlich:
This prevents ever using the FIM, no? Because with `sensi_order==2`, you would get the true hessian, not the FIM.

yannikschaelte:
Well, at the moment 2nd order is not implemented anyway, so whenever 2nd order is requested, fim is what they get. In the future, I would rather say there should be an option whether to compute 2nd order via fim or elsewise, and only return it indeed when sensi_order==2 was requested.

FFroehlich:
Just rechecked the code, I am pretty sure that we return the FIM for `sensi_order==1`, so this change effectively prevents that.

yannikschaelte:
yes, previously the FIM was returned for sensi_order == 1 (ok, not sure if gets deleted at some point, or also ends up in the final output). but what was the reason behind that (I don't remember)? and what should be done then when using adjoints?

FFroehlich:
See https://github.com/ICB-DCM/pyPESTO/blob/master/pypesto/objective/amici_objective.py#L249

Cant we just add a check for sensitivity method?

yannikschaelte:
sounds reasonable. I did not check, but for  ``sensi_orders=(0,1)`` the fim/hessian is only returnable in call mode ``return_dict=True``.

yannikschaelte:
Maybe even 2 fields ``FIM`` and ``HESS`` makes sense such that both can be returned alongside each other. At the moment, both would however have the same (FIM) content, so one might want to wait with that until true Hessians are available.

PR #197, opened 2019-11-25
fix doc errors

yannikschaelte:
* Fix documentation build errors.
* Test create docs on pull requests.
* Add favicon for docs.

Readthedocs complained about these things. On local build, I now have no more errors.


.travis.yml
```diff
@@ -20,11 +20,12 @@ addons:
       - libatlas-base-dev
       - lcov
       - swig3.0
+      - pandoc
 
 install:
   - mkdir -p ~/bin/ && ln -s /usr/bin/swig3.0 ~/bin/swig && export PATH=~/bin/:$PATH
   - pip3 install --upgrade setuptools wheel pkgconfig
-  - pip3 install --upgrade jupyter # for jupyter notebooks
+  - pip3 install --upgrade jupyter
```
dweindl: Why not put all of them into `travis_pip_reqs`?

.travis.yml
```diff
@@ -33,8 +34,9 @@ script:
   - python3 -m flake8 --exclude=build,doc,example,tmp,amici_models
   - travis_wait 20 python3 -m pytest --cov=pypesto ./test/test_*
   - travis_wait 20 xvfb-run -a python3 -m pytest --cov=pypesto --cov-append ./test/visualize/test_*
-  - coverage xml
+  - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then sphinx-build -W -b html -d doc/_build/html doc/ doc/_build/html; fi
```
dweindl: Can that go into the sphinx Makefile?

yannikschaelte: why?

dweindl: because `make html` is way more readable

yannikschaelte: yeah but I would leave the default options in make, s.t. the docs do not error upon minor inconsistencies (which are turned into errors here via -W). but feel free to edit it, possibly with options.

dweindl: `make travis` `make rtd` ... ?

PR #243, opened 2020-03-25
Document objective function type

yannikschaelte:


pypesto/profile/__init__.py
```diff
@@ -4,13 +4,8 @@
 
 """
 
-from .profile import (parameter_profile,
-                      ProfileOptions)
+from .profile import (
```
jvanhoefer: why new line here?

yannikschaelte: no reason really

PR #247, opened 2020-03-25
Parameter Mapping: Use amici.ParameterMapping

yannikschaelte:


pypesto/objective/amici_objective.py
```diff
@@ -507,33 +507,33 @@ def create_plist_from_par_opt_to_par_sim(mapping_par_opt_to_par_sim):
 
 def create_identity_parameter_mapping(
```
yannikschaelte: @FFroehlich maybe you can also directly implement your use case here, i.e. the most convenient way of generating a default mapping without petab. currently, fixed and free parameters are set on linear scale, preeq fixed parameters are assumed to be set already in edatas.not sure what makes the most sense.

FFroehlich: Looking at the code my only concern is that none of the fixedParameters are specified in the problem since most of them are condition specific parameters and are directly written to ExpData. I think the default that is currently implemented makes sense as is.

yannikschaelte: Not sure if I understand. By "currently implemented" you mean the implementation like from until like two weeks ago, i.e. only the dynamic parameter mapping sim_var should be set here, and the fixed simulation parameters sim_fix are not set here?

yannikschaelte: @FFroehlich implemented now accordingly. let me know if wrong ;)

PR #249, opened 2020-03-26
Feature inf visualization

DantongWang:


pypesto/visualize/clust_color.py
```diff
@@ -249,3 +249,41 @@ def assign_colors_for_result_list(num_results, colors=None):
                'or as one single [r, g, b, alpha] color.')
 
     return colors
+
+
+def delete_nan_inf(fvals, x=None):
+    """
+    Delete nan and inf values in fvals and the corresponding parameters.
```
yannikschaelte:
```suggestion
    Delete nan and inf values in fvals. If parameters `x` are passend, also the corresponding
    entries are deleted.
```

pypesto/visualize/waterfall.py
```diff
@@ -256,8 +259,8 @@ def get_fvals(result: Result,
     # reduce to indices for which the user asked
     fvals = fvals[start_indices]
 
-    # get the minimal value which should be plotted
-    min_val = np.min(fvals)
+    # get the minimal value which should be plotted, avoid the value being -inf or nan
+    min_val = np.nanmin(fvals[fvals != -np.inf])
```
yannikschaelte: is this necessary? nans+infs should have been removed already?

test/visualize/test_visualize.py
```diff
@@ -48,6 +48,18 @@ def create_optimization_result():
 
     return result
 
+def create_optimization_result_nan_inf():
+    # get result with only numbers
+    result = create_optimization_result()
+
+    # append nan and inf
+    optimizer_result = pypesto.OptimizerResult(fval=float('nan'), x=[float('nan'), float('nan')])
```
yannikschaelte:
```suggestion
    optimizer_result = pypesto.OptimizerResult(
        fval=float('nan'), x=[float('nan'), float('nan')])
```

test/visualize/test_visualize.py
```diff
@@ -48,6 +48,18 @@ def create_optimization_result():
...
+    result.optimize_result.append(optimizer_result=optimizer_result)
+    optimizer_result = pypesto.OptimizerResult(fval=-float('inf'), x=[-float('inf'), -float('inf')])
```
yannikschaelte:
```suggestion
    optimizer_result = pypesto.OptimizerResult(
        fval=-float('inf'), x=[-float('inf'), -float('inf')])
```

test/visualize/test_visualize.py
```diff
@@ -48,6 +48,18 @@ def create_optimization_result():
 
     return result
 
+def create_optimization_result_nan_inf():
```
yannikschaelte:
```suggestion
def create_optimization_result_nan_inf():
    """Create a result object containing nan and inf function values."""
```

test/visualize/test_visualize.py
```diff
@@ -174,6 +198,18 @@ def test_parameters():
...
+        # create the necessary results
+        result_1 = create_optimization_result_nan_inf()
+        result_2 = create_optimization_result_nan_inf()
+
+        # test a standard call
+        pypesto.visualize.parameters(result_1)
+
+        # test plotting of lists
+        pypesto.visualize.parameters([result_1, result_2])
+
```
yannikschaelte: great to have added tests already! can you add one more, which tests delete_nan_inf directly? just create e.g. function values ``[42, 1.5, np.nan, 67.01, np.inf]`` and then ``assert [42, 1.5, 67.01] == delte_nan_inf(...)``

paulstapor: > looks great. @paulstapor fine for you?

gimme a bit time. Will look at it today :)
First look seems nice, tests still need to pass.

pypesto/visualize/waterfall.py
paulstapor: Agree with Yannik. `nanmin ` should not be necessary any more...

test/visualize/test_visualize.py
```diff
@@ -119,6 +131,18 @@ def test_waterfall():
         # test plotting of lists
         pypesto.visualize.waterfall([result_1, result_2])
 
+    @staticmethod
+    def test_waterfall_with_nan_inf():
+        # create the necessary results
+        result_1 = create_optimization_result_nan_inf()
+        result_2 = create_optimization_result_nan_inf()
```
paulstapor: Doe the test also work if one result has nans/infs, and the other one doesn't?
Maybe briefly add such a test as well... :)

pypesto/visualize/waterfall.py
DantongWang: but if it is -inf, then min(fvals) will be -inf

yannikschaelte:
@DantongWang there were some updates necessary, now all other tests should work again ;) https://travis-ci.com/github/ICB-DCM/pyPESTO/builds/155521384

pypesto/visualize/waterfall.py
yannikschaelte: but -np.inf should also not happen, right? nan+ plus/minus infs have been removed in delete_nan_inf?

yannikschaelte: ah ok sorry, you are right (I think). delete_nan_inf is not called before in this function.

PR #254, opened 2020-03-27
Remove circular dependence history-objective

### LINKED ISSUE #253: Remove history.obj
It would be cleaner and avoid the circular reference (potentially leading to copying problems) if the ObjectiveHistory did not hold a reference to the objective function fun method. Can this be circumvented?

yannikschaelte:
fixes #253

### DIFF
--- a/test/test_amici_objective.py
+++ b/test/test_amici_objective.py
@@ -63,7 +63,7 @@ class AmiciObjectiveTest(unittest.TestCase):
             problem=problem, optimizer=optimizer, n_starts=2,
         )
 
-        self.assertTrue(obj.steadystate_guesses['fval'] < np.inf)
+        self.assertTrue(problem.objective.steadystate_guesses['fval'] < np.inf)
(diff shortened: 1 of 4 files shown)


test/test_amici_objective.py
```diff
@@ -63,7 +63,7 @@ def test_preeq_guesses(self):
             problem=problem, optimizer=optimizer, n_starts=2,
         )
 
-        self.assertTrue(obj.steadystate_guesses['fval'] < np.inf)
+        self.assertTrue(problem.objective.steadystate_guesses['fval'] < np.inf)
```
yannikschaelte: @FFroehlich if you used this in your code --- note that the objective is deep-copied when put into the problem. thus the original object does not hold any information afterwards.

FFroehlich: Should be fine, not currently used :)

yannikschaelte: Funnily, it was only filled due to a wrong copy action in the first place :see_no_evil:

PR #257, opened 2020-03-30
implement amici solver to/from hdf5 for pickling for PEtabObjective

### LINKED ISSUE #246: Different behaviour for singleCoreEngine and MultiProcessEngine
I tried optimizing the Salazar-Cavazos model from the benchmark collection with adjoint sensitivities for single and multi core engine. Optimization works with `singleCoreEngine` but fails with `MultiProcessEngine` because of AMICI integration errors. I get the same errors when using `singleCoreEngine` and forward sensitivities. Might it be, that `MultiProcessEngine` ignores the sensitivity method set in `obj.amici_solver.setSensitivityMethod`?

yannikschaelte:
fixes #246

### DIFF
--- a/pypesto/objective/petab_import.py
+++ b/pypesto/objective/petab_import.py
@@ -371,6 +371,10 @@ class PetabAmiciObjective(AmiciObjective):
                 {'amici_model', 'amici_solver', 'edatas'}:
             state[key] = self.__dict__[key]
 
+        amici_solver_file = tempfile.mkstemp()[1]
+        amici.writeSolverSettingsToHDF5(self.amici_solver, amici_solver_file)
+        state['amici_solver'] = amici_solver_file
+
         return state
 
     def __setstate__(self, state: Dict) -> None:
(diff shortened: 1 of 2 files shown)


dweindl:


```diff

Clones added
============
- test/test_engine.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dweindl/pyPESTO/pullRequest?prid=5270212&bid=17263225)

pypesto/objective/petab_import.py
```diff
@@ -371,6 +371,10 @@ def __getstate__(self) -> dict:
                 {'amici_model', 'amici_solver', 'edatas'}:
             state[key] = self.__dict__[key]
 
+        amici_solver_file = tempfile.mkstemp()[1]
+        amici.writeSolverSettingsToHDF5(self.amici_solver, amici_solver_file)
+        state['amici_solver'] = amici_solver_file
```
FFroehlich: is possible to store that file as bytestream in the state dict? In current implementation removing the file will make it impossible to unpickle the same pickle twice and make it really hard to move pickles across machines (need to check amici versions anyways). Nothing of immediate concern, but may become relevant in the future.

yannikschaelte: true. the current implementation in amici requires a file to write to, right? so, currently, one would need to write the file and directly read it in again and save the bytes in the state dict. is that what you mean?

PR #275, opened 2020-04-10
Fix covariance regularization; Fix namespace sampling-sample

yannikschaelte:
* Was not ensured to yield positive definite matrices, thus leading to errors in sampling sometimes. Now it should yield a minimum diagonal edit positive definite matrix.
* Fix namespace of sampling vs method name sample.


test/test_sampling.py
```diff
@@ -136,3 +136,15 @@ def test_multiple_startpoints():
...
+def test_regularize_covariance():
+    """
+    Make sure that `regularize_covariance` renders symmetric matrices
+    positive definite.
+    """
+    matrix = np.array([[-1., -4.], [-4., 1.]])
+    assert np.any(np.linalg.eigvals(matrix) < 0)
+    reg = pypesto.sampling.adaptive_metropolis.regularize_covariance(
+        matrix, 1e-6)
+    assert np.all(np.linalg.eigvals(reg) >= 0)
```
dilpath, line 150: Strictly > 0?

yannikschaelte: it will be >0 in the current implementation, but for the normal distribution >= 0 should be enough.

PR #293, opened 2020-04-28
Unpack model selection parameter definitions.

dilpath:
Just merging for now, this will be cleaned up when these functions are moved into appropriate classes/files with other work.

### DIFF
--- a/pypesto/modelselection/unique_models.py
+++ b/pypesto/modelselection/unique_models.py
@@ -1,0 +1,111 @@
+import itertools
+import tempfile
+from typing import List, Tuple
+
+# Zero-indexed column indices
+MODEL_NAME_COLUMN = 0
+SBML_FILENAME_COLUMN = 1
+PARAMETER_DEFINITIONS_START = 2
+HEADER_ROW = 0
+PARAMETER_VALUE_DELIMITER = ';'
+ESTIMATE_SYMBOL_UI = '-'
+# here, 'nan' is a string as it will be written to a file. The actual internal
+# symbol is float('nan')
+ESTIMATE_SYMBOL_INTERNAL = 'nan'
+
+#def split_file_definition(line: str) -> Tuple:
+#    '''
+#    Returns a 3-tuple that contains the model name, SBML file name, and all
+#    possible parameter values as a list (parameters) of tuples (parameter
+#    values).
+#    '''
+#    columns = line.strip().split('\t')
+#    return (columns[MODEL_NAME_COLUMN],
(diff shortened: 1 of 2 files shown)


pypesto/modelselection/unique_models.py
```diff
@@ -0,0 +1,111 @@
...
+                    expanded_models_file.write(
+                        '\t'.join([
+                            columns[MODEL_NAME_COLUMN]+f'_{index}',
+                            columns[SBML_FILENAME_COLUMN],
+                            *selection
+                        ]) + '\n'
+                    )
+            else:
+                expanded_models_file.write(line)
+    return expanded_models_file
```
MerktSimon, line 92: just for understanding: this now returns a file which has the model selection in a tab separated format and could be read by pandas data frame e.g. and loop over rows which our row2problem function could then convert to pypesto problems.
Correct?

dilpath: Yes the example:
| modelId | SBML      | parameterId1 | parameterId2 |
|---------|-----------|--------------|--------------|
| M1      | SBML1.xml | 0            | 0            |
| M2      | SBML1.xml | 1            | 0            |
| M3      | SBML2.xml | 0;1;-        | 0;1          |

is converted to:
| modelId | SBML      | parameterId1 | parameterId2 |
|---------|-----------|--------------|--------------|
| M1_0    | SBML1.xml | 0            | 0            |
| M2_0    | SBML1.xml | 1            | 0            |
| M3_0    | SBML2.xml | 0            | 0            |
| M3_1    | SBML2.xml | 0            | 1            |
| M3_2    | SBML2.xml | 1            | 0            |
| M3_3    | SBML2.xml | 1            | 1            |
| M3_4    | SBML2.xml | -            | 0            |
| M3_5    | SBML2.xml | -            | 1            |

and the temporary file object is returned. Do you think a data frame should be returned instead? If so, then the entire table would be stored in memory (at the moment, we intend to implement a generator, which would use less memory, but may be slower...).

MerktSimon: I think storing/returning in a (temp) file is fine. I would have just continued with reading this in as a dataframe since I find it easier to iterate over row while still keeping the head row.
But I guess there are mulitple ways to do this. Should we just discuss tomorrow morning?

dilpath: Sure, at the moment the generator will return each row as a dictionary, where the keys are the header elements, and the values are the row elements.

PR #302, opened 2020-04-29
Feature qualitative

LeonardSchmiester:


pypesto/hierarchical/optimal_scaling_solver.py
```diff
@@ -0,0 +1,299 @@
...
+STANDARD = 'standard'
+
+
+class OptimalScalingInnerSolver(InnerSolver):
+    """
+    Solve the inner subproblem of the
+    optimal scaling approach for ordinal data.
+    """
+
+    def __init__(self, n_starts: int = 1, n_records: int = 1,
```
yannikschaelte, line 19: document what options are possible

pypesto/hierarchical/optimal_scaling_solver.py
```diff
@@ -0,0 +1,299 @@
...
+        # x_names = [x.id for x in pars]
+        # data = problem.data
+
+        optimal_surrogate = \
+            compute_optimal_surrogate_data(problem, sim, self.options)
+
+        return optimal_surrogate
+
+    @staticmethod
+    def calculate_obj_function(x_inner_opt):
```
yannikschaelte, line 52: for the below functions, document purpose and arguments

pypesto/hierarchical/parameter.py
```diff
@@ -10,6 +10,7 @@ class InnerParameter:
     SCALING = 'scaling'
     OFFSET = 'offset'
     SIGMA = 'sigma'
+    OPTIMALSCALING = 'optimalScaling'
```
yannikschaelte, line 13: in what way is this different from the scaling? can one indicate that this is based on qualitative data -> `qualitative_scaling`?

LeonardSchmiester: The method is called optimal scaling. But also fine for me to change the name here, as its just about the parameter

yannikschaelte: but this "optimalScaling" is supposed to imply the qualitative aspect already, right? then probably better to include that in the name. if "scaling" can be used here (I guess not, because the method needs to know that it needs to apply the qualitative solver), also fine.

LeonardSchmiester: Any keyword would work for the method. Makes sense to change it to something clearly indicating, that its about qualitative data :+1:

LeonardSchmiester: Changed it to `qualitative_scaling`

Issue #301, opened 2020-04-29
Model selection: Convert table row to pypesto problem


MerktSimon:
- [x] write conversion function for input petab problem + parameter iterable

- [x] adjust for input from #297 

- [x] write test

PR #303, opened 2020-04-29
Feature modelselection row2problem

MerktSimon:
- `row2problem()` expects currently the "table-row" to be a `pandas.Series` object which has to be discussed 
- Also it loads petab problem from the file named in `SBML` column unless handed over explicitly
- tests are turned of since there seems to be an issue with getting objective for example
- the 2 constants should probably be put in some general constant file together with all other constants

refers to issue #301


pypesto/modelselection/row2problem.py
```diff
@@ -0,0 +1,57 @@
...
+
+def row2problem(row: pd.Series,
+                petab_problem: Union[petab.Problem, str] = None,
+                obj: Objective = None) -> Problem:
+    """
+    Create a pypesto.Problem from a single, unambiguous model selection row.
+    Optional petab.Problem and objective function can be provided to overwrite
+    model selection yaml entry and default PetabImporter objective
+    respectively.
+    """
```
yannikschaelte, line 24:
```suggestion
def row2problem(row: dict,
                petab_problem: Union[petab.Problem, str] = None,
                obj: Objective = None) -> Problem:
    """
    Create a pypesto.Problem from a single, unambiguous model selection row.
    Optional petab.Problem and objective function can be provided to overwrite
    model selection yaml entry and default PetabImporter objective
    respectively.
    
    Parameters
    ----------
    row:
        A single, unambiguous model selection row.
    petab_problem:
        The petab problem for which to perform model selection.
    obj:
        The objective to modify for model selection.
    """
```

pypesto/modelselection/row2problem.py
```diff
@@ -0,0 +1,57 @@
...
+    # overwrite petab_problem by problem in case it refers to yaml
+    if petab_problem is None:
+        petab_problem = row[YAML_FILENAME_COLUMN].str
+    if isinstance(petab_problem, str):
+        petab_problem = petab.Problem.from_yaml(petab_problem)
+    importer = PetabImporter(petab_problem)
+
+    # drop row entries not referring to parameters
+    row.drop(labels=[YAML_FILENAME_COLUMN, MODEL_NAME_COLUMN], inplace=True)
```
yannikschaelte, line 33:
```suggestion
    for key in YAML_FILENAME_COLUMN, MODEL_NAME_COLUMN:
        row.drop(key)
```

test/test_modelselection.py
```diff
@@ -0,0 +1,37 @@
...
+def test_row2problem_yaml_string(yaml_file_example):
+    petab_problem = petab.Problem.from_yaml(yaml_file_example)
+
+    importer = PetabImporter(petab_problem)
+    obj = importer.create_objective()
+    pypesto_problem = importer.create_problem(obj)
+
+    row = pd.Series()
+    assert row2problem(row, yaml_file_example) == pypesto_problem
+
```
yannikschaelte, line 29: explicitly assert that `x_names`, `x_fixed_vals`, `x_fixed_ids` etc. are as expected.

pypesto/modelselection/row2problem.py
yannikschaelte: oh yes sorry, that's what I meant, not drop.

pypesto/modelselection/row2problem.py
```diff
@@ -0,0 +1,71 @@
...
+            # petab_problem.parameter_df.lowerBound.loc[par_id] = float("NaN")
+            # petab_problem.parameter_df.upperBound.loc[par_id] = float("NaN")
+        else:
+            petab_problem.parameter_df[ESTIMATE].loc[par_id] = 1
+            # petab_problem.parameter_df.nominalValue.loc[par_id] = float(
+            # "NaN")
+
+    # chose standard objective in case none is provided
+    importer = PetabImporter(petab_problem)
+    if Objective is None:
```
dilpath, line 67:
```suggestion
    if obj is None:
```

PR #326, opened 2020-05-13
Create required directories before creating hdf5 file (Fixes #323)

dweindl:
* Create directories as required
* Add h5py as optional dependency
* Update tests

PR #329, opened 2020-05-13
Speed up travis

yannikschaelte:
* don't run push checks on pr level too


test/test_storage.py
```diff
@@ -12,7 +12,7 @@
 
 def test_storage_opt_result():
     minimize_result = create_optimization_result()
-    with tempfile.TemporaryDirectory(dir=f".") as tmpdirname:
```
dweindl, line 15: Haha, also did that on #326

PR #330, opened 2020-05-13
Update CI system

### LINKED ISSUE #318: Update CI build
I think it makes sense to always test against the latest stuff (can ask users to upgrade, but not to downgrade :))

* [x] python version currently 3.6 (not supported by AMICI) -> 3.8?
* [x] test system ubuntu 16.04 -> 18.04? (or newer, once available)

yannikschaelte:
fixes #318


yannikschaelte:
> Seems like the python files should not be part of this PR. Approving CI updates.

ah right. those moved in here from some trial and error. but shouldn't change any logic, just one additional check @FFroehlich.

PR #334, opened 2020-05-14
Feature storage result

plakrisenko:
saving also x_free_indices for x_labels in parameters plot


plakrisenko:
> @plakrisenko can you check which attributes of `Problem` are not stored yet? All should be stored except objective, I would say ...

ok, will check

yannikschaelte:
> > @plakrisenko can you check which attributes of `Problem` are not stored yet? All should be stored except objective, I would say ...
> 
> ok, will check

I think there should be a test on this (ok, maybe not sufficiently, if the free indices were not captured).

PR #337, opened 2020-05-16
implement partial trace

yannikschaelte:
* allow requesting arbitrary indices of traces
* therefore, simplify getter functions

### DIFF
--- a/pypesto/objective/history.py
+++ b/pypesto/objective/history.py
@@ -698,55 +698,55 @@ class CsvHistory(History):
     def get_x_trace(
             self, ix: Union[int, Sequence[int], None] = None
     ) -> Union[Sequence[np.ndarray], np.ndarray]:
-        return self._trace[X].values[ix]
+        return list(self._trace[X].values[ix])
 
     @trace_wrap
     def get_fval_trace(
             self, ix: Union[int, Sequence[int], None]
     ) -> Union[Sequence[float], float]:
-        return self._trace[(FVAL, np.nan)].values[ix]
+        return list(self._trace[(FVAL, np.nan)].values[ix])
 
     @trace_wrap
     def get_grad_trace(
             self, ix: Union[int, Sequence[int], None] = None
     ) -> Union[Sequence[np.ndarray], np.ndarray]:
-        return self._trace[GRAD].values[ix]
+        return list(self._trace[GRAD].values[ix])
 
     @trace_wrap
     def get_hess_trace(
             self, ix: Union[int, Sequence[int], None] = None
     ) -> Union[Sequence[np.ndarray], np.ndarray]:
-        return self._trace[(HESS, np.nan)].values[ix]
+        return list(self._trace[(HESS, np.nan)].values[ix])
 
     @trace_wrap
     def get_res_trace(
             self, ix: Union[int, Sequence[int], None] = None
     ) -> Union[Sequence[np.ndarray], np.ndarray]:
-        return self._trace[(RES, np.nan)].values[ix]
+        return list(self._trace[(RES, np.nan)].values[ix])
 
     @trace_wrap
     def get_sres_trace(
             self, ix: Union[int, Sequence[int], None] = None
     ) -> Union[Sequence[np.ndarray], np.ndarray]:
-        return self._trace[(SRES, np.nan)].values[ix]
+        return list(self._trace[(SRES, np.nan)].values[ix])
 
     @trace_wrap
     def get_chi2_trace(
             self, ix: Union[int, Sequence[int], None] = None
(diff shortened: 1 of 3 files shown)


pypesto/objective/history.py
```diff
@@ -14,6 +15,31 @@
...
 
 
+def trace_wrap(f):
+    """
+    Wrapper around trace getters that transforms input `ix` vectors to a valid
+    index list, and reduces for integer `ix` the output to a single value.
+    """
+    def wrapped_f(
+            self, ix: Union[Sequence[int], int, None] = None
+    ) -> Union[Sequence[Any], Any]:
```
FFroehlich, line 25: Can we be a bit more specific than `Any` here. Of course the function implements the most general case, but being so permissive may easily lead to false negatives in typechecking.

pypesto/objective/history.py
```diff
@@ -713,41 +694,59 @@ def _save_trace(self, finalize: bool = False):
...
-    def get_time_trace(self) -> Sequence[float]:
-        return self._trace[(TIME, np.NaN)].values
-
-    def get_time(self, ix) -> float:
-        return self._trace.loc[ix, (TIME, np.NaN)]
+    @trace_wrap
+    def get_x_trace(
+            self, ix: Union[int, Sequence[int], None] = None
+    ) -> Union[Sequence[np.ndarray], np.ndarray]:
+        return self._trace[X].values[ix]
```
FFroehlich, line 701: This will return `np.ndarray` and not `Sequence[np.ndarray]` right? I am surprised that tests are not failing. I believe this previously led to failures in visualization routines.

yannikschaelte: No tests failing. Do you remember which visualization routine? The different behavior of lists vs numpy arrays has led to problems in a few places already. Maybe it's safest to specify output as List here? Though it would be of interest what "List" vs "Sequence" specific behavior is used anywhere

FFroehlich: I think it was https://github.com/ICB-DCM/pyPESTO/blob/f74b4d0b54b2f159260664284da5e49709a8245a/pypesto/visualize/optimizer_history.py#L218

I would guess that `enumerate` is not equivalent for `np.ndarray` and `List`.

yannikschaelte: Hm. Enumerate should only require an Iterable, which both implement -- in both cases should loop over the first dimension. Anyway, implementing as a list should be fine, efficiency issues should be unlikely.

yannikschaelte: > This will return np.ndarray and not Sequence[np.ndarray] right? I am surprised that tests are not failing. I believe this previously led to failures in visualization routines.

Ok. It was meant to be a Sequence[np.ndarray], where the first dimension is just realized as a np.ndarray. I just realize that `issubclass(np.ndarray, collections.abc.Sequence)` is actually False. Don't really see the motivation, see also the thread here https://github.com/numpy/numpy/issues/2776.

yannikschaelte: > I think it was
> 
> https://github.com/ICB-DCM/pyPESTO/blob/f74b4d0b54b2f159260664284da5e49709a8245a/pypesto/visualize/optimizer_history.py#L218
> 
> I would guess that `enumerate` is not equivalent for `np.ndarray` and `List`.

Just seeing: In the current implementation it should be ndarrays, not lists, too. https://github.com/ICB-DCM/pyPESTO/blob/develop/pypesto/objective/history.py#L719, so I guess that's fine?

FFroehlich: Hmm, those are all the getters for individual entries though, the getters for full traces should return lists of `np.ndarray`s though.

FFroehlich:
added typecheck to test, should be good to go from my side.

FFroehlich:
> added typecheck to test, should be good to go from my side.

Although, looking at the test again, I think the typehints of some of the getters are wrong as they may return `np.NaN` instead of `np.ndarray`.

yannikschaelte:
> Although, looking at the test again, I think the typehints of some of the getters are wrong as they may return np.NaN instead of np.ndarray.

Right. This affects empty entries. What do we want to return in that case? The MemoryHistory should currently return `None`, the CsvHistory `np.nan`. We should unify this, to allow for better checking in visualization. Or one applies sth like `pd.isnull` which handles None and nan equally gracefully.

yannikschaelte:
Ah I see it is set to nan in history::extract_values already if it is missing.

yannikschaelte:
@FFroehlich just one incongruency remaining: CsvHistory seems to return schi2 = [nan, nan, ...] if only function value and gradient are provided, unlike MemoryHistory which returns nan.

PR #343, opened 2020-05-19
Fixes history issues with MODE_RES and aggregated objectives

FFroehlich:


dweindl:


```diff

Clones added
============
- test/test_history.py  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dweindl/pyPESTO/pullRequest?prid=5552517&bid=18205430)

yannikschaelte:
@jvanhoefer you are using AggregatedObjective now. Could you have a look?

pypesto/objective/aggregated.py
```diff
@@ -206,6 +161,48 @@ def reset_steadystate_guesses(self):
                 objective.reset_steadystate_guesses()
 
 
+def aggregate_results(rvals):
```
yannikschaelte, line 164: kinda obvious, but docstring nonetheless?

pypesto/objective/aggregated.py
```diff
@@ -206,6 +161,48 @@ def reset_steadystate_guesses(self):
...
+
+    # sum over fval/grad/hess
+    result = {
+        key: sum(rval[key] for rval in rvals)
+        for key in [FVAL, CHI2, SCHI2, GRAD, HESS, HESSP]
+        if rvals[0].get(key, None) is not None
```
yannikschaelte, line 170: only checking `rvals[0]` may be problematic (just an idea), e.g. if the first function succeeds but there is an error in some other function call, or if one function returns a value (e.g. hessian, this will definitely be the case for AmiciObjective+Prior), but another one not. in the case that only some functions return a value, the correct thing to do would be to set the value to nan/None I guess?

pypesto/objective/aggregated.py
```diff
@@ -206,6 +161,48 @@ def reset_steadystate_guesses(self):
...
+    }
+
+    # extract rdatas and flatten
+    result[RDATAS] = []
+    for rval in rvals:
+        if RDATAS in rval:
+            result[RDATAS].extend(rval[RDATAS])
```
yannikschaelte, line 177: should it be extend or append?

FFroehlich: extend. `AggregateObjective` is still an `Objective`, so it should follow the same output API and result[RDATAS] should be a list of ReturnData and not a list of lists.

pypesto/objective/aggregated.py
FFroehlich: The `__init__` routine checks whether all passed objectives have the support the same function calls, so this should be fine. I didn't change the implementation here, just moved it to a different function.

yannikschaelte: yes, i just wonder if the previous implementation is right. What happens if sensi orders (0,1) are asked for but objfun 0 (AmiciObjective) returns a Hessian approximation nonethelss and the objfun 1 (Prior) does not? Would this then throw an error here?

FFroehlich: Then thats an issue to fix in the respective Objective implementations, no? Maybe we should more clearly define the API of an objective (including implementing a ResultDict class that implements the decisions in code). My current understanding is that it always returns FVAL and everything else only if requested (which is indeed violated by some functions, see #344)

dweindl:


```diff

Clones added
============
- test/test_history.py  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dweindl/pyPESTO/pullRequest?prid=5552517&bid=18205430)

pypesto/objective/aggregated.py
yannikschaelte: Ah there was one misunderstanding on my side regarding that here fun/grad are called directly, which should indeed only return what is documented in the objective docstring, which is checked in the AggregatedObjective init function.

The best way to ensure this might be if one could get AggregatedObjective to only call `self.objectives[i].{__call__/_call_unprocessed/_call_mode{fun/res}}` directly, as this already handles the "returning only what was requested" ... :thinking: would this be possible? If things work the way they are implemented now, that should however also be fine I guess.

dweindl:


```diff

Clones added
============
- test/test_history.py  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dweindl/pyPESTO/pullRequest?prid=5552517&bid=18205430)

pypesto/objective/aggregated.py
FFroehlich: Yes, I wanted to change the implementation here and get rid of all the duplication in the `aggregate_...` funs (especially since those don't update histories etc), but that didn't turn out as simple as I thought, which was the primary reason I opened #344.

FFroehlich:
closing in favor of #347

PR #358, opened 2020-05-27
Changes to pymc3

lcontento:
* Added caching of the gradients so that the forward problem is only computed once
* If a `trace` has already been started, resume sampling from the last point (which I think is the behaviour for the other samplers?)

### DIFF
--- a/pypesto/sampling/pymc3.py
+++ b/pypesto/sampling/pymc3.py
@@ -135,21 +135,21 @@ class TheanoLogProbability(tt.Op):
 
     def __init__(self, problem: Problem, beta: float = 1.):
         self._objective: Objective = problem.objective
-
-        # initialize the log probability Op
-        self._log_prob = \
-            lambda x: - beta * self._objective(x, sensi_orders=(0,))
-
+        self._beta: float = beta
+        self._cached_theta = None
+        self._cached_dlogp = None
         # initialize the sensitivity Op
         if problem.objective.has_grad:
-            self._log_prob_grad = TheanoLogProbabilityGradient(problem, beta)
+            self._log_prob_grad = TheanoLogProbabilityGradient(self)
         else:
             self._log_prob_grad = None
 
     def perform(self, node, inputs, outputs, params=None):
         theta, = inputs
-        log_prob = self._log_prob(theta)
-        outputs[0][0] = np.array(log_prob)
+        logp, dlogp = self._objective(theta, sensi_orders=(0, 1))
+        outputs[0][0] = np.array(-self._beta * logp)
+        self._cached_theta = theta
+        self._cached_dlogp = dlogp
 
     def grad(self, inputs, g):
         # the method that calculates the gradients - it actually returns the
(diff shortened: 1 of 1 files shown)


dweindl:


```diff

Complexity increasing per file
==============================
- pypesto/sampling/pymc3.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dweindl/pyPESTO/pullRequest?prid=5594798&bid=18327862)

pypesto/sampling/pymc3.py
```diff
@@ -167,25 +167,23 @@ class TheanoLogProbabilityGradient(tt.Op):
...
+        self._beta = logp_op._beta
 
     def perform(self, node, inputs, outputs, params=None):
         theta, = inputs
-        # calculate gradients
-        log_prob_grad = self._log_prob_grad(theta)
-        outputs[0][0] = log_prob_grad
+
+        cached_theta = self._logp_op._cached_theta
+        if cached_theta is not None and np.all(theta == cached_theta):
```
FFroehlich, line 184:
```suggestion
        if cached_theta is not None and np.equal(theta, cached_theta):
```

FFroehlich, line 184:
```suggestion
        if cached_theta is not None and np.array_equal(theta, cached_theta):
```

pypesto/sampling/pymc3.py
```diff
@@ -167,25 +167,23 @@ class TheanoLogProbabilityGradient(tt.Op):
...
+        self._objective = logp_op._objective
+        self._beta = logp_op._beta
 
     def perform(self, node, inputs, outputs, params=None):
         theta, = inputs
-        # calculate gradients
-        log_prob_grad = self._log_prob_grad(theta)
-        outputs[0][0] = log_prob_grad
+
+        cached_theta = self._logp_op._cached_theta
```
yannikschaelte, line 183: sorry, in the meantime we had to change the file structure a bit, these two classes are now (slightly changed) in `theano.py`.

pypesto/sampling/pymc3.py
```diff
@@ -135,21 +135,21 @@ class TheanoLogProbability(tt.Op):
...
-            self._log_prob_grad = TheanoLogProbabilityGradient(problem, beta)
+            self._log_prob_grad = TheanoLogProbabilityGradient(self)
         else:
             self._log_prob_grad = None
 
     def perform(self, node, inputs, outputs, params=None):
         theta, = inputs
-        log_prob = self._log_prob(theta)
-        outputs[0][0] = np.array(log_prob)
+        logp, dlogp = self._objective(theta, sensi_orders=(0, 1))
```
yannikschaelte, line 149: I think it would make more sense to cache values from the gradient -> prob, not prob -> grad. this is because whenever we evaluate the gradient , amici also computes the function value, but not the other way round. thus, when using a gradient free sampler like `pypesto.Pymc3sample(step_function=pymc3.Slice)`, now always the gradient is calculated, which is however never required, if I see correctly.

lcontento: Indeed with this change the gradient is computed also for samplers which do not need it, so it is not ideal. But it looks like that the gradient is always computed after the function value, so if we cached inside the gradient there would be no cache hits I am afraid. Then, probably having two versions is the best way to do it.

dweindl:


```diff

Complexity increasing per file
==============================
- pypesto/sampling/pymc3.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dweindl/pyPESTO/pullRequest?prid=5594798&bid=18327862)

pypesto/sampling/pymc3.py
yannikschaelte: Agreeing on that. It might be difficult to find out whether gradients are required however. If nothing is specified, we try to use the Nuts sampler, with fallback to Slice if grad=NullType, i.e. when not objective.has_grad(). So I guess in that case one can just preemptively check objective.has_grad() and generate the log_prob accordingly. If a sampler is explicitly specified, this will not be so easy, one would need to figure out which samplers in pymc3 use gradients and which not ...

lcontento:
I'll rework the pull request against the new theano.py file tomorrow, taking care of the case of the derivative-free samplers and keeping the old code as it is for comparison purposes.

PR #361, opened 2020-05-28
fix minus sign in get_error_output

paulstapor:
It looks like the error output of the amici objective returned an NLLH value instead of an LLH value. Hence, if simulation failed, the nllh value was -infty, which no best parameter vector can ever beat, making simulation crashes highly desirable for the optimizer... :joy:

### DIFF
--- a/pypesto/objective/amici_util.py
+++ b/pypesto/objective/amici_util.py
@@ -225,7 +225,11 @@ def get_error_output(
         edatas: Sequence['amici.ExpData'],
         rdatas: Sequence['amici.ReturnData'],
         dim: int):
-    """Default output upon error."""
+    """Default output upon error.
+
+    Returns values indicative of an error, that is with nan entries in all
+    vectors, and a function value, i.e. nllh, of `np.inf`.
+    """
     if not amici_model.nt():
         nt = sum([data.nt() for data in edatas])
     else:
@@ -234,7 +238,7 @@ def get_error_output(
     n_res = nt * amici_model.nytrue
 
     return {
-        FVAL: - np.inf,
+        FVAL: np.inf,
         GRAD: np.nan * np.ones(dim),
         HESS: np.nan * np.ones([dim, dim]),
         RES:  np.nan * np.ones(n_res),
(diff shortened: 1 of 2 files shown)


pypesto/objective/amici_util.py
```diff
@@ -234,7 +234,7 @@ def get_error_output(
     n_res = nt * amici_model.nytrue
 
     return {
-        FVAL: np.inf,
+        FVAL: - np.inf,
```
yannikschaelte, line 237: on a second look, I think this was actually right, as the calculator is already supposed to return the nllh values (i.e. flipped sign).

yannikschaelte: however, https://github.com/ICB-DCM/pyPESTO/blob/develop/pypesto/objective/amici.py#L292 seems wrong as the `ret[FVAL] = nllh` already...

paulstapor: > however, https://github.com/ICB-DCM/pyPESTO/blob/develop/pypesto/objective/amici.py#L292 seems wrong as the `ret[FVAL] = nllh` already...

Right, the sign is flipped again in `call_unprocessed()`.
As this routine does the right job for non-failing simulations, I would guess we want to keep the minus sign there and just flip it in amici_util...

Or are there other opinions? Happy to discuss. However, this one needs to be merged soon, as it makes optimization impossible for quite some models atm...

yannikschaelte: @paulstapor might be good if the error can be reproduced in a unit test
@FFroehlich can you confirm whether there is a wrong sign?

FFroehlich: I think `call_unprocessed()` has an extraneous minus sign and `get_error_output` is missing one. Haven't used the steadystate guess code in a while, so I wouldn't be sure its currently doing the right thing ...

yannikschaelte: `get_error_output` is supposed to return the nllh value already. The worst possible value would be `+ np.inf`, which it should thus return, I think.

FFroehlich: you are right.

yannikschaelte:
@FFroehlich short look?

PR #366, opened 2020-05-29
progress bar for MCMC sampling

elbaraim:
Small pull request --> Added a nice progress bar to indicate sampling progress.


pypesto/sampling/sample.py
```diff
@@ -68,6 +68,8 @@ def sample(
     sampler.sample(n_samples=n_samples)
     t_elapsed = process_time() - t_start
 
+    logger.warning("Elapsed time: "+str(t_elapsed))
```
yannikschaelte, line 70: It may be nice to plot this, but I would not have it as a warning. maybe `logger.info` for the moment, and then we think about what information we really want to print.

pypesto/sampling/metropolis.py
```diff
@@ -32,13 +33,14 @@ def initialize(self, problem: Problem, x0: np.ndarray):
         self.trace_x = [x0]
         self.trace_fval = [self.objective(x0)]
 
-    def sample(self, n_samples: int, beta: float = 1.):
+    def sample(self, n_samples: int, beta: float = 1., hide_bar: bool = False):
```
yannikschaelte, line 36: this way, you cannot set the flag via `pypesto.sampling.sample.sample()`, right? I would say make `hide_bar` an argument of the `__init__` method instead.

yannikschaelte: strictly speaking, we cannot change the `sample` function signature, which is only allowed to take the arguments `n_samples` and `beta`.

pypesto/sampling/sample.py
elbaraim: I agree, however logger.info does not print the message in the python notebook. This is why I went for `warning` instead... any clue about this?

- Plotting how? What is your suggestion?

pypesto/sampling/metropolis.py
elbaraim: at the moment the progress bar will be displayed by default, may make sense to be able to disable it as well

elbaraim: > strictly speaking, we cannot change the `sample` function signature, which is only allowed to take the arguments `n_samples` and `beta`.

Then how would you suggest to do that?

pypesto/sampling/sample.py
yannikschaelte: plot = print, sorry. yes, but warning should be confined to things were we really nice to warn the user. we may soon decide to lift the default pypesto log level to info however, then such things would be shown.

elbaraim: done

yannikschaelte: see also #367 for further discussion

pypesto/sampling/metropolis.py
yannikschaelte: @elbaraim I made a suggestion in https://github.com/ICB-DCM/pyPESTO/pull/368, PR to here.

dweindl:


```diff

Complexity increasing per file
==============================
- pypesto/sampling/parallel_tempering.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dweindl/pyPESTO/pullRequest?prid=5608059&bid=18375252)

PR #368, opened 2020-05-29
add InternalSampler.make_internal; rerun sampling nbs

yannikschaelte:


pypesto/sampling/adaptive_metropolis.py
```diff
@@ -38,6 +38,8 @@ def default_options(cls):
             'cov0': None,
             # target acceptance rate
             'target_acceptance_rate': 0.234,
+            # show progress
+            'show_progress': True,
```
elbaraim, line 42: What about rename to `progress_bar`. It is more specific.

yannikschaelte: Both fine for me. I thought maybe at some point one there may be other "progress indicators" aside from the progressbar, and usually one wants to switch either all on or all off.

elbaraim:
codacy fails @yannikschaelte

yannikschaelte:
> codacy fails @yannikschaelte

it only checks some branches

PR #395, opened 2020-06-11
Additional code quality checks, minor simplifications

yannikschaelte:


pypesto/sampling/geweke_test.py
```diff
@@ -53,7 +53,7 @@ def spectrum(x: np.array,
     kmu = k * np.linalg.norm(w) ** 2
     spectral_density = np.zeros((nfft))
 
-    for i in range(k):
+    for _ in range(k):
```
elbaraim, line 56: that is correct

PR #405, opened 2020-06-18
small fix visu mcmc

elbaraim:
- Short/Small pull request:
MCMC visualization routines failed when fixed parameters where defined in the problem, since these are not sampled. Therefore the dim of the `parameter_names` did not agree with the actual sampled parameters. This is fixed now.

### DIFF
--- a/pypesto/visualize/sampling.py
+++ b/pypesto/visualize/sampling.py
@@ -351,7 +351,10 @@ def get_data_to_plot(
     theta_lb = result.problem.lb
     theta_ub = result.problem.ub
 
-    param_names = result.problem.x_names
+    # get parameter names from all non-fixed parameters
+    to_exclude = result.problem.x_fixed_indices
+    all_parameters = result.problem.x_names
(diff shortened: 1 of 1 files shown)


pypesto/visualize/sampling.py
```diff
@@ -351,7 +351,10 @@ def get_data_to_plot(
     theta_lb = result.problem.lb
     theta_ub = result.problem.ub
 
-    param_names = result.problem.x_names
+    # get parameter names from all non-fixed parameters
+    to_exclude = result.problem.x_fixed_indices
+    all_parameters = result.problem.x_names
+    param_names = [element for i, element in enumerate(all_parameters) if i not in to_exclude]
```
yannikschaelte, line 357: should be possible via
```python
param_names = problem.get_reduced_vector(problem.x_names)
```
, please check

elbaraim: indeed, thanks!

PR #408, opened 2020-06-18
outlaw large files; fixes #388

### LINKED ISSUE #388: Outlaw large files
Large files, e.g. from hdf5 or csv storage (happened formerly) appearing in the history should be avoided for reasons. Thus,
* Always check before commiting whether you have unwanted large files in there (git will also notifiy you)
* Squash-merging to develop should remove large files that existed at an intermediate stage, I think, so do that (now also required via https://help.github.com/en/github/administering-a-repository/requiring-a-linear-commit-history)
* Can we add a unit test that checks whether the .git history has increased too much?

yannikschaelte:
add a test on travis

### DIFF
--- a/test/check_git_size.sh
+++ b/test/check_git_size.sh
@@ -1,8 +1,13 @@
 #!/bin/sh
 
-size=$(du -s .git | cut -f1)
+# Check .git log size
+# Over time, this will increase s.t. the value will need to be adjusted.
+# However, if jumps in the size occur, this may indicate that undesirably
+# large files have been included.
 
-if [ "$size" -gt 20000 ]; then
-  echo "Git history is suspiciously large: $size"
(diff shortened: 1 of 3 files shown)


test/check_git_size.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+# Check .git log size
+# Over time, this will increase s.t. the value will need to be adjusted.
```
dweindl, line 4: Maybe helpful: https://stackoverflow.com/questions/10845051/git-show-total-file-size-difference-between-two-commits/10847242
Also okay to keep as is...

yannikschaelte: looks good, I think I will leave it as is for the moment though.

Issue #418, opened 2020-06-25
Visualization: add option for plotting a subset of parameters for parameters plot


plakrisenko:
It would be nice to be able to plot only a subset of parameters. It becomes especially important if the number of parameters is rather large.  There could be an argument of `pypesto.visualize.parameters.parameters` function that takes on values 'all', 'free_only' or a list with parameter indices, instead of `free_indices_only` argument.

Additionally, `results` argument in `pypesto.visualize.parameters.parameters` can be a `pypesto.Result` or a list of `pypesto.Result`. In the latter case, is it assumed that all results in the list have the same `pypesto.Problem`?

PR #425, opened 2020-07-03
Parameters plot, #418

plakrisenko:


pypesto/problem.py
```diff
@@ -303,23 +303,30 @@ def get_full_matrix(
...
-        Map vector from dim_full to dim, i.e. delete fixed indices.
+        Keep only those elements, which indices are specified in x_indices
+        If x_indices is not provided, delete fixed indices.
 
         Parameters
         ----------
         x_full: array_like, ndim=1
             The vector in dimension dim_full.
+        x_indices:
+            indices of x_full that should remain
```
paulstapor, line 318: Very good that you added that the full vector is meant! This can always be a little confusing...

pypesto/visualize/parameters.py
```diff
@@ -1,64 +1,89 @@
...
+        parameter_indices: Union[str, Sequence[int]] = 'free_only',
+        lb: Optional[Union[np.ndarray, List[float]]] = None,
+        ub: Optional[Union[np.ndarray, List[float]]] = None,
+        size: Optional[Tuple[float, float]] = None,
+        reference: Optional[List[ReferencePoint]] = None,
+        colors: Optional[Union[List[float], List[List[float]]]] = None,
+        legends: Optional[Union[str, List[str]]] = None,
+        balance_alpha: bool = True,
+        start_indices: Optional[Union[int, Iterable[int]]] = None
+) -> matplotlib.axes.Axes:
```
paulstapor, line 28: Thanks for the proper type hints here and everywhere! :+1:

pypesto/visualize/parameters.py
```diff
@@ -1,64 +1,89 @@
...
-    reference: list, optional
-        List of reference points for optimization results, containing et
+    reference:
+        List of reference points for optimization results, containing at
         least a function value fval
-    colors: list, or RGBA, optional
-        list of colors, or single color
+    colors:
+        list of RGBA colors, or single RGBA color
         color or list of colors for plotting. If not set, clustering is done
```
paulstapor, line 53: I think the first part of this line is redundant. 🙃

pypesto/visualize/parameters.py
```diff
@@ -1,64 +1,89 @@
...
+    ax:
         Axes object to use.
-    free_indices_only: bool, optional
-        If True, only free parameters are shown. If
-        False, also the fixed parameters are shown.
-    lb, ub: ndarray, optional
+    parameter_indices:
+        Specifies which parameters should be plotted. Allowed string values are
+        'all' (both fixed and free parameters will be plotted)  and
+        'free_only' (only free parameters will be plotted)
```
paulstapor, line 41: Cool feature. Was using hacks in PESTO back then to have precisely this functionality...

test/visualize/test_visualize.py
```diff
@@ -256,19 +256,19 @@ def test_parameters_with_options():
...
     visualize.parameters([result_1, result_2],
-                         free_indices_only=False,
+                         parameter_indices='all',
                          reference=ref_point,
                          balance_alpha=False,
                          start_indices=(0, 1, 4))
 
     visualize.parameters([result_1, result_2],
-                         free_indices_only=True,
+                         parameter_indices='all',
```
paulstapor, line 271: I think here it should be 'free_only'...
Would be good to us in the above example a list of indices, to also have the new functionality checked.

pypesto/visualize/parameters.py
plakrisenko: true :smile:

test/visualize/test_visualize.py
plakrisenko: Oh, you're right, thanks!

pypesto/visualize/parameters.py
```diff
@@ -1,64 +1,89 @@
...
 from .misc import process_result_list
 
 
-def parameters(results, ax=None, free_indices_only=True, lb=None, ub=None,
-               size=None, reference=None, colors=None, legends=None,
-               balance_alpha=True, start_indices=None):
+def parameters(
+        results: Union[Result, Sequence[Result]],
+        ax: Optional[matplotlib.axes.Axes] = None,
+        parameter_indices: Union[str, Sequence[int]] = 'free_only',
```
FFroehlich, line 19:
```suggestion
        parameter_indices: Union[Literal['all','free_only'], Sequence[int]] = 'free_only',
```

Literal can be imported from `typing`, see https://www.python.org/dev/peps/pep-0586/.

plakrisenko: But only starting python 3.8, right?

yannikschaelte: Looks like. Currently, pyPESTO is >=3.6. We could move to 3.8, but as long as e.g. anaconda still defaults to 3.7, I would not be too restrictive.

FFroehlich: Good point, probably shouldn't be changed yet to keep backwards compatibility.

PR #452, opened 2020-09-04
Fix optimization with priors

FFroehlich:
allows optimization with prior, otherwise aggregate objective can't aggregate (and history wont work).

### DIFF
--- a/pypesto/objective/priors.py
+++ b/pypesto/objective/priors.py
@@ -77,7 +77,7 @@ class NegLogParameterPriors(ObjectiveBase):
     ) -> ResultDict:
 
         res = {}
-        
+
         res[FVAL] = self.neg_log_density(x)
 
         for order in sensi_orders:
--- a/pypesto/objective/priors.py
+++ b/pypesto/objective/priors.py
@@ -77,10 +77,12 @@ class NegLogParameterPriors(ObjectiveBase):
     ) -> ResultDict:
 
         res = {}
+        
+        res[FVAL] = self.neg_log_density(x)
 
         for order in sensi_orders:
             if order == 0:
-                res[FVAL] = self.neg_log_density(x)
+                continue
             elif order == 1:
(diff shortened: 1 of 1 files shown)


yannikschaelte (approved):
?

yannikschaelte:
I actually missed the first green change. I'm not sure if I understand what the problem is with the previous implementation?

yannikschaelte:
so now the fval is always calculated. why is that required in the history / aggregated? wouldn't that rather be a potential problem if we require functions to always compute fval?

FFroehlich:
> so now the fval is always calculated. why is that required in the history / aggregated? wouldn't that rather be a potential problem if we require functions to always compute fval?



> I actually missed the first green change. I'm not sure if I understand what the problem is with the previous implementation?

https://github.com/ICB-DCM/pyPESTO/blob/master/pypesto/objective/amici_calculator.py 

We implemented that `AmiciObjective` does this by default. Aggregated objective could account for the fact that RVALS are inconsistent, but ideally this would be consistent across objectives. As far as I remember, always having fval is also what we expect in history.

yannikschaelte:
> > so now the fval is always calculated. why is that required in the history / aggregated? wouldn't that rather be a potential problem if we require functions to always compute fval?
> 
> > I actually missed the first green change. I'm not sure if I understand what the problem is with the previous implementation?
> 
> https://github.com/ICB-DCM/pyPESTO/blob/master/pypesto/objective/amici_calculator.py
> 
> We implemented that `AmiciObjective` does this by default. Aggregated objective could account for the fact that RVALS are inconsistent, but ideally this would be consistent across objectives. As far as I remember, always having fval is also what we expect in history.

Yes, Amici does it by default. I thought we had said the history does not require fval (is it only CsvHistory?), but I may definitely be wrong there, too many changes. As this here is not wrong, just minimally less efficient, fine for me.

PR #469, opened 2020-09-22
fix for new amici release

FFroehlich:


dweindl (approved):
This won't work with the current version, right? (tests will show in a minute..)
If so, we should also update the AMICI requirement, or switch based on AMICI version here.
Otherwise :+1:

FFroehlich:
> This won't work with the current version, right? (tests will show in a minute..)

Correct

> If so, we should also update the AMICI requirement, or switch based on AMICI version here.

Good Point.

> Otherwise 👍

Issue #467, opened 2020-09-22
check, if AMICI is installed in PEtab importer


jvanhoefer:
I think, when running `importer.create_problem()`,  it would be good to check if AMICI is installed and throw a more specific error. Currently you get `NameError: name 'amici' is not defined`.

https://github.com/ICB-DCM/pyPESTO/blob/cbec7ee1343b4d66f86eb26e1d77cf77571184a4/pypesto/petab/importer.py#L68

Similarly for `PEtab` and `pypesto.petab.PetabImporter.from_yaml()`.

PR #470, opened 2020-09-22
add check for installation ipopt

jvanhoefer:
Currently it was not checked, if IpOpt is installed.


yannikschaelte:
I think we should do this similarly for many optional dependencies (pyswarm, pymc3, ipopt, dlib, ...). Checking for availability like here when invoking the main class? As e.g. the IpoptOptimizer is imported into pypesto.optimize by default, catching on import submodule is currently not possible, so this way here would seem fine to me. Opinions? @FFroehlich @dweindl

FFroehlich:
> I think we should do this similarly for many optional dependencies (pyswarm, pymc3, ipopt, dlib, ...). Checking for availability like here when invoking the main class? As e.g. the IpoptOptimizer is imported into pypesto.optimize by default, catching on import submodule is currently not possible, so this way here would seem fine to me. Opinions? @FFroehlich @dweindl

Looks good to me.

dweindl:
Works for me. For this and similar examples, one could also postpone the import to the respective functions. This is the only place ipopt is actually used. No need to load all those libraries that are rarely used in combination...

jvanhoefer:
This is also how it is done for all other optimizers except IpOpt...

jvanhoefer:
This is also how it is done for all other optimizers except IpOpt...

jvanhoefer:
Still agree, that we should do more specific error messages, if optional dependencies are used, see also e.g. #467

yannikschaelte (approved):

PR #476, opened 2020-10-01
add maxiter to Dlibs default options fixes #474

### LINKED ISSUE #474: Dlib: Default value for 'maxiter'
If no options are specified, Dlib sets `maxiter` per default to 0 (= no steps) => no single optimization step is taken...

Solution: specify `maxiter` in https://github.com/ICB-DCM/pyPESTO/blob/cbec7ee1343b4d66f86eb26e1d77cf77571184a4/pypesto/optimize/optimizer.py#L517

jvanhoefer:


jvanhoefer:
I now choose `maxiter = 10000`, since 1000 does not converge at all for the Böhm model, while 10 000 does...

pypesto/optimize/optimizer.py
```diff
@@ -481,6 +481,9 @@ def __init__(self,
         self.options = options
         if self.options is None:
             self.options = DlibOptimizer.get_default_options(self)
+        elif 'maxiter' not in self.options:
+            raise RuntimeWarning('Dlib options are missing the key word '
```
yannikschaelte, line 485:
```suggestion
            raise KeyError('Dlib options are missing the key word '
```

yannikschaelte (approved):

PR #477, opened 2020-10-01
check optional requirements amici/petab in PEtab import closes #467

### LINKED ISSUE #467: check, if AMICI is installed in PEtab importer

jvanhoefer:


yannikschaelte (approved):
you tested on a system without petab/amici installed, right? would probably also be something to set up a little test system for (not necessary now imho).

jvanhoefer:
I tried the PEtab part without PEtab installed...

PR #482, opened 2020-10-06
Feature synthetic data

dilpath:
To use this, the `feature_synthetic_data` branch of AMICI and PEtab (and pyPESTO for the example) should be used together.
Implementation in AMICI: [PR](https://github.com/AMICI-dev/AMICI/pull/1283)
Example optimisation with synthetic data: [notebook](https://github.com/ICB-DCM/pyPESTO/blob/feature_synthetic_data/doc/example/synthetic_data.ipynb), [PR](https://github.com/ICB-DCM/pyPESTO/pull/482)
PEtab base class: [code](https://github.com/PEtab-dev/PEtab/blob/feature_synthetic_data/petab/simulate.py), [PR](https://github.com/PEtab-dev/PEtab/pull/472)


dilpath:
Ready for review, but merging will be delayed as this PR requires the next AMICI release

setup.py
```diff
@@ -30,12 +30,15 @@ def read(fname):
                         'seaborn>=0.10.0',
                         'cloudpickle>=0.7.0',
                         'h5py>=2.10.0',
+                        'amici @ git+https://github.com/amici-dev/amici'
```
jvanhoefer, line 33: Do we really want to hardcode the dependency on the AMICI develop branch in our current branch? @yannikschaelte Any opinions? :)

doc/example/synthetic_data.ipynb
```diff
@@ -0,0 +1,258 @@
+{
```
jvanhoefer, line 1: Can you maybe also give short comments (as `.md` cells) between the many cells, that you have. One sentence is probably enough per cell, but helps the reader to understand what you are doing there :)

jvanhoefer (changes requested):

setup.py
yannikschaelte: I would suggest to first merge amici.

dilpath: Yep I agree :+1: Meant that in my earlier comment

> merging will be delayed as this PR requires the next AMICI release

`develop` is just used now to check that everything will work. Similarly, the current edit to ".travis.yml" will also be removed from this PR before merging.

doc/example/synthetic_data.ipynb
dilpath: Thanks, added some additional descriptions.

dilpath:
This PR is now in a state that I would merge (if approved)

yannikschaelte (approved):

jvanhoefer (approved):

PR #485, opened 2020-10-19
Parameters hist

plakrisenko:
Plot parameter values from different multistarts as a histogram.


plakrisenko:
the amici_import example error is fixed in another pull-request (visualization of optimization result properties)

pypesto/visualize/parameters.py
```diff
@@ -115,6 +115,68 @@ def parameters(
     return ax
 
 
+def parameter_hist(
+        result: Result,
+        parameter_name: str,
```
paulstapor, line 120: Also allows for lists? :thinking: ?

paulstapor (approved):

pypesto/visualize/parameters.py
plakrisenko: Actually, lists are now allowed at the moment

PR #493, opened 2020-11-03
NLOpt Optimizers

FFroehlich:


pypesto/optimize/optimizer.py
```diff
@@ -630,3 +636,162 @@ def minimize(
...
+        elif 'maxiter' in options:
+            options['maxeval'] = options.pop('maxiter')
+        if local_options is None:
+            local_options = {}
+        self.options = options
+        self.local_options = local_options
+        if nlopt is None:
+            raise ImportError(
+                "This optimizer requires an installation of NLopt. You can "
+                "install NLopt via pip install nlopt.")
```
LeonardSchmiester, line 681: Is checked twice, here and in `minimize()`.

LeonardSchmiester (approved):
Looks good to me :+1:

pypesto/optimize/optimizer.py
```diff
@@ -630,3 +636,162 @@ def minimize(
...
+        needs_local_method = [
+            nlopt.G_MLSL, nlopt.G_MLSL_LDS, nlopt.GD_MLSL, nlopt.GD_MLSL_LDS,
+            nlopt.AUGLAG, nlopt.AUGLAG_EQ
+        ]
+
+        if local_method is None and method in needs_local_method:
+            local_method = nlopt.LD_LBFGS
+
+        if local_method is not None and method not in needs_local_method:
+            raise ValueError(f'Method "{method}" does not allow a local '
```
jvanhoefer, line 695: Shall we rather raise a warning, that the local method is unused and continue without it?

jvanhoefer (approved):
👍

pypesto/optimize/optimizer.py
FFroehlich: I see no good reason why a user would want that and I generally like working with sanitized input.

pypesto/optimize/optimizer.py
FFroehlich: True, if we check in `__init__`, it should be available in `minimize`.

test/test_optimize.py
```diff
@@ -28,6 +29,20 @@ def mode(request):
     ('dlib', ''),
     ('pyswarm', ''),
     ('cmaes', ''),
+    *[('nlopt', method) for method in [
```
yannikschaelte, line 32: wow, testing all :laughing: but yes, makes sense.

yannikschaelte (commented):
I was anyway too late, but looks good! :+1:

test/test_optimize.py
FFroehlich: with the rosen objective these finish in a matter of a couple of seconds, makes me more confident they are not completely terrible :D

PR #495, opened 2020-11-04
Close file descriptors of temporary files

lcontento:
In the pickling machinery for AMICI objectives, temporary files are used to store the AMICI solver settings. The files themselves were removed when no longer needed, but the file descriptors were not closed. This resulted in orphaned file descriptors, which can eventually lead to a `Too many open files` error (happened to me while using the highly-parallel `emcee` sampler, which does a lot of pickling).


pypesto/objective/amici.py
```diff
@@ -222,7 +222,8 @@ def __getstate__(self) -> Dict:
         # read in byte stream
         amici_solver_settings = open(amici_solver_file, 'rb').read()
         state['amici_solver_settings'] = amici_solver_settings
-        # remove temporary file
+        # close file descriptor and remove temporary file
+        os.close(fd)
```
dweindl, line 226: What about letting [tempfile.TemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile) handle that?

lcontento: It is a possibility (`NamedTemporaryFile` should be used since AMICI needs to access it too), doing so will open the file for I/O so that a `flush()` will probably be needed before reading from AMICI. I would personally prefer to keep the two things explicitly separate, but it is just a matter of taste and both solutions would be perfectly equivalent.

dweindl: The problem with the current implementation is that it's not exception safe. If anything `raise`s before `close`, nothing will be closed.

lcontento: Added a `try .. finally`.

dweindl (approved):

yannikschaelte (approved):

lcontento:
I am not authorized to merge (I think I was at some point?), could someone do it for me?

yannikschaelte:
does it work now?

lcontento:
> does it work now?

Nope, still says "You’re not authorized to merge this pull request"...

yannikschaelte:
Now?

lcontento:
> Now?

Works, thanks!!

PR #500, opened 2020-11-12
Feature fides optimizer

FFroehlich:


pypesto/optimize/optimizer.py
```diff
@@ -790,3 +795,78 @@ def nlopt_objective(x, grad):
 
     def is_least_squares(self):
         return False
+
+
+class FidesOptimizer(Optimizer):
+    """
+    Global/Local optimization using the trust region optimizer fides.
```
yannikschaelte, line 802: add a link to the github page (or rtd, when exists)?

pypesto/optimize/optimizer.py
```diff
@@ -790,3 +795,78 @@ def nlopt_objective(x, grad):
...
+                 hessian_update: Optional[fides.HessianApproximation] =
+                 None,
+                 options: Optional[Dict] = None):
+        """
+        Parameters
+        ----------
+        options:
+            Optimizer options.
+
+        hessian_update:
```
yannikschaelte, line 815: make it clearer that None means the objective's Hessian (/approximation) is used?

yannikschaelte (approved):

pypesto/optimize/optimizer.py
FFroehlich: 👍

pypesto/optimize/optimizer.py
FFroehlich: added

PR #512, opened 2020-11-20
add requirements petabtests closes #511

### LINKED ISSUE #511: petabtest missing in requirements

jvanhoefer:


yannikschaelte (commented):
petabtests ist afaik not available as a pypi package, because it requires a custom install (would be possible, but didn't bother at the time), therefore it is custom-installed on CI from github.

jvanhoefer:
This made some problems with local testing for Simon and me today... are there any reasons, why this should not be on pypi? Otherwise, we can just do the link to the current GitHub Repo. Just I think, we should have this requirement in there.

dweindl:
> are there any reasons, why this should not be on pypi?

No. Just needs to be packaged.

jvanhoefer:
> No. Just needs to be packaged.

Then I/we should wait with this :)

yannikschaelte:
> > No. Just needs to be packaged.
> 
> Then I/we should wait with this PR until `petabtests` is on pypi :)

@jvanhoefer done

yannikschaelte (approved):
petabtests>=0.0.0a5 now also comes versioned :tada:

PR #513, opened 2020-11-20
Disable FIM for models with parameter dependent sigma

FFroehlich:
Disable use of FIM for parameter dependent sigma until https://github.com/AMICI-dev/AMICI/issues/1335 is resolved


LeonardSchmiester (approved):
:+1:

yannikschaelte (approved):

paulstapor:
Let's see whether https://github.com/AMICI-dev/AMICI/pull/1338 resolves it...

FFroehlich:
> Let's see whether [AMICI-dev/AMICI#1338](https://github.com/AMICI-dev/AMICI/pull/1338) resolves it...
> I think you suggested the Schwen model to double check. Whic optimizer and setting would make the change in the FIM most apparent?

default options :)

paulstapor:
> default options :)

Default options is L-BFGS-B... Does this really use a FIM? (running it atm)

FFroehlich:
fixed in amici

PR #516, opened 2020-11-23
add bounds check fixes #515 and fixes #509

### LINKED ISSUE #509: NaN check for lb/ub
Would be nice if `problem.{ub,lb}` is checked for nan values instead of generating startpoints containing nans.

### LINKED ISSUE #515: normalize() accidentally overwrites bounds
When fixing #509, I found, that we accidentally overwrite the bounds in the `normalize()` function `problem`: If the bounds do not contain the fixed values, we overwrite `self.ub_full`:

https://github.com/ICB-DCM/pyPESTO/blob/8275751eba87efedd6ad1e6923a94d7de0603ddd/pypesto/problem.py#L174

We then set `self.ub_full[self.x_free_indices] = self.ub`, which looks like it recovers the values from `self.ub`, BUT it doesn't: `self.ub` is a `@property` function, that itself only returns the corresponding values of `self.ub_full`. 

Those were lost, when initializing/overwriting these values with `np.empty`.

jvanhoefer:


pypesto/problem.py
```diff
@@ -199,8 +205,14 @@ def normalize(self) -> None:
             raise AssertionError("x_names must be of length dim_full.")
         if len(self.x_fixed_indices) != len(self.x_fixed_vals):
             raise AssertionError(
-                "x_fixed_indices and x_fixed_vals musti have the same length."
+                "x_fixed_indices and x_fixed_vals must have the same length."
```
yannikschaelte, line 208: :thinking:

yannikschaelte (approved):

pypesto/problem.py
```diff
@@ -117,6 +117,8 @@ def __init__(self,
 
         self._x_free_indices: Union[List[int], None] = None
 
+        # check if lower bounds are not nan
```
FFroehlich, line 120:
```suggestion
```

jvanhoefer: oh, yep 🤔

PR #517, opened 2020-11-23
fill result from history

plakrisenko:
read_result_from_file updated to collect results from multiple history files


yannikschaelte (commented):
@FFroehlich before reviewing: I think `read_result_from_file` was from you?
I guess it would be preferable to keep both singular and plural versions.

pypesto/optimize/optimizer.py
```diff
@@ -197,33 +200,71 @@ def fill_result_from_objective_history(
...
+        )
+
+        result = OptimizerResult(
+            id=identifier,
+            message='loaded from file',
+            exitflag=EXITFLAG_LOADED_FROM_FILE,
+            time=max(history.get_time_trace())
+        )
+        result.id = identifier
+        result = fill_result_from_objective_history(result, opt_hist)
```
yannikschaelte, line 262: The repeated execution here will not work properly yet, as it won't take the minimum, but at the moment rather only performs one replacement.

FFroehlich:
> @FFroehlich before reviewing: I think `read_result_from_file` was from you?
> I guess it would be preferable to keep both singular and plural versions.

Correct. Ideally `read_result_from_files` does some comprehension using `read_result_from_file` and then performs merging. Would also yield smaller functions.

yannikschaelte:
> > @FFroehlich before reviewing: I think `read_result_from_file` was from you?
> > I guess it would be preferable to keep both singular and plural versions.
> 
> Correct. Ideally `read_result_from_files` does some comprehension using `read_result_from_file` and then performs merging. Would also yield smaller functions.

Exactly. I think merging the results list to a OptimizeResult, as currently done, is good. So @plakrisenko , could you make the plural function loop over and merge the singular function? Then that should be fine.

PR #519, opened 2020-11-24
Unconstrained Optimization

### LINKED ISSUE #518: Add "unbounded" mode
... where lb/ub are only used for startpoint sampling and optimization problem is unbounded

FFroehlich:
Works with all optimizers but `pyswarm` and `dlib`. Also some optimizers are terribad ...


yannikschaelte (approved):
I believe this should all work fine, no question. However, I was wondering conceptually whether it may be desirable to make the distinction clearer between startpoint and objective bounds, same as it is done in [petab](https://petab.readthedocs.io/en/stable/documentation_data_format.html#id3) (the difference is not supported yet), e.g. by not using the "ConstraintRemover" but instead allowing to pass both ub and ub_init to the problem class, where the former could then just be inf directly. Having both ub and ub_init will be necessary at some point anyway. What would be your opinion on that? Maybe also @dweindl @jvanhoefer ?

A problem with what I just said might be that in sampling, inf bounds make not so very much sense, so having this "unconstrained optimization" in addition may also work.

FFroehlich:
> I believe this should all work fine, no question. However, I was wondering conceptually whether it may be desirable to make the distinction clearer between startpoint and objective bounds, same as it is done in [petab](https://petab.readthedocs.io/en/stable/documentation_data_format.html#id3) (the difference is not supported yet), e.g. by not using the "ConstraintRemover" but instead allowing to pass both ub and ub_init to the problem class, where the former could then just be inf directly. Having both ub and ub_init will be necessary at some point anyway. What would be your opinion on that? Maybe also @dweindl @jvanhoefer ?
> 
> A problem with what I just said might be that in sampling, inf bounds make not so very much sense, so having this "unconstrained optimization" in addition may also work.

Oh that actually makes a lot of sense, implementation would also be easier and less hacky. Wasn't aware this was already an (unsupported) petab feature. Will adapt implementation.

pypesto/optimize/options.py
```diff
@@ -10,18 +10,23 @@ class OptimizeOptions(dict):
     startpoint_resample:
         Flag indicating whether initial points are supposed to be resampled if
         function evaluation fails at the initial point
-    allow_failed_starts: bool, optional
+    allow_failed_starts:
         Flag indicating whether we tolerate that exceptions are thrown during
         the minimization process.
+    unbounded_optimization:
```
jvanhoefer, line 16: I would call this `unconstrained_optimization`, in order to have this consistent

jvanhoefer (approved):

jvanhoefer:
Do we also want to allow to be unconstrained only in "some dimensions/directions", e.g `lb = [0, 0, -inf]`?

dweindl:
>  it may be desirable to make the distinction clearer between startpoint and objective bounds, same as it is done in [petab](https://petab.readthedocs.io/en/stable/documentation_data_format.html#id3) (the difference is not supported yet)

Yes, would be great to get that into pypesto. Probably relevant to several people.

>  allowing to pass both ub and ub_init to the problem class, where the former could then just be inf directly

:+1: 

Besides that, I can't really judge how helpful it will be to have `ConstraintRemover` on top. Might be convenient for interactive use, but not sure.

jvanhoefer:
So in general I really like @yannikschaelte's idea of having `ub` and `ub_init` separated. Just, isn't this to some degree already possible in PEtab, via specifying `ub=inf` and specifying an `initializationPriorType ` in the parameter table? (Also isn't at least for PEtab models it already possible to have unconstrained problems?)

FFroehlich:
> So in general I really like @yannikschaelte's idea of having `ub` and `ub_init` separated. Just, isn't this to some degree already possible in PEtab, via specifying `ub=inf` and specifying an `initializationPriorType ` in the parameter table? (Also isn't at least for PEtab models it already possible to have unconstrained problems?)

I think uniform prior with boundaries should be equivalent to specifying `ub_init`.

FFroehlich:
Reimplemented using `{lb,ub}_init`

pypesto/optimize/optimizer.py
```diff
@@ -276,6 +276,12 @@ def get_default_options(self):
         return None
 
 
+def check_finite_bounds(lb, ub):
+    if not np.isfinite(lb).all() or not np.isfinite(ub).all():
```
yannikschaelte, line 280:
```suggestion
def check_finite_bounds(lb, ub):
    """Raise if bounds are not finite."""
    if not np.isfinite(lb).all() or not np.isfinite(ub).all():
```

pypesto/problem.py
```diff
@@ -32,7 +32,10 @@ class Problem:
         The objective function for minimization. Note that a shallow copy
         is created.
     lb, ub:
-        The lower and upper bounds. For unbounded directions set to inf.
+        The lower and upper bounds for optimization. For unbounded directions
+        set to +-inf.
+    lb_init, ub_init:
+        The lower and upper bounds for initialization.
```
yannikschaelte, line 38:
```suggestion
        The lower and upper bounds for initialization, typically for defining
        search start points.
        If not set, set to lb, ub.
```

yannikschaelte (approved):
Left to do afterwards: Adjust in PEtab import, #522.

jvanhoefer:
Don't we need to adapt the `normalize` function to also deal with `lb_init` and `ub_init`?

yannikschaelte:
> Don't we need to adapt the `normalize` function to also deal with `lb_init` and `ub_init`?

Yep

PR #521, opened 2020-11-25
add additional nan checks to fval, grad and hess, fixes #514

### LINKED ISSUE #514: Failing to optimize PEtab problem
I am trying to fit a PEtab problem:
```Python
import pypesto
import pypesto.petab
import pypesto.optimize as optimize
import os

dir_name = os.path.dirname(__file__)
yaml_config = os.path.join(dir_name, 'PEtab', 'v3.0.0.yaml')

importer = pypesto.petab.PetabImporter.from_yaml(yaml_config)
problem = importer.create_problem()
```
but get the following error:
```
  File "launch_pypesto.py", line 10, in <module>
    problem = importer.create_problem()
  File "/media/sf_DPhil_Project/Project04_Full cell cycle modelling/cell_cycle_model/venv/lib/python3.8/site-packages/pypesto/petab/importer.py", line 326, in create_problem
    objective = self.create_objective(**kwargs)
  File "/media/sf_DPhil_Project/Project04_Full cell cycle modelling/cell_cycle_model/venv/lib/python3.8/site-packages/pypesto/petab/importer.py", line 261, in create_objective
    obj = AmiciObjective(
  File "/media/sf_DPhil_Project/Project04_Full cell cycle modelling/cell_cycle_model/venv/lib/python3.8/site-packages/pypesto/objective/amici.py", line 151, in __init__
    raise ValueError('Steadystate prediction is not supported for '
ValueError: Steadystate prediction is not supported for models with conservation laws!
```
I am confused, because I do not think the model has conservation laws and I also do not (consciously) try to predict a steady state. Do you have an idea why that is happening? I can send the PEtab problem per email if needed.

I am using In [4]: petab v0.1.11, amici v0.11.8 and pypesto v0.2.2.

FFroehlich:


yannikschaelte (approved):
just to clarify: this will return "error" whenever any component of gradient|hessian is nan or inf. probably makes sense as hardly any optimizer will be able to take a meaningful step, but why is this necessary?

FFroehlich:
> just to clarify: this will return "error" whenever any component of gradient|hessian is nan or inf. probably makes sense as hardly any optimizer will be able to take a meaningful step, but why is this necessary?

Corret. #514 to me suggests this is necessary. One example where I can imagine this happening is multiplying inf with zero while applying the chain rule for parameter transformation.

PR #527, opened 2020-11-30
Feature mcmc conditions plot

dilpath:
- group plots by condition
  - example in notebook
- parallelize sample evaluation


elbaraim:
Thanks for the contribution!
I am wondering: Wouldn't it be better to have one plotting function instead of two separated ones?
This could be potentially done as an additional argument:
`sampling_prediction_profiles(group_by='states')`
or
`sampling_prediction_profiles(group_by='conditions')`

- What are your thoughts on this?

dilpath:
> Thanks for the contribution!
> I am wondering: Wouldn't it be better to have one plotting function instead of two separated ones?
> This could be potentially done as an additional argument:
> `sampling_prediction_profiles(group_by='states')`
> or
> `sampling_prediction_profiles(group_by='conditions')`
> 
>     * What are your thoughts on this?

Thanks, I agree. Might also be nice to be able to save the samples for use in other plots, such that all four combinations of plots (states or observables, grouped by state/observable or condition) can be produced without having to repeat the sampling.

I didn't combine done at this stage to avoid possible conflicts with the code that this would be merged into (since that is also a current PR that might have changes made to it). Shall I combine them now?

dilpath:
> Thanks for the contribution!
> I am wondering: Wouldn't it be better to have one plotting function instead of two separated ones?
> This could be potentially done as an additional argument:
> `sampling_prediction_profiles(group_by='states')`
> or
> `sampling_prediction_profiles(group_by='conditions')`
> 
>     * What are your thoughts on this?

Thanks, I agree. Might also be nice to be able to save the samples for use in other plots, such that all four combinations of plots (states or observables, grouped by state/observable or condition) can be produced without having to repeat the sampling.

I didn't combine them at this stage to avoid possible conflicts with the code that this would be merged into (since that is also a current PR that might have changes made to it). I'll combine them in the next few days then.

dilpath:
Now ready for review

yannikschaelte:
The diagnostics notebook is only partially evaluated. Could you run it through and upload?

dilpath:
> The diagnostics notebook is only partially evaluated. Could you run it through and upload?

Thanks for checking, should be OK now

https://github.com/ICB-DCM/pyPESTO/blob/354cf0e6a13d6bf09189dabb8e8bf70207c95c27/doc/example/sampling_diagnostics.ipynb

yannikschaelte:
Would it make sense to invert the colors, i.e. 99% -> light, 90% -> dark? Seems a bit counter-intuitive this way for me ... but the other way round the center might be too dark with the black ...

pypesto/visualize/__init__.py
```diff
@@ -6,34 +6,52 @@
...
-                       sampling_1d_marginals,
-                       sampling_parameters_cis,
-                       sampling_prediction_profiles)
-from .ensemble import ensemble_identifiability
\ No newline at end of file
+from .reference_points import (  # noqa: F401
+    ReferencePoint,
+    create_references,
+)
+from .clust_color import (  # noqa: F401
```
yannikschaelte, line 13: s.a.

pypesto/visualize/sampling.py
```diff
@@ -1,18 +1,38 @@
...
+from ..ensemble import (
+    get_percentile_label,
+    EnsemblePrediction,
+    MEDIAN,
+)
 
 logger = logging.getLogger(__name__)
 
 
+# TODO observable here refers to the output of an ensemble prediction, which
```
yannikschaelte, line 29: is this still TODO or for later?

pypesto/visualize/sampling.py
```diff
@@ -92,119 +112,347 @@ def sampling_fval_trace(
...
+        n_row = int(np.round(np.sqrt(n_subplots)))
+        n_col = int(np.ceil(n_subplots / n_row))
+        fig, axes = plt.subplots(n_row, n_col, figsize=size, squeeze=False)
+    elif len(axes.flat) < n_subplots:
+        raise ValueError(
+            'Provided `ax` contains insufficient subplots. At least '
+            f'{n_subplots} are required.'
+        )
+
+    def get_statistic_data(
```
yannikschaelte, line 254: docstring

pypesto/visualize/sampling.py
```diff
@@ -92,119 +112,347 @@ def sampling_fval_trace(
...
-    alpha:
-        List of lower tail probabilities, defaults to 95% interval.
-    stepsize:
-        Only one in `stepsize` values is simulated for the intervals
-        generation. Recommended for long MCMC chains. Defaults to 1.
-    plot_type:
-        Visualization mode for prediction intervals {‘states’, ‘observables’}.
-        Defaults to ‘states’.
+    percentiles:
+        Lower tail probabilities, e.g. `[95]` for a 95% credibility interval.
```
yannikschaelte, line 145: "Confidence levels"? And maybe note that these are translated into symmetric intervals, e.g. 95% -> 2.5% - 97.5%. An alternative would be e.g. highest-density regions.

pypesto/visualize/sampling.py
```diff
@@ -92,119 +112,347 @@ def sampling_fval_trace(
...
-                # Update X axis array
-                t += n_timepoints
+    # define colormap
+    variables_color = [
+        list(cmap(v))[:LEN_RGB]
+        for v in np.linspace(cmap_min, cmap_max, n_variables)
+    ]
+
+    if axes is None:
+        n_row = int(np.round(np.sqrt(n_subplots)))
```
yannikschaelte, line 245: I think usually wider plots are preferable to higher plots. Thus, e.g. if n=6, here it would be 3 rows, 2 cols, right? Thus it may make sense to calculate the cols first?

pypesto/visualize/sampling.py
```diff
@@ -92,119 +112,347 @@ def sampling_fval_trace(
...
+            condition_index: int,
+            observable_id: int,
+    ):
+        condition_result = summary[statistic].conditions[condition_index]
+        t = condition_result.timepoints
+        observable_index = condition_result.observable_ids.index(observable_id)
+        y = condition_result.output[:, observable_index]
+        return (t, y)
+
+    if groupby == CONDITION:
```
yannikschaelte, line 265: I think these might want to become own private functions ... makes dependencies clearer ...

yannikschaelte (changes requested):

yannikschaelte:
Looks very nice, plots as well!

dilpath:
> Would it make sense to invert the colors, i.e. 99% -> light, 90% -> dark? Seems a bit counter-intuitive this way for me ... but the other way round the center might be too dark with the black ...

Yes I think the center can be too dark w.r.t. the median line, and the edges can be too faint against the white background, if the colors are reversed.

pypesto/visualize/sampling.py
dilpath: This is because the current prediction code names all output columns `observable_ids` (e.g. accessed with `PredictionResult.conditions[0].observable_ids`), regardless of whether the post-processor only returns states. Not sure if this might be confusing to users. Any opinion @paulstapor ? Can leave as-is, and remove the TODO. I used `trajectories` in the `sampling_prediction_trajectories` function, instead of observables, to avoid confusion there. Other alternatives could be
- variables / variable_ids
- timecourses / timecourse_ids

dilpath: Also perhaps `pypesto.prediction` could be renamed to `pypesto.predict`, to match the other pyPESTO submodules

yannikschaelte: Leaving the former point to Paul, on the latter: Yes, we use infinitive verbs for module names (and functions). Reason: Don't remember, probably shorter + sounded better.

pypesto/visualize/sampling.py
dilpath: In the current state I think it would actually be 2 rows by 3 cols?

dilpath:
In addition to addressing the resolved comments, commits since Yannik's review made the following changes.
- duplicated the conversion reaction PEtab problem to have multiple conditions, such that multiple conditions can be plotted in the notebook
    - (re-ran the notebook to include the new plots)
- changed labeling to truncate percentiles to 2 decimal places in labels, with a `'...'` suffix if truncation changed the value of the percentile
    - I expect there will be a bug if percentiles are requested that are equivalent to 2 decimal places, though I guess this is not a reasonable use case

Something not addressed here but might be helpful: combine all constants into a single file.

The parameter name `percentiles` is still used in the plotting function, for consistency with the ensemble prediction code. A more appropriate name might be e.g. `levels`.

test/visualize/test_visualize.py
```diff
@@ -177,6 +179,25 @@ def create_plotting_options():
     return ref1, ref2, ref3, ref4, ref_point
 
 
+def post_processor(
+        amici_outputs,
+        output_type,
+        observable_ids,
+):
+    # This post_processor will transform the output of the simulation tool
+    # such that the output is compatible with the next steps.
```
yannikschaelte, line 188: convert to docstring

yannikschaelte (approved):
Looks good to me!

pypesto/visualize/misc.py
```diff
@@ -225,3 +225,57 @@ def process_start_indices(start_indices: Union[int, Iterable[int]],
...
+LEN_RGB = 3
+LEN_RGBA = 4
+TYPING_RGB = Tuple[(float,) * LEN_RGB]
+TYPING_RGBA = Tuple[(float,) * LEN_RGBA]
+TYPING_RGB_RGBA = Union[TYPING_RGB, TYPING_RGBA]
+RGBA_MIN = 0  # min value for an RGBA element
+RGBA_MAX = 1  # max value for an RGBA element
+RGBA_ALPHA = 3  # zero-indexed fourth element in RGBA
+RGBA_WHITE = (RGBA_MAX, RGBA_MAX, RGBA_MAX, RGBA_MAX)
+RGBA_BLACK = (RGBA_MIN, RGBA_MIN, RGBA_MIN, RGBA_MAX)
```
paulstapor, line 239: I thought we had something similar in the waterfall visualization routine... Unify?

paulstapor (approved):
Looks good, minor things only

doc/example/conversion_reaction_conditions_multiple/model_conversion_reaction.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
yannikschaelte: I was wondering too. If it is the base model extended for multiple conditions, it would be fine, but in general it would be nice to keep the number of models included low, so if there are synergies, preferable.

pypesto/visualize/misc.py
dilpath: Unified into a "constants.py" file.

pypesto/ensemble/ensemble.py
```diff
@@ -283,6 +288,12 @@ def __init__(self,
         upper_bound:
             array of potential upper bounds for the parameters
         """
+        # Do we have a representative sample or just random ensemble?
+        self.ensemble_type = EnsembleType.ensemble
+        if ensemble_type is not None:
+            self.ensemble_type = ensemble_type
+            if ensemble_type == EnsembleType.sample:
+                x_vectors = x_vectors.T
```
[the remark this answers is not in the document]
dilpath: I think it's more user-friendly to allow a user to supply a sample chain and specify that it is a sample chain (`EnsembleType.sample`), and expect the `Ensemble` class to correctly handle parameters from different areas of pyPESTO (e.g. also `EnsembleType.optimize` / `EnsembleType.profile`).

Alternatively, I could add the following method to the `Ensemble` class.
```python
@staticmethod
def from_sample(result: pypesto.Result, chain_index: int = 0, **kwargs):
    x_vectors = result.sample_result.trace_x[chain_index].T
    return Ensemble(x_vectors, **kwargs)
```

doc/example/conversion_reaction_conditions_multiple/model_conversion_reaction.xml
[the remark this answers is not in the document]
dilpath: Moved into a subfolder: `doc/example/conversion_reaction/multiple_conditions/`, with all identical files reused.

dilpath:
Renamed `percentiles` to `levels` in the visualization code, to avoid confusion with the use of percentiles in the `pypesto.ensemble.EnsemblePrediction.compute_summary` method (I think `levels` is a better name anyway).

pypesto/visualize/sampling.py
paulstapor: One could also go for "output_ids"?

paulstapor: Renaming is totally fine for me
Yes, in ensemble and prediction, there are many things to be settled still. The main functionality is there, but many things may still converge to some formulation which suits everybody most

pypesto/ensemble/ensemble.py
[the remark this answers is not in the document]
paulstapor: Totally in favor. This would be analogous to from_csv, and at some point, from_h5 and from_optimizer_history

pypesto/visualize/misc.py
paulstapor: töp

PR #528, opened 2020-12-01
Fix reading problem.x_names from HDF5

dweindl:
From h5py 3.0.0 on, strings are read as `bytes` instrad of `str`, so we need to convert.


plakrisenko (approved):

jvanhoefer (approved):

dweindl:
Hm, what's going on with `test_unbounded_minimize` here?

PR #532, opened 2020-12-03
added engines to profiling as well as engine_test in test_profile

PaulJonasJost:
Hello, 

i added the engines to the profiling. Currently (until approval) i left the option to use the current version of profiling when using `engine=None`. I also added a test in test_profiles.

Kind regards,
Paul


pypesto/profile/profile.py
```diff
@@ -86,139 +89,64 @@ def create_next_guess(x, par_index, par_direction_, profile_options_,
...
-
-        # compute profile in descending and ascending direction
-        for par_direction in [-1, 1]:
-            # flip profile
-            current_profile.flip_profile()
-
-            # compute the current profile
-            current_profile = walk_along_profile(
+    # if engine==None do the original profiling
+    if engine is None:
```
yannikschaelte, line 93: would it be possible to define a "dummy engine" which just executes the for loop sequentially? it could use the same ProfilerTask definition as below, so one would simplify the code by not having that flip_profile, walk_along_profile etc twice.

yannikschaelte:
Thanks @PaulJonasJost , much appreciated!

pypesto/profile/profile.py
yannikschaelte: could just be SingleCoreEngine, right?

yannikschaelte: which should then be the default.

PaulJonasJost: Could do, but normally the engine used would be the `SingleCoreEngine` wouldn't it? I thought that if it get approved, i would delete the whole block and put instead:
```python
if engine is None: engine = SingleCoreEngine()
```
That is at least the case in the optimize function.

PaulJonasJost: oh ok, the messages did not load. Will do so 👍

yannikschaelte: yes, exactly. please go ahead and remove the old code already, then a review will be easier.

PaulJonasJost: @yannikschaelte should be done now.

yannikschaelte:
@PaulJonasJost sorry, we did not respond overly quickly here. Could you rerun the tests to see whether all works with the latest releases of AMICI and PEtab? Then this should be almost gtg (@jvanhoefer @dilpath ?).

pypesto/profile/profile.py
```diff
@@ -86,139 +88,37 @@ def create_next_guess(x, par_index, par_direction_, profile_options_,
...
-
-        if par_direction == 1:
-            stop_profile = (x_now[i_par] >= problem.ub_full[[i_par]]) \
-                or (current_profile.ratio_path[-1] < options.ratio_min)
-
-        if stop_profile:
-            break
+    # fill in the ProfilerResults at the right index
+    for r in ret:
+        result.profile_result.list[-1][r['par_ind']] = r['prof']
```
dilpath, line 122: Are these abbreviations already used elsewhere in pyPESTO? If not, could rename some things to be a bit more readable, e.g.

```suggestion
    for indexed_profile in indexed_profiles:
        result.profile_result.list[-1][indexed_profile['index']] = \
            indexed_profile['profile']
```

pypesto/profile/task.py
```diff
@@ -0,0 +1,78 @@
...
+from .walk_along_profile import walk_along_profile
+import pypesto.optimize
+
+logger = logging.getLogger(__name__)
+
+
+class ProfilerTask(Task):
+    """
+    A multistart optimization task, performed in `pypesto.minimize`.
+    """
```
dilpath, line 17:
```suggestion
    """A parameter likelihood profiling task."""
```

pypesto/profile/profile.py
```diff
@@ -86,139 +88,37 @@ def create_next_guess(x, par_index, par_direction_, profile_options_,
...
-                current_profile=current_profile,
-                problem=problem,
-                par_direction=par_direction,
-                optimizer=optimizer,
-                options=profile_options,
-                create_next_guess=create_next_guess,
-                global_opt=global_opt,
-                i_par=i_par)
+    # execute the tasks with Engine
+    ret = engine.execute(tasks)
```
dilpath, line 118:
```suggestion
    indexed_profiles = engine.execute(tasks)
```

To match the suggestion a few lines below.

pypesto/profile/task.py
```diff
@@ -0,0 +1,78 @@
...
+                problem=self.problem,
+                par_direction=par_direction,
+                optimizer=self.optimizer,
+                options=self.options,
+                create_next_guess=self.create_next_guess,
+                global_opt=self.global_opt,
+                i_par=self.i_par)
+
+        # return the ProfilerResult and the index of the parameter profiled
+        return {'prof': self.current_profile, 'par_ind': self.i_par}
```
dilpath, line 78:
```suggestion
        return {'profile': self.current_profile, 'index': self.i_par}
```

To match the suggestion where the result is combined in `parameter_profile()`.

pypesto/profile/walk_along_profile.py
```diff
@@ -0,0 +1,118 @@
...
+from ..objective.constants import GRAD
+from ..optimize import Optimizer, OptimizerResult
+from ..problem import Problem
+from .result import ProfilerResult
+from .options import ProfileOptions
+
+logger = logging.getLogger(__name__)
+
+
+def walk_along_profile(
```
dilpath, line 14: Were changes made to this function? Fine if not

test/profile/test_profile.py
```diff
@@ -72,6 +73,33 @@ def test_default_profiling(self):
...
+            profile.parameter_profile(
+                problem=self.problem,
+                result=self.result,
+                optimizer=self.optimizer,
+                next_guess_method='fixed_step',
+                engine=engine,)
+
+        # check results
+        for count, _engine in enumerate(engines):
+            for j in range(len(self.result.profile_result.list[0])):
```
dilpath, line 96:
```suggestion
            for j in range(1, len(self.result.profile_result.list[0])):
```

I guess the first can be skipped

dilpath (approved):
Looks good :+1:

pypesto/profile/walk_along_profile.py
PaulJonasJost: no changes were made as far as i remember

test/profile/test_profile.py
PaulJonasJost: why can the first be skipped? i think the first count can be skipped not j.

test/profile/test_profile.py
```diff
@@ -72,6 +73,33 @@ def test_default_profiling(self):
...
+            # in self.result.profile_result
+            profile.parameter_profile(
+                problem=self.problem,
+                result=self.result,
+                optimizer=self.optimizer,
+                next_guess_method='fixed_step',
+                engine=engine,)
+
+        # check results
+        for count, _engine in enumerate(engines):
```
dilpath, line 95:
```suggestion
        for count, _engine in enumerate(engines[1:]):
```

test/profile/test_profile.py
dilpath: Oops, meant `count`, added a new suggestion.

PR #542, opened 2020-12-16
Feature optimise across nodes

PaulJonasJost:
Hey, 
i implemented an engine, that can optimise across nodes. Tested it also for a small model (Benchmarkcollection: Zhao) and got a definite speedup. The exact speedup most likely depends on the complexity of the model.
I also wanted to add the engine to the test_engine.py file, but one cannot test it without using multiple nodes so that probably would not make too much sense? Any suggestions on that?
Lastly, some things to note: 

- In Bonna, srun and MPI do not work together and send out an error. This already happens as soon as one imports mpi4py so for the convenience, i did not add the MPIPoolEngine to the `__init__.py` file. Thus when using it one has to import it separately through `from pypesto.engine.mpi_pool import MPIPoolEngine`.

- Since srun does not work, one can use in the slurm file `mpiexec -n [number of nodes] python -m mpi4py.futures [filename]`. Additionally when using it, one has to clarify that all the work before the optimisation should be done on only one node. For this one can put the whole code as it is in a `if __name__ == '__main__':`-condition.

- Currently the optimisation works in the following way: one node is considered the master while the other are considered the workers. The master sends the workers batches (of the size of #cpus/node) of tasks, which they in turn execute with a multiprocessing Pool. Thus currently a whole node is only coordinating and not working. That would be a next step and i would be happy for any suggestions.


dweindl:
>     * Thus currently a whole node is only coordinating and not working.

That would be a major limitation. But simply telling `mpiexec` about the number of cores (per node) should take care of that. Not?

pypesto/engine/mpi_pool.py
```diff
@@ -0,0 +1,61 @@
...
+        n_procs = MPI.COMM_WORLD.Get_size()   # Size of communicator
+        logger.info(f"Performing parallel task execution on {n_procs} "
+                    f"nodes with chunksize of {self.chunksize}.")
+        with MPIPoolExecutor() as executor:
+            results_list = executor.map(work_per_node, chunks)
+
+        results_flat = []
+        for results in results_list:
+            for result in results:
+                results_flat.append(result)
```
dweindl, line 60:
```suggestion
            results_flat.extend(results)
```

pypesto/engine/mpi_pool.py
```diff
@@ -0,0 +1,61 @@
...
+    def __init__(self, chunksize: int = 2):
+        super().__init__()
+        self.chunksize: int = chunksize
+
+    def execute(self, tasks: List[Task]):
+        """Pickle tasks and distribute work over nodes."""
+
+        pickled_tasks = [pickle.dumps(task) for task in tasks]
+        # put the tasks in batchsizes of the number of cores used
+        chunks = [pickled_tasks[x:x+self.chunksize]
```
dweindl, line 48: Indexing error if `len(tasks)` is not a multiple of chunksize?

pypesto/engine/mpi_pool.py
```diff
@@ -0,0 +1,61 @@
...
+                  for x in range(0, len(pickled_tasks), self.chunksize)]
+
+        n_procs = MPI.COMM_WORLD.Get_size()   # Size of communicator
+        logger.info(f"Performing parallel task execution on {n_procs} "
+                    f"nodes with chunksize of {self.chunksize}.")
```
dweindl, line 53: Communicator size is not necessarily the number of nodes.

dweindl (commented):
Looks like a good start to me. There is obviously room for improving efficiency by going for dynamic scheduling, but that can be extended later.

Please extend the documentation on how to use it (as posted in the PR description).

Do you already have any benchmarking / scaling results? Would be interesting to see.

pypesto/engine/mpi_pool.py
PaulJonasJost: Checked that and if it is not a multiple, the last bit will be shorter so e.g. [1,2,3,4,5] with `chunksize=2` would be `[[1,2],[3,4],[5]]`.

pypesto/engine/mpi_pool.py
PaulJonasJost: so i do think it is, at least when i tested, if i allocated e.g. 5 nodes in my slurm file, the communicator size was also 5. Not sure whether this is exactly want you meant though

dweindl: It depends on how you invoke mpiexec. As [mentioned](https://github.com/ICB-DCM/pyPESTO/pull/542#issuecomment-746891307), I don't see any obvious problem with having one MPI rank per core instead of per node. There slightly more expensive communication should be overcompensated by not wasting one full node.

PaulJonasJost:
> > ```
> > * Thus currently a whole node is only coordinating and not working.
> > ```
> 
> That would be a major limitation. But simply telling `mpiexec` about the number of cores (per node) should take care of that. Not?

Had the idea as well, tested some things that did not work but will try a but more

pypesto/engine/mpi_pool.py
PaulJonasJost: might be because i invoked mpiexec with `-n [Number]`that it always only ran one process per node.

PaulJonasJost:
I updated the `MPIPoolEngine()'. The main problem was actually my `.slum`script i used before.
Also the engine is not in the `__init.py__` since otherwise whenever importing `pypesto.engine` it would also import `mpi4py` which produces problem when used in combination with a normal `srun` in a slurm-file.

pypesto/engine/mpi_pool.py
```diff
@@ -0,0 +1,45 @@
...
+class MPIPoolEngine(Engine):
+    """
+    Parallelize the task execution using multiprocessing.
+
+    Parameters
+    ----------
+    chunksize:
+        The chunksize the MPIPoolExecutor uses.
+        Default is two, but should always be set to
+        the number of cores used in each node.
```
dweindl, line 29:
```suggestion
```

dweindl (commented):
Looks good, but would be great to include a test case (which would also serve as usage example).

PaulJonasJost:
> Looks good, but would be great to include a test case (which would also serve as usage example).

Can i write a test which requires a slurm file and a cluster and would still pass any pytest test? That was the main issue i had.

dweindl:
> Can i write a test which requires a slurm file and a cluster and would still pass any pytest test? That was the main issue i had.

It shouldn't require slurm, just mpiexec should do the job, not? I see that it's a bit inconvenient. Easiest might be creating a test script that runs a (cheap - Rosenbrock?) optimization using the MPIPoolEngine and writing the result to a file. From a pytest test case you could then launch `mpiexec ... $thatOtherScript` via `subprocess.run` and verify the number of optimizations and final objective.

PaulJonasJost:
ok, so the test seems to work. @dweindl i noticed that if i run the slurm file in Bonna, it does not stop altough the program has finished.

doc/example/exampleBatchFile.slurm
```diff
@@ -0,0 +1,29 @@
...
+
+## Give your job a name to distinguish it from other jobs you run.
+#SBATCH --job-name=Test
+#SBATCH --partition=short
+## use --exclusive to get the whole nodes exclusively for this job
+#SBATCH --exclusive
+
+## Separate output and error messages into 2 files.
+## NOTE: %u=userID, %x=jobName, %N=nodeID, %j=jobID, %A=arrayID, %a=arrayTaskID
+#SBATCH --output=logBonna/%x-%j-%j.log
```
dweindl, line 11:
```suggestion
#SBATCH --output=log/%x-%j-%j.log
```

doc/example/exampleBatchFile.slurm
```diff
@@ -0,0 +1,29 @@
...
+## Specify how much time your job needs. (default: see partition above)
+#SBATCH --time=0-08:00:00   # Total time needed for job: Days-Hours:Minutes
+
+#SBATCH --ntasks=3   # 2 workers, 1 manager
+
+## Load the relevant modules needed for the job
+module purge
+module load OpenMPI/4.0.1-GCC-8.3.0-2.32
+module load Python/3.8.2-GCCcore-9.3.0
+module load SWIG/4.0.1-GCCcore-8.3.0 HDF5/1.10.6-gompi-2020a OpenBLAS/0.3.9-GCC-9.3.0 CMake/3.16.4-GCCcore-9.3.0 Boost/1.71.0-gompi-2019b
```
dweindl, line 24: I think this is too specific to be useful for the broader community. I'd suggest using the bare minimum, and ignoring any amici dependencies here.

doc/example/exampleBatchFile.slurm
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
```
dweindl, line 1: This file is not referenced anywhere, so chances are low that it will be helpful to anybody. I'd rather include it verbatim in the respective module docstring.

doc/example/example_MPIPool.py
```diff
@@ -0,0 +1,77 @@
+import pypesto
```
dweindl, line 1: Please add file-level docstring what this script is supposed to demonstrate.

doc/example/example_MPIPool.py
```diff
@@ -0,0 +1,77 @@
...
+    else:
+        assert_almost_equal(result1.optimize_result.list[0]['x'],
+                            result2.optimize_result.list[1]['x'],
+                            err_msg='The final parameter values '
+                                    'do not agree for the engines.')
+        assert_almost_equal(result1.optimize_result.list[1]['x'],
+                            result2.optimize_result.list[0]['x'],
+                            err_msg='The final parameter values '
+                                    'do not agree for the engines.')
+    print('also done with tests')
```
dweindl, line 77: Maybe you can clean that up a bit. Feels the relevant parts of the example are getting lost. (-> `def set_up_problem`)

Not sure how helpful it is to combine test and example in the same file, I'd rather split that.

doc/example/example_MPIPool.py
```diff
@@ -0,0 +1,77 @@
...
+    z = np.zeros_like(x)
+    for j in range(0, x.shape[0]):
+        for k in range(0, x.shape[1]):
+            z[j, k] = objective1([x[j, k], y[j, k]], (0,))
+
+    # create different optimizers
+    optimizer = optimize.FidesOptimizer()
+
+    # save optimizer trace
+    history_options = pypesto.HistoryOptions(trace_record=True)
```
dweindl, line 45: Unused?!

doc/example/example_MPIPool.py
```diff
@@ -0,0 +1,77 @@
...
+    problem1 = pypesto.Problem(objective=objective1, lb=lb, ub=ub,
+                               x_guesses=startpoints)
+
+    x = np.arange(-2, 2, 0.1)
+    y = np.arange(-2, 2, 0.1)
+    x, y = np.meshgrid(x, y)
+    z = np.zeros_like(x)
+    for j in range(0, x.shape[0]):
+        for k in range(0, x.shape[1]):
+            z[j, k] = objective1([x[j, k], y[j, k]], (0,))
```
dweindl, line 39: What's happening here?

test/optimize/test_optimize.py
```diff
@@ -199,3 +200,14 @@ def check_minimize(objective, library, solver, allow_failed_starts=False):
...
+
+def test_MPIPool_Engine():
+    """
+    Test the MPIPoolEngine by calling an example script with mpiexec.
+    """
+    # get the path to this file:
+    path = os.path.dirname(__file__)
+    # run the example file.
+    os.system(f"mpiexec -np 2 python -m mpi4py.futures"
+              f" {path}../../doc/example/example_MPIPool.py")
```
dweindl, line 213: Can this test ever fail? Please check the return code. Would suggest to use `subprocess`.

dweindl (changes requested):

doc/example/exampleBatchFile.slurm
PaulJonasJost: it was meant to be usable with the `example_MPIPool.py`.

doc/example/example_MPIPool.py
PaulJonasJost: removed.

doc/example/example_MPIPool.py
PaulJonasJost: removed.

doc/example/example_MPIPool.py
PaulJonasJost: Removed some stuff. If i exclude the tests, how should i then test?

doc/example/exampleBatchFile.slurm
PaulJonasJost: you mean in MPIPoolEngine() docstring?

test/optimize/test_optimize.py
PaulJonasJost: Tried it with subprocess (smhw did not work). Now os.sytems() return is saved and an error is raised if it is not a 0.

doc/example/example_MPIPool.py
yannikschaelte: docstrings always with double quotes

doc/example/exampleBatchFile.slurm
yannikschaelte: I think a bare minimum example batch file would be preferable (which works e.g. for Rosenbrock) would be preferable. In addition specifying e.g. modules required for Bonna for AMICI could be added somewhere else -- e.g. the versions are likely to change here.

dweindl: > it was meant to be usable with the example_MPIPool.py.

That shouldn't require the last line. For the other lines, chances are pretty low that they will work on any other than Bonna.

doc/example/exampleBatchFile.slurm
dweindl: I meant a file-level docstring, but the class documentation is good as well.

doc/example/example_MPIPool.py
dweindl: That would go into the pytest test case you added. You'd run the non-MPI optimization from there. The results would have to saved here for comparison (and then deleted).

test/optimize/test_optimize.py
dweindl: Must work with subprocess. If the return code is checked - ok. Not sure if output is shown if launched with subprocess.

doc/example/exampleBatchFile.slurm
```diff
@@ -0,0 +1,29 @@
...
+
+#SBATCH --ntasks=3   # 2 workers, 1 manager
+
+## Load the relevant modules needed for the job
+module purge
+module load OpenMPI/4.0.1-GCC-8.3.0-2.32
+module load Python/3.8.2-GCCcore-9.3.0
+module load SWIG/4.0.1-GCCcore-8.3.0 HDF5/1.10.6-gompi-2020a OpenBLAS/0.3.9-GCC-9.3.0 CMake/3.16.4-GCCcore-9.3.0 Boost/1.71.0-gompi-2019b
+
+source venv/bin/activate
```
dweindl, line 26: The reader may wonder where this venv comes from... Would remove.

doc/example/example_MPIPool.py
dweindl: (Would still suggest to move the whole problem setup to a separate functions. Makes it much easier to focus on the important parts of this example)

doc/example/exampleBatchFile.slurm
PaulJonasJost: i removed the example slurm file and added the `mpiexec -np ...`-command to the docstring of MPIPool, i think that this command is the most essential part.

doc/example/exampleBatchFile.slurm
PaulJonasJost: moved it to MPIPoolEngine

test/optimize/test_optimize.py
PaulJonasJost: So on my computer i can call the example_MPIPool.py with the os and it actually returns 0 if it runs without problems and 256 otherwise. But in the test on github it returns 256 with 
```
Primary job  terminated normally, but 1 process returned
a non-zero exit code.. Per user-direction, the job has been aborted.
-------------------------------------------------------
--------------------------------------------------------------------------
mpiexec detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:
```
but as far as i could read the implementation of `mpi4py.futures.MPIPoolExecutor`, it automatically performs a shutdown if you use with a `with` statement. any Idea what the problem could be?

dweindl: > . But in the test on github it returns 256 with [...]

From what I can see, `mpi4py` is simply not installed

```
----------------------------- Captured stderr call -----------------------------
/opt/hostedtoolcache/Python/3.8.7/x64/bin/python: Error while finding module specification for 'mpi4py.futures' (ModuleNotFoundError: No module named 'mpi4py')
/opt/hostedtoolcache/Python/3.8.7/x64/bin/python: Error while finding module specification for 'mpi4py.futures' (ModuleNotFoundError: No module named 'mpi4py')
```

.github/workflows/ci_push.yml
```diff
@@ -112,6 +112,8 @@ jobs:
       run: |
         .github/workflows/install_deps.sh ipopt
         pip install -e .[test,ipopt,dlib,pyswarm,cmaes,nlopt,fides]
+        sudo apt install libopenmpi-dev
```
dweindl, line 115: Dunno if @yannikschaelte has an opinion on including that in `install_deps.sh` 2 lines above? Either fine for me.

doc/example/example_MPIPool.py
```diff
@@ -0,0 +1,71 @@
...
+import numpy as np
+import scipy as sp
+import pypesto.optimize as optimize
+# you need to manually import the MPIPoolEninge
+from pypesto.engine.mpi_pool import MPIPoolEngine
+# the below is needed for testing purposes.
+from numpy.testing import assert_almost_equal
+
+def setup_rosen_problem(n_starts: int = 10):
+    """sets up the rosenbrock problem and return
```
dweindl, line 14:
```suggestion
    """Sets up the Rosenbrock problem and returns
```

pypesto/engine/mpi_pool.py
```diff
@@ -0,0 +1,40 @@
...
+def work(pickled_task):
+    """Unpickle and execute task."""
+    task = pickle.loads(pickled_task)
+    return task.execute()
+
+
+class MPIPoolEngine(Engine):
+    """
+    Parallelize the task execution using multiprocessing.
+    To be called with:
```
dweindl, line 23:
```suggestion

    To be called with:
```

.github/workflows/ci_push.yml
```diff
@@ -112,6 +112,8 @@ jobs:
       run: |
         .github/workflows/install_deps.sh ipopt
         pip install -e .[test,ipopt,dlib,pyswarm,cmaes,nlopt,fides]
```
dweindl, line 114:
```suggestion
        pip install -e .[test,ipopt,dlib,pyswarm,cmaes,nlopt,fides,MPI]
```

as you added it to setup.py already

.github/workflows/ci_push.yml
```diff
@@ -112,6 +112,8 @@ jobs:
...
+        sudo apt install libopenmpi-dev
+        pip install mpi4py
```
dweindl, line 116:
```suggestion
```
(in combination with suggestion above)

doc/example/example_MPIPool.py
```diff
@@ -0,0 +1,71 @@
...
+    # result2 is the way to call the optimization with MPIPoolEngine.
+    result1 = optimize.minimize(
+            problem=problem, optimizer=optimizer,
+            n_starts=n_starts, engine=pypesto.engine.MultiProcessEngine())
+    result2 = optimize.minimize(
+            problem=problem, optimizer=optimizer,
+            n_starts=n_starts, engine=MPIPoolEngine())
+
+
+    # starting here are the tests (not needed in your code)
```
dweindl, line 52: Would still prefer those outside the example, but if nobody else thinks so, okay to leave.

dweindl (approved):

.github/workflows/ci_push.yml
yannikschaelte: May be nicer to include in install_deps.sh, as I also use that to install things locally

doc/example/example_MPIPool.py
yannikschaelte: Improving upon that: "Set up and return" --> usually use imperative/infinitive in those one-liner docstrings.

.github/workflows/ci_push.yml
yannikschaelte: for consistency, maybe lower-case?

setup.py
```diff
@@ -44,6 +44,7 @@ def read(fname):
                       'pyswarm': ['pyswarm>=0.6'],
                       'cmaes': ['cma>=3.0.3'],
                       'fides': ['fides>=0.2.0'],
+                      'MPI': ['mpi4py'],
```
yannikschaelte, line 47: maybe add minimum version (just `pip show mpi4py`  your version?)

test/optimize/test_optimize.py
```diff
@@ -199,3 +200,16 @@ def check_minimize(objective, library, solver, allow_failed_starts=False):
     ]:
         assert np.isfinite(result.optimize_result.list[0]['fval'])
         assert result.optimize_result.list[0]['x'] is not None
+
+
+def test_MPIPool_Engine():
```
yannikschaelte, line 205: should probably all be lower-case

doc/example/example_MPIPool.py
yannikschaelte: I would agree that separating would be preferable. That would require storing the results. If that is easily possible (lost track of the status of hdf5 storage), that would be nice. Otherwise fine for me to leave it as is.

pypesto/engine/mpi_pool.py
```diff
@@ -0,0 +1,40 @@
...
+
+def work(pickled_task):
+    """Unpickle and execute task."""
+    task = pickle.loads(pickled_task)
+    return task.execute()
+
+
+class MPIPoolEngine(Engine):
+    """
+    Parallelize the task execution using multiprocessing.
```
yannikschaelte, line 22: MPI?

yannikschaelte: maybe even add a link to a good webpage on how mpi works etc.

yannikschaelte (approved):

pypesto/engine/mpi_pool.py
dweindl: maybe just https://mpi4py.readthedocs.io/en/stable/, anything deeper is probably not that helpful here

doc/example/example_MPIPool.py
PaulJonasJost: i was unsure whether this works with tempfile. Since the results would be stored in a subprocess, wont the tempfile not be also deleted upon completion?

dweindl: It would, but you can use a regular file and delet it at the end of the test.

PaulJonasJost: True, i will do so :+1:

.github/workflows/ci_push.yml
PaulJonasJost: added to install_deps.sh

doc/example/example_MPIPool.py
PaulJonasJost: Used the infinitive, hope that is gramatically correct.

doc/example/example_MPIPool.py
PaulJonasJost: I still performed both optimizations in the example file (since they need the same startpoints), saved the results and removed them later on in the `test_optimize`.

test/optimize/test_optimize.py
```diff
@@ -199,3 +203,45 @@ def check_minimize(objective, library, solver, allow_failed_starts=False):
...
+                            err_msg='The final parameter values '
+                                    'do not agree for the engines.')
+        assert_almost_equal(result1.optimize_result.list[1]['x'],
+                            result2.optimize_result.list[0]['x'],
+                            err_msg='The final parameter values '
+                                    'do not agree for the engines.')
+
+    # delete data
+    os.system('rm temp_result1.h5')
+    os.system('rm temp_result2.h5')
```
dweindl, line 247: Use os.remove() or os.unlink() for corresponding functions from Pathlib

test/optimize/test_optimize.py
```diff
@@ -199,3 +203,45 @@ def check_minimize(objective, library, solver, allow_failed_starts=False):
...
+    """
+    # get the path to this file:
+    path = os.path.dirname(__file__)
+    # run the example file.
+    subprocess.check_call(['mpiexec', '-np', '2', 'python', '-m',
+                           'mpi4py.futures',
+                           f'{path}/../..//doc/example/example_MPIPool.py'])
+
+    # read results
+    opt_result_reader = OptimizationResultHDF5Reader('temp_result1')
```
dweindl, line 220:
```suggestion
    opt_result_reader = OptimizationResultHDF5Reader('temp_result1.h5')
```
?

test/optimize/test_optimize.py
```diff
@@ -199,3 +203,45 @@ def check_minimize(objective, library, solver, allow_failed_starts=False):
...
+    result1 = opt_result_reader.read()
+    opt_result_reader = OptimizationResultHDF5Reader('temp_result2')
```
dweindl, line 222:
```suggestion
    opt_result_reader = OptimizationResultHDF5Reader('temp_result2.h5')
```
?

pypesto/engine/mpi_pool.py
```diff
@@ -0,0 +1,41 @@
...
+    task = pickle.loads(pickled_task)
+    return task.execute()
+
+
+class MPIPoolEngine(Engine):
+    """
+    Parallelize the task execution using mpi4py.
+    Link:  https://mpi4py.readthedocs.io/en/stable/
+    To be called with:
+    mpiexec -np #Workers+1 python -m mpi4py.futures YOURFILE.py
```
dweindl, line 25:
```suggestion
    Parallelize the task execution using 
    `mpi4py <https://mpi4py.readthedocs.io/en/stable/>`_.
    
    To be called with: 
    ``mpiexec -np #Workers+1 python -m mpi4py.futures YOURFILE.py``
```

dweindl (approved):
Looks good.

PR #545, opened 2020-12-29
do not delete optimizer history when writing to hdf5 file

beimbusch:


yannikschaelte:
Hi @beimbusch , sorry for having missed this PR somehow. Thanks for your contribution! As just this file seems to have been changed recently, could you update your PR to the latest version? We can have a look too there, this problem should be soonishly addressed.

pypesto/store/save_to_hdf5.py
```diff
@@ -130,17 +130,20 @@ def write(self, result: Result, overwrite=False):
             for start in result.optimize_result.list:
                 start_id = start['id']
                 start_grp = get_or_create_group(results_grp, start_id)
-                start['history'] = None  # TOOD temporary fix
                 if not overwrite:
                     for key in start.keys():
+                        if key == 'history':
+                            pass
```
FFroehlich, line 136: don't we want `continue` here? Same below.

beimbusch: Thanks for the feedback. I'll fix this as soon as possible and update the PR.

PR #546, opened 2021-01-11
Feature hdf5 profile sample

PaulJonasJost:
Hello,

i created a Reader and Writer for both Profiles and Samples. 
I do need to mention that there is still some slight bug, when writing into an already existing file. It then expectes the shape to be the same as before (e.g. a profile of a parameter which ad (20,1) before is now expected to have the same shape). I do have an Idea on how to fix this but this here is already working right now. One only has to create a new file or delete the old before as of now.


pypesto/store/read_from_hdf5.py
```diff
@@ -1,8 +1,38 @@
...
+                      parameter_id: h5py.Group) -> 'ProfilerResult':
+    """
+    Read HDF5 results per start.
+
+    Parameters
+    -------------
+    f:
+        The HDF5 result file
+    opt_id:
+        Specifies the start that is read from the HDF5 file
```
dweindl, line 22: Arguments do not match the function signature

pypesto/store/read_from_hdf5.py
```diff
@@ -1,8 +1,38 @@
...
+from ..profile.result import ProfilerResult
+from ..sample.result import McmcPtResult
 from ..problem import Problem
 from ..objective import Objective, ObjectiveBase
+import numpy as np
+
+
+def read_hdf5_profile(f: h5py.File,
+                      profile_id: h5py.Group,
+                      parameter_id: h5py.Group) -> 'ProfilerResult':
```
dweindl, line 13: Seem unintuitive that an ID would be of type h5py.Group

pypesto/store/read_from_hdf5.py
```diff
@@ -120,3 +150,94 @@ def read(self) -> Result:
...
+            HDF5 result file name
+        """
+        self.storage_filename = storage_filename
+        self.results = Result()
+
+    def read(self) -> Result:
+        """
+        Read HDF5 result file and return pyPESTO result object.
+        """
+        sampleResult = {}
```
dweindl, line 180:
```suggestion
        sample_result = {}
```
+ following occurrences

pypesto/store/read_from_hdf5.py
```diff
@@ -120,3 +150,94 @@ def read(self) -> Result:
...
+            for key in f['/sampling/results']:
+                sampleResult[key] = \
+                    f[f'/sampling/results/{key}'][:]
+            for key in f['/sampling/results'].attrs:
+                sampleResult[key] = \
+                    f['/sampling/results'].attrs[key]
+        try:
+            self.results.sample_result = McmcPtResult(**sampleResult)
+        except ValueError:
+            # TODO : Warning here
```
dweindl, line 194: TODO

pypesto/store/save_to_hdf5.py
```diff
@@ -145,3 +145,121 @@ def write(self, result: Result, overwrite=False):
...
+        """
+        Parameters
+        ----------
+
+        storage_filename: str
+            HDF5 result file name
+        """
+        self.storage_filename = storage_filename
+
+    def write(self, result: Result, overwrite=False):
```
dweindl, line 170:
```suggestion
    def write(self, result: Result, overwrite: bool = False):
```

pypesto/store/save_to_hdf5.py
```diff
@@ -145,3 +145,121 @@ def write(self, result: Result, overwrite=False):
...
```
dweindl, line 224:
```suggestion
    def write(self, result: Result, overwrite: bool = False):
```

pypesto/store/save_to_hdf5.py
```diff
@@ -145,3 +145,121 @@ def write(self, result: Result, overwrite=False):
...
+                        if not overwrite:
+                            for key in parameter_profile.keys():
+                                if (key in result_grp.keys()
+                                        or key in result_grp.attrs):
+                                    raise Exception("The file already exists"
+                                                    " and contains information"
+                                                    " about optimization"
+                                                    " result. If you wish"
+                                                    " to overwrite it, set"
+                                                    " overwrite=True.")
```
dweindl, line 257: Reduce nesting depth. Also seems to be redundant with some function further up. Move to separate function?

dweindl (changes requested):
Good to have HDF5 IO support for profiling/sampling results.

Please also include a test case writing some result, reading it back in and ensuring that it is the same as the original one.

pypesto/store/read_from_hdf5.py
PaulJonasJost: To me this was the most explanatory name, since `profile_id` specifies the profile start and `parameter_id` specifies the parameter index. But i am open to change the names if you suggest something :)

pypesto/store/read_from_hdf5.py
PaulJonasJost: added 
```python
logger.warning(
                "Could not create a McmcPtResult from sample_result."
                "Will continue with a list.")
```
ok like that?

pypesto/store/save_to_hdf5.py
PaulJonasJost: Reduced the nesting depth by one.

pypesto/store/read_from_hdf5.py
PaulJonasJost: I checked again and you are right, `profile_id` is not a h5py.Group but a str, changed that for `profile_id`, `parameter_id` as wellt as `opt_id` and the function below.

pypesto/store/read_from_hdf5.py
dweindl: Moderately informative. Why would that happen? Append the original exception?

pypesto/store/read_from_hdf5.py
```diff
@@ -120,3 +158,96 @@ def read(self) -> Result:
...
+            for key in f['/sampling/results'].attrs:
+                sample_result[key] = \
+                    f['/sampling/results'].attrs[key]
+        try:
+            self.results.sample_result = McmcPtResult(**sample_result)
+        except ValueError:
+            logger.warning(
+                "Could not create a McmcPtResult from sample_result."
+                "Will continue with a list.")
+            pass
```
dweindl, line 205:
```suggestion
            logger.warning(
                "Could not create a McmcPtResult from sample_result. "
                "Will continue with a list.")
```

pypesto/store/read_from_hdf5.py
dweindl: Well, and it's not a list, but a dict. Please also update the return type typehint accordingly.

Not too sure how helpful it is to return the dict instead here. Either all callers will have to handle both types, or will probably fail. Maybe it's easier to just fail here?

PaulJonasJost: you might be right that it is easier to fail here. Although since McmcPtResult inherits from dict, it should actually still work fine i think.

dweindl: >  But there is no real reason why self.results.sample_result = McmcPtResult(**sampleResult)should fail so i think i might as well remove the try... except?

Agreed

dweindl (approved):
Good for me now.

pypesto/store/save_to_hdf5.py
PaulJonasJost: added a check_overwrite function for this, which now also deletes the files if overwrite=True and the files exist. Which solves the bug i mentioned in the initial PR.

pypesto/store/read_from_hdf5.py
PaulJonasJost: Done :+1:

PR #490, opened 2020-10-29
Feature MCMC prediction confidence intervals - close #458

elbaraim:
This branch features:
- Calculation and Visualization of parameter confidence intervals
- Calculation and Visualization of prediction confidence intervals (MCMC)

PR #557, opened 2021-01-20
Ensembles

paulstapor:
The idea of this PR is adding functionality for ensemble analysis, i.e., one can create an parameter collection (=ensemble or sample) from a csv file and run predictions for the whole collection and compute summaries of those, i.e., means, stds, medians, percentiles... more to be added.

If someone has suggestions what to add: I'm happy to include that


paulstapor:
On the to do list here:
 - [ ] add a notebook, which documents predictions and ensembles
 - [ ] add visualizations for uncertainties (at least of parameters)
 - [ ] add more convenience stuff for PEtab (e.g.: write_to_PEtab for Predictions and CollectionProdictions)
 - [ ] add dimension reduction analysis (helpful for large ensembles of very large models), such as PCA and UMAP

elbaraim:
> On the to do list here:
> 
>     * [x]  add a notebook, which documents predictions and ensembles
> 
>     * [ ]  add visualizations for uncertainties (at least of parameters)
> 
>     * [ ]  add more convenience stuff for PEtab (e.g.: write_to_PEtab for Predictions and CollectionProdictions)
> 
> 
> Anyone feels like adding something here?

Hi @paulstapor
Regarding the point `add visualizations for uncertainties (at least of parameters)`
I suggest to combine it with https://github.com/ICB-DCM/pyPESTO/pull/490

paulstapor:
Hi, okay, yes, I'm happy not to add visualization here atm. Makes merging easier, as less stuff to implement... ;)
however, there's one idea for plotting which I have, which would be something similar to the parallel coordinate plot...

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+from .constants import (PREDICTOR, PREDICTION_ID, PREDICTION_RESULTS,
+                        PREDICTION_ARRAYS, PREDICTION_SUMMARY, OUTPUT,
+                        OUTPUT_SENSI, TIMEPOINTS, X_VECTOR, NX, X_NAMES,
+                        NVECTORS, VECTOR_TAGS, PREDICTIONS, MODE_FUN,
+                        CollectionType, COLLECTION_TYPE, MEAN, MEDIAN,
+                        STANDARD_DEVIATION, PERCENTILE, SUMMARY, LOWER_BOUND,
+                        UPPER_BOUND)
+
+
+class CollectionPrediction:
```
FFroehlich, line 15: Any reason to call this class/file "collection" instead of "ensemble"?

paulstapor: The idea was to have a term covering "ensemble" and "sample" ... ;)

paulstapor: For me, an "ensemble" is something that explicitly doesn't attempt to be representative, at least that's the way it way recently used in literature... So I didn't want to go for "ensemble", as a lot of the stuff can be reused 1:1 for samples...
However: no big issue to just rename this all to "ensemble", if that less misleading...

elbaraim:
> Hi, okay, yes, I'm happy not to add visualization here atm. Makes merging easier, as less stuff to implement... ;)
> however, there's one idea for plotting which I have, which would be something similar to the parallel coordinate plot...

This is how the visu for the parameter confidence intervals is currently implemented (feature branch)
Maybe you like it and it could be used :) or it could be combined with your idea
Screenshot from 2021-01-22 09-54-52

paulstapor:
Should be pretty easy to implement that accordingly... The technique used for this is the same for samples and ensembles. So that can be combined... the question is just which PR to merge first. If we merge this one here first (it's basically done now, I think, except for another idea which I have, which is tailored to large-scale models), I can help you implementing samples in this collection or ensemble framework and adapt visualization, if necessary...

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+
+    def __iter__(self):
+        yield PREDICTOR, self.predictor
+        yield PREDICTION_ID, self.prediction_id
+        yield PREDICTION_RESULTS, self.prediction_results
+        yield PREDICTION_ARRAYS, self.prediction_arrays
+        yield PREDICTION_SUMMARY, {i_key: dict(self.prediction_summary[i_key])
+                                   for i_key in self.prediction_summary.keys()}
+        yield LOWER_BOUND, self.lower_bound
+        yield UPPER_BOUND, self.upper_bound
```
dilpath, line 70: How is this `__iter__` used? Is there an example? Not sure what it can be used for yet.

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+        them as a member to the CollectionPrediction objects.
+        """
+        # prepare for storing results over all predictions
+        output = []
+        output_sensi = []
+        timepoints = []
+
+        for result in self.prediction_results:
+            # stack outputs, output sensitivities and timepoints to one array
+            # use first element as dummy, to see if outputs ahve been computed
```
dilpath, line 84:
```suggestion
            # use first element as dummy, to see if outputs have been computed
```

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+        # check if prediction results are available
+        if not self.prediction_results:
+            raise ArithmeticError('Cannot compute summary statistics from '
+                                  'empty prediction results.')
+        n_conditions = len(self.prediction_results[0].conditions)
+
+        def _stack_outputs(ic: int):
+            """
+            This function groups outputs for different parameter vectors of
+            one ensemble together, if the belong to the same simulation
```
dilpath, line 142:
```suggestion
            one ensemble together, if they belong to the same simulation
```

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+            for i_key in cond_lists.keys()
+        }
+
+        # also return the object
+        return self.prediction_summary
+
+
+class Collection:
+    """
+    A collection is of a thin wrapper around an numpy array.
```
dilpath, line 236:
```suggestion
    A collection is a thin wrapper around an numpy array.
```

Add description of what it can be used to do (e.g., why use this instead of a numpy array)? Or what it's usual use case (e.g. used by pyPESTO to handle a set of parameter vectors).

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+    """
+
+    def __init__(self,
+                 x_vectors: np.ndarray,
+                 x_names: Sequence[str] = None,
+                 vector_tags: Tuple[int, int] = None,
```
dilpath, line 242:
```suggestion
                 vector_tags: Sequence[Tuple[int, int]] = None,
```

Should it be one tuple per vector?

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+        # create DataFrame
+        parameter_identifiability = pd.DataFrame(parameter_identifiability)
+        parameter_identifiability.index = \
+            parameter_identifiability['parameterId']
+
+        return parameter_identifiability
+
+
+def read_from_csv(path: str,
+                  sep: str = '\t',
```
dilpath, line 437: Seems odd (as in the previous `Prediction` PR) to call this a `csv` method, since it defaults to `tsv`.

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+    """
+    A collection is of a thin wrapper around an numpy array.
+    """
+
+    def __init__(self,
+                 x_vectors: np.ndarray,
+                 x_names: Sequence[str] = None,
+                 vector_tags: Tuple[int, int] = None,
+                 coll_type: CollectionType = None,
+                 predictions: Sequence[CollectionPrediction] = None,
```
dilpath, line 244: Should this be here?

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+        yield X_NAMES, self.x_names
+        yield NVECTORS, self.n_vectors
+        yield VECTOR_TAGS, self.vector_tags
+        yield COLLECTION_TYPE, self.coll_type
+        yield PREDICTIONS, self.predictions
+        yield SUMMARY, self.summary
+        yield LOWER_BOUND, self.lower_bound
+        yield UPPER_BOUND, self.upper_bound
+
+    def predict(self,
```
dilpath, line 318: Could move prediction bounds to this method, and remove from `CollectionPrediction`

pypesto/petab/importer.py
```diff
@@ -320,16 +320,31 @@ def create_prediction(self,
...
+        # which can then be stored in the prediction result
+        edata_conditions = objective.amici_object_builder.petab_problem.\
+            get_simulation_conditions_from_measurement_df()
+        if 'preequilibrationConditionId' not in list(edata_conditions.columns):
+            preeq_dummy = [''] * edata_conditions.shape[0]
+            edata_conditions['preequilibrationConditionId'] = preeq_dummy
+        edata_conditions.drop_duplicates(inplace=True)
+        condition_ids = [
+            edata_conditions.loc[id, 'preequilibrationConditionId'] + '::' +
+            edata_conditions.loc[id, 'simulationConditionId']
```
dilpath, line 333:
```suggestion
        if PREEQUILIBRATION_CONDITION_ID not in list(edata_conditions.columns):
            preeq_dummy = [''] * edata_conditions.shape[0]
            edata_conditions[PREEQUILIBRATION_CONDITION_ID] = preeq_dummy
        edata_conditions.drop_duplicates(inplace=True)
        condition_ids = [
            edata_conditions.loc[id, PREEQUILIBRATION_CONDITION_ID] + '::' +
            edata_conditions.loc[id, SIMULATION_CONDITION_ID]
```

with 
```python
from petab.C import PREEQUILIBRATION_CONDITION_ID, SIMULATION_CONDITION_ID
```

What is `'::'`? Should this be a constant?

pypesto/petab/importer.py
```diff
@@ -475,6 +490,54 @@ def rdatas_to_simulation_df(
...
+            predictor: AmiciPredictor = None
+    ) -> pd.DataFrame:
+        """
+        If a PEtab problem is simulated without post-processing, then the
+        result can be cast into a PEtab measurement or simulation dataframe
+
+        Parameters
+        ----------
+        prediction:
+            A prediction resul as produced created by an AmiciPredictor
```
dilpath, line 505:
```suggestion
            A prediction result as produced by an AmiciPredictor
```

pypesto/petab/importer.py
```diff
@@ -475,6 +490,54 @@ def rdatas_to_simulation_df(
         return self.rdatas_to_measurement_df(rdatas, model).rename(
             {petab.MEASUREMENT: petab.SIMULATION})
 
+    def petab_prediction_to_measurement_df(
```
dilpath, line 493:
```suggestion
    def prediction_to_petab_measurement_df(
```

pypesto/petab/importer.py
```diff
@@ -475,6 +490,54 @@ def rdatas_to_simulation_df(
...
+    def petab_prediction_to_simulation_df(
+            self,
+            prediction: PredictionResult,
+            predictor: AmiciPredictor = None
+    ) -> pd.DataFrame:
+        """Same as `petab_prediction_to_measurement_df`, execpt a petab
+        simulation dataframe is created, i.e. the measurement column label is
+        adjusted.
+        """
+        return self.petab_prediction_to_measurement_df(
```
dilpath, line 537:
```suggestion
    def prediction_to_petab_simulation_df(
            self,
            prediction: PredictionResult,
            predictor: AmiciPredictor = None
    ) -> pd.DataFrame:
        """Same as `prediction_to_petab_measurement_df`, except a PEtab
        simulation dataframe is created, i.e. the measurement column label is
        adjusted.
        """
        return self.prediction_to_petab_measurement_df(
```

test/petab/test_amici_prediction.py
```diff
@@ -296,7 +299,55 @@ def test_petab_prediction():
...
     prediction = importer.create_prediction()
 
-    # run test
+    # ===== run test for prediction ===========================================
     p = prediction(np.array(petab_problem.x_nominal_free_scaled),
                    sensi_orders=(0, 1))
     check_outputs(p, out=(0, 1), n_cond=1, n_timepoints=10, n_obs=1, n_par=2)
+    # check outputs for simulation and measurement dataframes
+    importer.petab_prediction_to_measurement_df(p, prediction)
+    importer.petab_prediction_to_simulation_df(p, prediction)
```
dilpath, line 308:
```suggestion
    importer.prediction_to_petab_measurement_df(p, prediction)
    importer.prediction_to_petab_simulation_df(p, prediction)
```

test/petab/test_amici_prediction.py
```diff
@@ -296,7 +299,55 @@ def test_petab_prediction():
...
+    assert isinstance(summary, dict)
+    assert set(summary.keys()) == {'mean', 'std', 'median', 'percentile 10',
+                                   'percentile 25', 'percentile 75',
+                                   'percentile 90'}
+
+    parameter_identifiability = ensemble.check_identifiability()
+    assert isinstance(parameter_identifiability, pd.DataFrame)
+
+    # perform a prediction for the ensemble
+    ensemble_prediction = ensemble.predict(predictor=prediction)
```
dilpath, line 331: `predictor=prediction` seems wrong? Maybe some things could be renamed

dilpath (approved):
:+1: Looks good but there is a lot here, would be good to request another reviewer I think.

Thanks for this!

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+                        STANDARD_DEVIATION, PERCENTILE, SUMMARY, LOWER_BOUND,
+                        UPPER_BOUND)
+
+
+class CollectionPrediction:
+    """
+    A collection prediction consists of a collection, i.e., a set of parameter
+    vectors and their identifiers such as a sample or an ensemble, and a
+    prediction function. It can be attached to a collection-type object
+    """
```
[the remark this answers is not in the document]
paulstapor: I find it very, very hard to come up with a reasonable name for this... One could also go for "ensemble", as @FFroehlich suggested... no clue... 
I also had this class named PredictionCollection and CollectionPrediction already.... I find CellectionPrediction a bit better, but no strong opinion...

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+
+    def __init__(self,
+                 predictor: Callable,
```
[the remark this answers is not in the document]
paulstapor: depends a bot on how the final API will look like.... :D
But I think its not wrong to store the predictor function...  That may later be helpful for some features...

pypesto/collections/collections.py
paulstapor: Adding such a function makes your class iterable, i.e., you can directly apply dict() to instances of this class, which creates a dictionary from the object. I find that rather handy, so I added this __iter__

pypesto/collections/collections.py
```diff
@@ -0,0 +1,484 @@
...
+            Prediction function, e.g., an AmiciPredictor
+        prediction_id:
+            Identifier for the predictions
+        prediction_results:
+            List of Prediction results
+        lower_bound:
+            array of potential lower bounds for the parameters, should have the
+            same shape as the output of the predictions, i.e., a list of numpy
+            array (one list entry per condition), with the arrays having the
+            shape of n_timpoints x n_outputs for each condition
```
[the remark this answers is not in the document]
paulstapor: Ah... no, the idea is: Add bounds to your prediction, to specify when you would say that a specific prediction is "non-identifiable". That's kinda analogous to parameters, but for model predictions... That's at least what it's meant for. According functionalities, examples, or checks are not yet implemented, but it will be useful in some not too long future...

test/petab/test_amici_prediction.py
paulstapor: yeah, agreed, the naming is inconsistent. Will change :)

pypesto/collections/collections.py
paulstapor: in practice, we will always want to use tsv, I think. But for some reason, csv is the most common term for a file formatted like this... Just as Pandas also calls its function `read_csv` and `to_csv` ... I just used an analogous naming...

pypesto/collections/collections.py
paulstapor: hm... design question...
Do we want to store predictions from a collection in that collection object? I would vote for "yes", as one might want to use multiple prediction functions on the same collection object. ... But well, this is up to discussion. If a majority is against this, we can keep things separated...

pypesto/petab/importer.py
paulstapor: '::' is meant to be a separator, which can be used later to reconstruct the ids of the respective conditions. If one splits with respect to this, one gets more info about the simulations...

pypesto/collections/collections.py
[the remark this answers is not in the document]
dilpath: When I first saw it and made the suggestion, I thought it meant "Collection of predictions", not "Predictions of a collection (of parameter vectors)", so agreed, fine as is :+1:

pypesto/collections/collections.py
dilpath: Agreed but e.g. pandas `read_csv` defaults to `read_csv(..., sep=',')`, not `read_csv(..., sep='\t')`. Not so important but may confuse users who are used to CSV meaning comma by default.

pypesto/collections/collections.py
dilpath: Then we could have `Collection`, `CollectionPrediction`, and `CollectionPredictionCollection` :grin: 

Fine as is then :+1:

pypesto/petab/importer.py
dilpath: Makes sense but then it should be a constant I guess? e.g. `CONDITION_DELIMITER = '::'`

pypesto/visualize/collections.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+def identifiability_overview(collection: Collection,
+                             ax: Optional[plt.Axes] = None,
+                             size: Optional[Tuple[float]] = (12, 6)):
+    """
+    Plots an overview about how many parameters hit the parameter bounds based
+    on a collection of parameters. confidence intervals/credible ranges are
+    computed via the collection mean plus/minus 1 standard deviation.
+    This highlevel routine expects (atm) a collection object as input.
```
FFroehlich, line 21:
```suggestion
    This highlevel routine expects a collection object as input.
```

test/petab/test_amici_prediction.py
```diff
@@ -5,14 +5,17 @@
 import amici
```
FFroehlich, line 5: ^ intentional that submodule was added above?

pypesto/prediction/prediction.py
```diff
@@ -202,6 +209,14 @@ def write_to_h5(self,
                     f.create_dataset(f'{i_cond}/{OUTPUT_SENSI}',
                                      data=cond.output_sensi)
 
+    def write_to_petab_file(self,
+                            output_file: str,
+                            measurement_df: pd.DataFrame):
+        """
+        This function writes prediction results to a petab simulation file,
+        which is shaped similarly to the measurement DataFrame, which is passed
+        """
```
FFroehlich, line 218: Uuuh, I would argue this function doesn't really do anything ;). In case this is intended as abstract method, please raise `NotImplementedError()`.

pypesto/collections/collections.py
FFroehlich: Hmm, I would still call parameter sets from sampling an "ensemble". Looking at https://en.wikipedia.org/wiki/Ensemble_learning they list Bayesian model averaging also as ensemble method. Similarly https://en.wikipedia.org/wiki/Statistical_ensemble_(mathematical_physics) includes Monte Carlo samplings. Accordingly, I would say that samples are also ensembles, but not the other way around so calling this "Ensemble" is probably more intuitive for most. Happy to hear other opinions though.

pypesto/collections/collections.py
FFroehlich: maybe add this as comment in the function documentation ;)

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+        lower and the upper bound are violated
+
+        Returns
+        -------
+        parameter_identifiability:
+            DataFrame indicating parameter identifiability based on mean
+            plus/minus standard deviations and parameter bounds
+        """
+        # first check if summary was computed. If not, do so
+        if self.summary is None:
```
FFroehlich, line 399: this won't update if the user changes the underlying collection after running check_identifiability the first time.

FFroehlich (approved):

pypesto/prediction/prediction.py
paulstapor: Nope, that's a "I started implementing this and then ended up doing it another way" :joy: 
Good you found it. Yeah, I may rework this one. Currently "PEtab" export only works for predictions without post-processing. This function could allow to do this also for post-processed settings... Let's see. Either I'll remove or I'll finish it...

pypesto/collections/collections.py
dweindl: > samples are also ensembles, but not the other way around so calling this "Ensemble" is probably more intuitive for most

I consider myself as part of *most* :)

paulstapor: No, totally fine. I was really struggling with a good naming scheme. If literature also lists samples as ensembles, it really makes sense to adapt this. I'll do so.

pypesto/collections/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
```
dweindl, line 1: I think `pypesto.collections` is missing in sphinx docs

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+                        OUTPUT_SENSI, TIMEPOINTS, X_VECTOR, NX, X_NAMES,
+                        NVECTORS, VECTOR_TAGS, PREDICTIONS, MODE_FUN,
+                        CollectionType, COLLECTION_TYPE, MEAN, MEDIAN,
+                        STANDARD_DEVIATION, PERCENTILE, SUMMARY, LOWER_BOUND,
+                        UPPER_BOUND)
+
+
+class CollectionPrediction:
+    """
+    A collection prediction consists of a collection, i.e., a set of parameter
```
dweindl, line 17: `set`: In the python context I'd assume this collection to be unordered and having unique parameter vectors, which is not necessarily the case, right?

pypesto/collections/collections.py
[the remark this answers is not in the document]
dweindl: If used, it would be good to extend the typehint and/or docs to describe the expected input/output

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+        yield PREDICTOR, self.predictor
+        yield PREDICTION_ID, self.prediction_id
+        yield PREDICTION_RESULTS, self.prediction_results
+        yield PREDICTION_ARRAYS, self.prediction_arrays
+        yield PREDICTION_SUMMARY, {i_key: dict(self.prediction_summary[i_key])
+                                   for i_key in self.prediction_summary.keys()}
+        yield LOWER_BOUND, self.lower_bound
+        yield UPPER_BOUND, self.upper_bound
+
+    def condense_to_arrays(self):
```
dweindl, line 72:
```suggestion
    def _condense_to_arrays(self):
```
Not sure, but maybe something to be considered as private? (Also check for others)

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+            OUTPUT: output,
+            OUTPUT_SENSI: output_sensi,
+            TIMEPOINTS: np.stack(timepoints, axis=-1)
+        }
+
+    def compute_summary(self,
+                        percentiles_list: Sequence[int] = (5, 20, 80, 95)
+                        ) -> Dict:
+        """
+        This function computes the mean, the median, the standard deviation
```
dweindl, line 117:
```suggestion
        Compute the mean, the median, the standard deviation
```

pypesto/collections/collections.py
paulstapor: Right... Question is whether one wants to leave this to the user... Probably not. The computational effort to compute this is minor, so we can really do this every time...

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+            simulation condition, and stacks them in one array
+            """
+            # Were output sensitivitiess computed
+            if self.prediction_results[0].conditions[ic].output_sensi is None:
+                return None
+            # stack predictions
+            output_sensi_list = [prediction.conditions[ic].output_sensi
+                                 for prediction in self.prediction_results]
+            # stack into one numpy array
+            return np.stack(output_sensi_list, axis=-1)
```
dweindl, line 167: Looks rather similar to `condense_to_arrays`. Can this be combined?

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+        """
+        # check if prediction results are available
+        if not self.prediction_results:
+            raise ArithmeticError('Cannot compute summary statistics from '
+                                  'empty prediction results.')
+        n_conditions = len(self.prediction_results[0].conditions)
+
+        def _stack_outputs(ic: int):
+            """
+            This function groups outputs for different parameter vectors of
```
dweindl, line 141:
```suggestion
            Group outputs for different parameter vectors of
```

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+                return None
+            # stack predictions
+            output_list = [prediction.conditions[ic].output
+                           for prediction in self.prediction_results]
+            # stack into one numpy array
+            return np.stack(output_list, axis=-1)
+
+        def _stack_outputs_sensi(ic: int):
+            """
+            This function groups output sensitivities for different parameter
```
dweindl, line 156:
```suggestion
            Group output sensitivities for different parameter
```

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+                return None
+            # stack predictions
+            output_sensi_list = [prediction.conditions[ic].output_sensi
+                                 for prediction in self.prediction_results]
+            # stack into one numpy array
+            return np.stack(output_sensi_list, axis=-1)
+
+        def _compute_summary(tmp_array, percentiles_list):
+            """
+            This function computes means, standard deviation, median, and
```
dweindl, line 171:
```suggestion
            Compute means, standard deviation, median, and
```

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+            for i_key in cond_lists.keys()
+        }
+
+        # also return the object
+        return self.prediction_summary
+
+
+class Collection:
+    """
+    A collection is a thin wrapper around an numpy array.
```
dweindl, line 236: ... providing any additional functionality? One could argue about "thin"...

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+            parameter vectors of the collection, in the format
+            n_parameter x n_vectors
+        x_names:
+            Names or identifiers of the parameters
+        vector_tags:
+            Additional tag, which adds information about the the parameter
+            vectors of the form (optimization_run, optimization_step) if the
+            collection is created from an optimization result or
+            (sampling_chain, sampling_step) if the collection is created from a
+            sampling result.
```
dweindl, line 262: Does this have to be restricted to `(int, int)`? Might be convenient to have more flexibility. But also fine as is.

pypesto/collections/collections.py
```diff
@@ -0,0 +1,497 @@
...
+        for perc in percentiles_list:
+            summary[f'{PERCENTILE} {perc}'] = np.percentile(self.x_vectors,
+                                                            perc, axis=1)
+        # store and return results
+        self.summary = summary
+        return summary
+
+    def check_identifiability(self) -> pd.DataFrame:
+        """
+        This functions uses ensemble mean and standard deviation to assess
```
dweindl, line 387:
```suggestion
        Use ensemble mean and standard deviation to assess
```

dweindl (commented):
:+1: 

This will come in handy soon...

pypesto/collections/collections.py
paulstapor: no necessarily private, also tested (to be called) from the outside.

pypesto/collections/collections.py
paulstapor: In principle, there is no ordering assumed. Unique parameters vectors: not necessarily.... There may be applications where non-uniqueness may be helpful, so I don't want to enforce it...

pypesto/collections/collections.py
paulstapor: hm... maybe... They are similar, indeed, but there is one important difference: the `_stack_...` functions keep the substructure of having conditions, and therefore allow for different observables per condition. `condense_to_arrays` doesn't do so, it's the brute force approach crunching everything together. Therefore, it's really only usable if all conditions have the same shape. Will document this a bit more...
I added those functions, because I'm pretty sure they will be helpful soon in one or the other way. condense_to_arrays does more than only stacking, however... Not quite, would rather like to wait until the first person really uses them. Then it's still minor effort to change and it will be easier to see whether or not these functions need to be adapted...

pypesto/collections/collections.py
paulstapor: No, doesn't really have to. Many things here are beta version like... That was the most typical example I could think of. The whole ensemble thing is just about to start, much more will come in the future (e.g., create ensembles from pyPESTO/parPE optimization results in an automated fashion). Some of the data members I added try to already account for that, but it's not always easy to anticipate how to best implement things which are yet to come...

PR #559, opened 2021-01-20
Fix nlopt optimization test by ignoring it

yannikschaelte:
One of the algorithms (GD_STOGO_RAND, id 9) failed in like 40% of the cases with a non-finite return value. Therefore ignoring. Needs to be investigated if someone wants to use that exact algorithm.


jvanhoefer:
a) do you want to push into `master` or into develop? 
b) do you see this as a short term or long term solution?

yannikschaelte:
Thanks, develop should have been the target. I see it as a solution until someone complains, bc probably that specific optimizer is just not too robust.

jvanhoefer (approved):

PR #565, opened 2021-02-04
Ensemble viz

paulstapor:
This PR should complete very most of the ensemble functionality, including nice plotting and some more convenience things, such as storing ensemble_predictions directly to h5 files

### DIFF
--- a/pypesto/ensemble/covariance_analysis.py
+++ b/pypesto/ensemble/covariance_analysis.py
@@ -1,0 +1,31 @@
+import numpy as np
+import pandas as pd
+from typing import Sequence, Tuple, Callable, Dict
+
+from ..prediction import PredictionResult, PredictionConditionResult
+from .constants import (PREDICTOR, PREDICTION_ID, PREDICTION_RESULTS,
+                        PREDICTION_ARRAYS, PREDICTION_SUMMARY, OUTPUT,
+                        OUTPUT_SENSI, TIMEPOINTS, X_VECTOR, NX, X_NAMES,
+                        NVECTORS, VECTOR_TAGS, PREDICTIONS, MODE_FUN,
+                        EnsembleType, ENSEMBLE_TYPE, MEAN, MEDIAN,
+                        STANDARD_DEVIATION, PERCENTILE, SUMMARY, LOWER_BOUND,
+                        UPPER_BOUND)
+
+
+def get_covariance_matrix_parameters():
+    pass
+
+def get_covariance_matrix_predictions():
+    pass
+
+def get_spectral_decomposition_parameters():
+    pass
+
+def get_spectral_decomposition_predictions():
+    pass
+
+def _get_covariance_matrix_lowlevel():
+    pass
+
+def _get_spectral_decomposition_lowlevel():
+    pass
--- a/pypesto/ensemble/covariance_analysis.py
+++ b/pypesto/ensemble/covariance_analysis.py
@@ -1,11 +1,11 @@
 import numpy as np
-from typing import Union
+from typing import Union, Tuple
 
 from ..ensemble import Ensemble, EnsemblePrediction
 from .utils import get_prediction_dataset
 
 
-def get_covariance_matrix_parameters(ens: Ensemble):
+def get_covariance_matrix_parameters(ens: Ensemble) -> np.ndarray:
     """
     Compute the covariance of ensemble parameters.
 
@@ -21,7 +21,7 @@ def get_covariance_matrix_parameters(ens: Ensemble):
     """
 
     # call lowlevel routine using the parameter ensemble
-    return get_covariance_matrix_lowlevel(dataset=ens.x_vectors.transpose())
+    return np.cov(ens.x_vectors.transpose())
 
 
 def get_covariance_matrix_predictions(
@@ -50,18 +50,19 @@ def get_covariance_matrix_predictions(
     dataset = get_prediction_dataset(ens, prediction_index)
 
     # call lowlevel routine using the prediction ensemble
-    return get_covariance_matrix_lowlevel(dataset=dataset)
+    return np.cov(dataset)
 
 
 def get_spectral_decomposition_parameters(
         ens: Ensemble,
         normalize: bool = False,
-        lower_absolute_cutoff: float = None,
-        lower_relative_cutoff: float = None,
-        inv_upper_absolute_cutoff: float = None,
-        inv_upper_relative_cutoff: float = None,
+        only_separable_directions: bool = False,
+        cutoff_absolute_separable: float = 1e-16,
+        cutoff_relative_separable: float = 1e-16,
         only_identifiable_directions: bool = False,
-        only_separable_directions: bool = False):
+        cutoff_absolute_identifiable: float = 1e-16,
+        cutoff_relative_identifiable: float = 1e-16
+) -> Tuple[np.ndarray, np.ndarray]:
     """
     Compute the spectral decomposition of ensemble parameters.
 
@@ -74,55 +75,65 @@ def get_spectral_decomposition_parameters(
         flag indicating whether the returned Eigenvalues should be normalized
         with respect to the largest Eigenvalue
 
-    lower_absolute_cutoff:
+    only_separable_directions:
+        return only separable directions according to
+        cutoff_[absolute/relative]_separable
+
+    cutoff_absolute_separable:
         Consider only eigenvalues of the covariance matrix above this cutoff
         (only applied when only_separable_directions is True)
 
-    lower_relative_cutoff:
+    cutoff_relative_separable:
         Consider only eigenvalues of the covariance matrix above this cutoff,
         when rescaled with the largest eigenvalue
         (only applied when only_separable_directions is True)
 
-    inv_upper_absolute_cutoff:
+    only_identifiable_directions:
+        return only identifiable directions according to
+        cutoff_[absolute/relative]_identifiable
+
+    cutoff_absolute_identifiable:
         Consider only low eigenvalues of the covariance matrix with inverses
         above of this cutoff
         (only applied when only_identifiable_directions is True)
 
-    inv_upper_relative_cutoff:
+    cutoff_relative_identifiable:
         Consider only low eigenvalues of the covariance matrix when rescaled
         with the largest eigenvalue with inverses above of this cutoff
         (only applied when only_identifiable_directions is True)
 
-    only_identifiable_directions:
-        return only identifiable directions according to inv_upper_cutoff
-
-    only_separable_directions:
-        return only separable directions according to lower_cutoff
-
     Returns
     =======
-    eigen_values:
+    eigenvalues:
         Eigenvalues of the covariance matrix
 
-    eigen_vectors:
-        eigenvectors of the covariance matrix
+    eigenvectors:
+        Eigenvectors of the covariance matrix
     """
+    # check inputs
+    assert sum([only_identifiable_directions, only_separable_directions]) < 2
+
     covariance = get_covariance_matrix_parameters(ens)
     return get_spectral_decomposition_lowlevel(
-        covariance, normalize, lower_absolute_cutoff, lower_relative_cutoff,
-        inv_upper_absolute_cutoff, inv_upper_relative_cutoff,
-        only_identifiable_directions, only_separable_directions)
+        matrix=covariance, normalize=normalize,
+        only_separable_directions=only_separable_directions,
+        cutoff_absolute_separable=cutoff_absolute_separable,
+        cutoff_relative_separable=cutoff_relative_separable,
+        only_identifiable_directions=only_identifiable_directions,
+        cutoff_absolute_identifiable=cutoff_absolute_identifiable,
+        cutoff_relative_identifiable=cutoff_relative_identifiable)
 
 
 def get_spectral_decomposition_predictions(
         ens: Ensemble,
         normalize: bool = False,
-        lower_absolute_cutoff: float = None,
-        lower_relative_cutoff: float = None,
-        inv_upper_absolute_cutoff: float = None,
-        inv_upper_relative_cutoff: float = None,
+        only_separable_directions: bool = False,
+        cutoff_absolute_separable: float = 1e-16,
+        cutoff_relative_separable: float = 1e-16,
         only_identifiable_directions: bool = False,
-        only_separable_directions: bool = False):
+        cutoff_absolute_identifiable: float = 1e-16,
+        cutoff_relative_identifiable: float = 1e-16
+) -> Tuple[np.ndarray, np.ndarray]:
     """
     Compute the spectral decomposition of ensemble predictions.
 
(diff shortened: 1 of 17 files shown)


paulstapor:
Shall I just also add an engine and create a task to be executed for predictions? I just found out that multi-processing within the ICB objective doesn't work without an engine...

yannikschaelte:
What is "ICB objective"? If you want to multi-process the execution, a first thing would be to check whether you can formalize this as an "EnsembleTask" derived from https://github.com/ICB-DCM/pyPESTO/blob/master/pypesto/engine/task.py, similar to for optimization and profiling. If that does not work (e.g. because you need more dynamic scheduling or interprocess communication), one would need to write the full Engine anew.

paulstapor:
> What is "ICB objective"? If you want to multi-process the execution, a first thing would be to check whether you can formalize this as an "EnsembleTask" derived from https://github.com/ICB-DCM/pyPESTO/blob/master/pypesto/engine/task.py, similar to for optimization and profiling. If that does not work (e.g. because you need more dynamic scheduling or interprocess communication), one would need to write the full Engine anew.

amici objecive... Too much distraction due to icb (com health) orga retreat... :joy:

yannikschaelte:
> > What is "ICB objective"? If you want to multi-process the execution, a first thing would be to check whether you can formalize this as an "EnsembleTask" derived from https://github.com/ICB-DCM/pyPESTO/blob/master/pypesto/engine/task.py, similar to for optimization and profiling. If that does not work (e.g. because you need more dynamic scheduling or interprocess communication), one would need to write the full Engine anew.
> 
> amici objecive... Too much distraction due to icb (com health) orga retreat... 
> I looked at the optimization task... I think that should be redoable for predictions or ensembles

Nice! ICB objective is a classical example of failed multitasking :laughing:

paulstapor:
So, what's the consense? Adding an engine to predictions + ensembles? Or leave that t @dilpath in #527 `:D

yannikschaelte:
No objections to anything from my side ;)

dilpath:
> So, what's the consense? Adding an engine to predictions + ensembles? Or leave that t @dilpath in #527 `:D

:+1: Can work on this now

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+        ens: Ensemble,
+        normalize: bool = False,
+        lower_absolute_cutoff: float = None,
+        lower_relative_cutoff: float = None,
+        inv_upper_absolute_cutoff: float = None,
+        inv_upper_relative_cutoff: float = None,
+        only_identifiable_directions: bool = False,
+        only_separable_directions: bool = False):
+    """
+    Compute the spectral docmposition of ensemble parameters.
```
dilpath, line 66:
```suggestion
    Compute the spectral decomposition of ensemble parameters.
```

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+    only_separable_directions:
+        return only separable directions according to lower_cutoff
+
+    Returns
+    =======
+    eigen_values:
+        Eigenvalues of the covariance matrix
+
+    eigen_vectors:
+        eigenvectors of the covariance matrix
```
dilpath, line 108:
```suggestion
    A tuple with the following values in the following order.

    eigenvalues:
        Eigenvalues of the covariance matrix.

    eigenvectors:
        Eigenvectors of the covariance matrix.
```

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+        ens: Ensemble,
+        normalize: bool = False,
+        lower_absolute_cutoff: float = None,
+        lower_relative_cutoff: float = None,
+        inv_upper_absolute_cutoff: float = None,
+        inv_upper_relative_cutoff: float = None,
+        only_identifiable_directions: bool = False,
+        only_separable_directions: bool = False):
+    """
+    Compute the spectral docmposition of ensemble predictions.
```
dilpath, line 127:
```suggestion
    Compute the spectral decomposition of ensemble predictions.
```

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
+import numpy as np
+from typing import Union
+
+from ..ensemble import Ensemble, EnsemblePrediction
+from .utils import get_prediction_dataset
+
+
+def get_covariance_matrix_parameters(ens: Ensemble):
```
dilpath, line 8: Typehints for returned values of this and other methods here are missing

paulstapor:
> > So, what's the consense? Adding an engine to predictions + ensembles? Or leave that t @dilpath in #527 `:D
> 
> +1 Can work on this now


No super big rush there... Just found out that mutlithreading within AmiciObjective also works well for the AmiciPredictor...
I'm more interested in getting this here into develop :)

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+
+def get_spectral_decomposition_parameters(
+        ens: Ensemble,
+        normalize: bool = False,
+        lower_absolute_cutoff: float = None,
+        lower_relative_cutoff: float = None,
+        inv_upper_absolute_cutoff: float = None,
+        inv_upper_relative_cutoff: float = None,
+        only_identifiable_directions: bool = False,
+        only_separable_directions: bool = False):
```
dilpath, line 64: Should this be considered a private method? If not, maybe the arguments could be simplified to e.g.
```python
cutoffs_separable = (0.1, None)
entry_point(cutoffs_separable=cutoffs_separable)
```
Then parse e.g.
```python
def entry_point(cutoffs_identifiable = (None, None), cutoffs_separable = (None, None):
    only_identifiable_directions = np.array(cutoffs_identifiable).any()
    only_separable_directions = np.array(cutoffs_separable).any()
    assert sum([only_identifiable_directions, only_separable_directions]) < 2  # or == 1?
    lower_absolute_cutoff, upper_absolute_cutoff = cutoffs_separable
    inv_lower_absolute_cutoff, inv_upper_absolute_cutoff = cutoffs_identifiable
```

Might make it easier for users then.

pypesto/ensemble/covariance_analysis.py
dilpath: Should change `eigen_value` -> `eigenvalue` and `eigen_vector` -> `eigenvector` everywhere

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+        eigenvectors of the covariance matrix
+    """
+    covariance = get_covariance_matrix_predictions(ens)
+    return get_spectral_decomposition_lowlevel(
+        covariance, normalize, lower_absolute_cutoff, lower_relative_cutoff,
+        inv_upper_absolute_cutoff, inv_upper_relative_cutoff,
+        only_identifiable_directions, only_separable_directions)
+
+
+def get_covariance_matrix_lowlevel(dataset):
```
dilpath, line 179: Is it expected that this function may get more complicated? If not, could remove it?

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+
+    # get the eigenvalue decomposition
+    eigen_vals, eigen_vectors = np.linalg.eigh(matrix_to_analyze)
+
+    # get a normalized version
+    rel_eigen_vals = eigen_vals / np.max(eigen_vals)
+
+    # If no filtering is wanted, we can return
+    if not only_identifiable_directions and not only_separable_directions:
+        # apply normlization
```
dilpath, line 260:
```suggestion
        # apply normalization
```

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+        if normalize:
+            eigen_vals = rel_eigen_vals
+        return eigen_vals, eigen_vectors
+
+    if only_identifiable_directions and only_identifiable_directions:
```
dilpath, line 265: Should one be `only_separable_directions`?

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+        matrix_to_analyze: np.ndarray,
+        normalize: bool = False,
+        lower_absolute_cutoff: float = None,
+        lower_relative_cutoff: float = None,
+        inv_upper_absolute_cutoff: float = None,
+        inv_upper_relative_cutoff: float = None,
+        only_identifiable_directions: bool = False,
+        only_separable_directions: bool = False):
+    """
+    Compute the spectral docmposition of ensemble parameters or predictions.
```
dilpath, line 206:
```suggestion
    Compute the spectral decomposition of ensemble parameters or predictions.
```

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+
+    Parameters
+    ==========
+    matrix_to_analyze:
```
dilpath, line 210:
```suggestion
    matrix:
```

Also elsewhere. There's only one matrix argument so I think it's already clear with just `matrix`

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+
+    eigen_vectors:
+        eigenvectors of the covariance matrix
+    """
+
+    # get the eigenvalue decomposition
+    eigen_vals, eigen_vectors = np.linalg.eigh(matrix_to_analyze)
+
+    # get a normalized version
+    rel_eigen_vals = eigen_vals / np.max(eigen_vals)
```
dilpath, line 256:
```suggestion
    rel_eigenvalues = eigenvalues / np.max(eigenvalues)
```

Also rename `rel_eigen_vals` everywhere to `rel_eigenvalues`

pypesto/ensemble/covariance_analysis.py
dilpath: Could this could be checked much earlier?

pypesto/ensemble/covariance_analysis.py
dilpath: And set default cutoffs to `-np.inf` and `np.inf` I guess (not sure, maybe this is already done?)

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+                lower_relative_cutoff is not None:
+            above_cutoff = np.array([
+                eigen_vals[i_eig] > lower_absolute_cutoff and
+                rel_eigen_vals[i_eig] > lower_relative_cutoff
+                for i_eig in range(len(eigen_vals))
+            ])
+        elif lower_absolute_cutoff is not None:
+            above_cutoff = eigen_vals > lower_absolute_cutoff
+        elif lower_relative_cutoff is not None:
+            above_cutoff = rel_eigen_vals > lower_relative_cutoff
```
dilpath, line 282: Could both relative and absolute cutoffs be useful together? Looks like this `if-elif` will only allow for one to be used?

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+        else:
+            raise Exception('Need a lower cutoff (aboslute or relative, '
```
dilpath, line 284:
```suggestion
            raise Exception('Need a lower cutoff (absolute or relative, '
```

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+        if normalize:
+            eigen_vals = rel_eigen_vals
+        return eigen_vals, eigen_vectors
+
+    # Identifiable directions are wanted: an filtering of the inverse
+    # eigenvalues is needed (upper pass of inverse = lower pass of original)
+    if inv_upper_absolute_cutoff is not None and \
+            inv_upper_relative_cutoff is not None:
+        below_cutoff = np.array([
+            eigen_vals[i_eig] < 1 / eigen_vals > inv_upper_absolute_cutoff and
```
dilpath, line 300: Should `1 / eigen_vals > inv_upper_absolute_cutoff` be checked each time?

pypesto/ensemble/covariance_analysis.py
```diff
@@ -0,0 +1,318 @@
...
+            1 / rel_eigen_vals[i_eig] > inv_upper_relative_cutoff
+            for i_eig in range(len(eigen_vals))
+        ])
```
dilpath, line 303:
```suggestion
        below_cutoff = np.array([
            eig_abs_i < 1 / eigen_vals > inv_upper_absolute_cutoff and
            1 / eig_rel_i > inv_upper_relative_cutoff
            for eig_abs_i, eig_rel_i in zip(eigen_vals, rel_eigen_vals)
        ])
```

dilpath (commented):
Will review the rest in a little bit, sorry for delays

pypesto/visualize/dimension_reduction.py
```diff
@@ -0,0 +1,306 @@
...
+    -------
+    ax: matplotlib.Axes
+        The plot axes.
+    """
+
+    # first get the data to check identifiability
+    # axes
+    if ax is None:
+        ax = plt.subplots()[1]
+        fig = plt.gcf()
```
dilpath, line 258:
```suggestion
        fig, ax = plt.subplots()
```

dilpath (approved):
:+1: 

There are no tests that assert that the new methods work as intended, only tests that ensure that errors are not produced during a standard workflow.

pypesto/ensemble/covariance_analysis.py
paulstapor: Maybe in some way... On might want to choose whether we want to use the algorithms from scipy or numpy... but not in the next future. You're right, will remove it...

pypesto/ensemble/covariance_analysis.py
paulstapor: I like your idea of parsing the inputs in the beginning. The default cutoffs are implicitly implemented already. Default values at the level of 1e-16 would make sense, as this is machine precision on most computers.

pypesto/ensemble/covariance_analysis.py
paulstapor: It's designed that both will be used, if both are given (the if clause), and that only one is used, if only one is given (the elif-clauses)

pypesto/ensemble/covariance_analysis.py
paulstapor: Is now checked much earlier as you suggested... and corrected...

pypesto/ensemble/covariance_analysis.py
paulstapor: thanks, was an error!
Need to add a test...

pypesto/ensemble/covariance_analysis.py
paulstapor: Okay, added a check and most hopefully meaningful default parameters. One could collect thresholds in a tuple, but I'm against that, as I prefer more function parameters with clear names over slimming down function api... Although that's a style-decision, I have to admit

pypesto/ensemble/utils.py
```diff
@@ -0,0 +1,187 @@
...
+    for i_result, result in enumerate(ensemble_prediction.prediction_results):
+        tmp_base_path = base + f'{PREDICTION_RESULTS}_{i_result}'
+        result.write_to_h5(output_file, base_path=tmp_base_path)
+
+    # write lower bounds per condition, if available
+    if ensemble_prediction.lower_bound is not None:
+        f.create_group(base + f'{LOWER_BOUND}s')
+        for i_cond, lower_bounds in enumerate(ensemble_prediction.lower_bound):
+            condition_id = \
+                ensemble_prediction.prediction_results[0].condition_ids[i_cond]
```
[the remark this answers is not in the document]
paulstapor: In principle yes, but we store a multitude of PredictionResults here. As PredictionResult also must work as a class of its own, it makes sense to also store the conditionIds in there. One could now either reduce what's stored, therefore introduce a new class... or storing something redundant... What do you think?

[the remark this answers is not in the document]
dilpath: For the purposes of e.g. plotting by condition or observable, it would be nice to automatically ensure that all `PredictionResults` share the same condition IDs, and the prediction for the same condition between different `PredictionResults` contains the same observable IDs. And also to provide methods to easily get all data by condition or observable, from an `EnsemblePrediction`. Maybe a wrapper for `SampleEnsemblePrediction`? Anyway, I can do this, let me know what you think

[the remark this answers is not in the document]
paulstapor: I would actually suggest to open 3 issues:
 - wishes and stuff we can use the ensemble functionality for (e.g., what you just wrote)
 - wishes and stuff we can use the prediction functionality for (e.g., prediction profiles)
 - wishes and stuff we can use the dimesnion reduction functionality for (e.g., analyzing optimization runs from a multi-start visually)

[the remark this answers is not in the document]
paulstapor: But yes: Asserting these things are equal would be good. Currently, they must be equal, just by the way the predict functions creates things...

PR #568, opened 2021-02-19
Minor fixups ensembles

paulstapor:
Making the predictor more flexible and fixing a bug in the parameter mapping...


pypesto/petab/importer.py
```diff
@@ -285,6 +286,9 @@ def create_prediction(self,
         ----------
         objective:
             An objective object, which will be used to get model simulations
+        amici_output_fields:
+            keys which exist in the return data object from Amici, which should
+            be available for the post_processor routines
```
dilpath, line 291:
```suggestion
            keys that exist in the return data object from AMICI, which should
            be available for the post-processors
```

pypesto/prediction/amici_predictor.py
```diff
@@ -39,6 +41,9 @@ def __init__(self,
         ----------
         amici_objective:
             An objective object, which will be used to get model simulations
+        amici_output_fields:
+            keys which exist in the return data object from Amici, which should
+            be available for the post_processor routines
```
dilpath, line 46:
```suggestion
            keys that exist in the return data object from AMICI, which should
            be available for the post-processors
```

pypesto/prediction/amici_predictor.py
```diff
@@ -274,19 +292,19 @@ def _default_output(amici_outputs):
...
-                                  AMICI_T: rdata[AMICI_T],
-                                  AMICI_X: rdata[AMICI_X],
-                                  AMICI_SX: rdata[AMICI_SX],
-                                  AMICI_Y: rdata[AMICI_Y],
-                                  AMICI_SY: rdata[AMICI_SY]})
+            amici_outputs.append({
+                output_field: deepcopy(rdata[output_field])
+                for output_field in self.amici_output_fields
+            })
+        del chunk
```
dilpath, line 310: Is `deepcopy` used because `chunk` remained in memory (or I guess, because after `del chunk`, maybe the value in `amici_outputs` is also deleted without `deepcopy`)? If so, is both `deepcopy` and `del` necessary? Since `chunk` should get garbage collected I guess, if there are no references to it? Just curious, fine to leave as-is

dilpath (approved):
:+1:

pypesto/prediction/amici_predictor.py
paulstapor: I actually intended t go without the deepcopy and the del, and it seemed to work fine. However, in my debugger I experienced that chunk was still present i nthe memory at a later call of the function and just got overwritten... That puzzled me extremely...
Might be, that this is really only due to the debugger. However, I wanted make really, really safe that chunk is dead, dead, DEAD! :bat: No, not undead, dead. :coffin:  ... That's why deepcopy and del... :man_shrugging:

PR #572, opened 2021-02-24
fix aggregated objective deepcopy

FFroehlich:


yannikschaelte (approved):
might also work without manual definition I guess

FFroehlich:
> might also work without manual definition I guess



> might also work without manual definition I guess

Nope:

```
    def __deepcopy__(self, memodict=None) -> 'ObjectiveBase':
>       other = type(self)()  # maintain type for derived classes
E       TypeError: __init__() missing 1 required positional argument: 'objectives'

pypesto/objective/base.py:52: TypeError
```

PR #578, opened 2021-03-05
Update for fides 0.3.0 and make hybrid happ new default

FFroehlich:
Fides has a new hybrid mode where we can first use the FIM and then later on use BFGS (which is updated from the start) for superior convergence.


yannikschaelte (approved):
Sounds useful! Unrelated: It may be good to check objective.has_hess if hybrid or None is used, as otherwise the error message when trying to obtain a Hessian may be a bit cryptic.

pypesto/optimize/optimizer.py
```diff
@@ -936,8 +936,16 @@ def minimize(
         args = {'mode': MODE_FUN}
         if self.hessian_update is None or isinstance(self.hessian_update,
                                                      fides.HybridUpdate):
+            if not problem.objective.has_hess:
+                raise ValueError('Specified hessian update scheme cannot be '
+                                 'used with objectives that do not support '
+                                 'Hessian computation.')
             args['sensi_orders'] = (0, 1, 2)
         else:
+            if not problem.objective.has_grad:
```
yannikschaelte, line 945: the gradient question should not require an "else", right?

FFroehlich: correct

PR #585, opened 2021-03-10
Added ProfileStorage test

PaulJonasJost:
- [x] ProfileStorage test
- [x] SamplingStorage test


test/base/test_store.py
```diff
@@ -141,3 +145,83 @@ def test_storage_trace():
                                 getattr(mem_res['history'],
                                         f'get_{entry}_trace')()[iteration],
                                 hdf5_entry_trace[iteration])
+
+
+def test_storage_profiling():
```
yannikschaelte, line 150: for (new/all) tests please a one-line """This test tests this and that stuff""" docstring at the beginning. (or a longer description too if necessary)

test/base/test_store.py
```diff
@@ -141,3 +145,83 @@ def test_storage_trace():
...
+    with tempfile.TemporaryDirectory(dir=".") as tmpdirname:
+        _, fn = tempfile.mkstemp(".hdf5", dir=f"{tmpdirname}")
+
+    pypesto_profile_writer = ProfileResultHDF5Writer(fn)
+    pypesto_profile_writer.write(profile_original)
+    pypesto_profile_reader = ProfileResultHDF5Reader(fn)
+    profile_read = pypesto_profile_reader.read()
+
+    # compare the x_paths of both profiles
+    np.testing.assert_array_equal(
```
yannikschaelte, line 181: I think it would actually be good to test whether all entries in the two objects are the same (up to floating point precision) (should be done like that in the optimization result tests).

test/base/test_store.py
```diff
@@ -141,3 +145,83 @@ def test_storage_trace():
...
+    with tempfile.TemporaryDirectory(dir=".") as tmpdirname:
+        _, fn = tempfile.mkstemp(".hdf5", dir=f"{tmpdirname}")
+
+    pypesto_sample_writer = SamplingResultHDF5Writer(fn)
+    pypesto_sample_writer.write(sample_original)
+    pypesto_sample_reader = SamplingResultHDF5Reader(fn)
+    sample_read = pypesto_sample_reader.read()
+
+    # compare the x_paths of both profiles
+    np.testing.assert_array_equal(
```
yannikschaelte, line 224: same as above: check all content

yannikschaelte (changes requested):

test/base/test_store.py
yannikschaelte: this would allow noticing when e.g. some variables are not saved for whatever reason

test/base/test_store.py
```diff
@@ -141,3 +145,83 @@ def test_storage_trace():
...
+    optimizer = pypesto.optimize.ScipyOptimizer()
+
+    result_optimization = pypesto.optimize.minimize(
+        problem=problem, optimizer=optimizer,
+        n_starts=n_starts)
+    profile_original = profile.parameter_profile(
+        problem=problem, result=result_optimization,
+        profile_index=[0], optimizer=optimizer)
+    with tempfile.TemporaryDirectory(dir=".") as tmpdirname:
+        _, fn = tempfile.mkstemp(".hdf5", dir=f"{tmpdirname}")
```
yannikschaelte, line 173: I think this "with" statement does not make sense here -- what should happen is that a temporary folder is created, then files inserted, or just a single temporary file is created, then the file is filled, and in the end the file is deleted again (can be done either using a with context, or a try--finally.

test/base/test_store.py
PaulJonasJost: well all I think is not a good thing since time might vary. Will implement a check for all np.ndarrays, floats and ints aside from time

PaulJonasJost:
I think I have integrated all asked changes :)

test/base/test_store.py
```diff
@@ -141,3 +145,125 @@ def test_storage_trace():
                                 getattr(mem_res['history'],
                                         f'get_{entry}_trace')()[iteration],
                                 hdf5_entry_trace[iteration])
+
+
+def test_storage_profiling():
+    """
+    This test tests the saving and loading of profiles
+    into HDF5 through pypesto.store.ProfileResultHDF5Writer
+    and pypesto.store.ProfileResultHDF5Reader.Tests all entries
```
yannikschaelte, line 154:
```suggestion
    and pypesto.store.ProfileResultHDF5Reader. Tests all entries
```

test/base/test_store.py
```diff
@@ -141,3 +145,125 @@ def test_storage_trace():
...
+            elif isinstance(
+                    profile_original.profile_result.list[0][0][key],
+                    int):
+                assert profile_original.profile_result.list[0][0][key] == \
+                       profile_read.profile_result.list[0][0][key]
+    finally:
+        if os.path.exists(fn):
+            os.remove(fn)
+        else:
+            print("The file does not exist")
```
yannikschaelte, line 206: I think this can be omitted.

test/base/test_store.py
```diff
@@ -141,3 +145,125 @@ def test_storage_trace():
...
+            elif isinstance(sample_original.sample_result[key], (float, int)):
+                np.testing.assert_almost_equal(
+                    sample_original.sample_result[key],
+                    sample_read.sample_result[key],
+                )
+    finally:
+        if os.path.exists(fn):
+            os.remove(fn)
+        else:
+            print("The file does not exist")
```
yannikschaelte, line 269: --

yannikschaelte (approved):

PR #592, opened 2021-03-12
Minor fixups ensembles

paulstapor:
Fixing parameter mapping in predictor, and saving conditionIds automatically


pypesto/prediction/amici_predictor.py
```diff
@@ -44,6 +44,7 @@ def __init__(self,
         amici_output_fields:
             keys that exist in the return data object from AMICI, which should
             be available for the post-processors
+
```
dilpath, line 47:
```suggestion
```

dilpath (changes requested):

PR #593, opened 2021-03-13
udate min fval

FFroehlich:
This corrects the computation of min_val in the waterfall plot. This is a breaking change, since now log-plots yield use the standard `offset_y=1-minval`  by default instead of an `offst_y=0` by default. However, I believe this makes sense, since the previous implementation made it impossible to use the `offset_y=1-minval`, since this would always erroneously trigger https://github.com/ICB-DCM/pyPESTO/blob/cb8bffc825235a8d94208b4737b513bc5980b58a/pypesto/visualize/misc.py#L115. This implementation corrects this and the previous can still be restored by specifying `offset_y=0`.


yannikschaelte (approved):
sounds fine for me. @paulstapor @DantongWang any concerns?

PR #601, opened 2021-03-17
Fix wrong default bounds for parameters plot

### LINKED ISSUE #595: wrong bounds are chosen in parameter plot when not all parameters are plotted
the desired parameter indices are related to the full x vector. 
in the handle_inputs() function the vectors ub and lb contain only the bounds of the free parameters. 
That leads to wrong bounds in the plot, I guess.

yannikschaelte:
fixes #595


FFroehlich (approved):
👍

PR #606, opened 2021-03-23
Implement Emcee sampler

yannikschaelte:


dilpath (approved):
Some missing docstrings but i guess those functions correspond to functions/docstrings in other samplers. Could be copied here anyway.

Is a list of samplers documented anywhere or should a user look into the code?

yannikschaelte:
> Some missing docstrings but i guess those functions correspond to functions/docstrings in other samplers. Could be copied here anyway.

The undocumented functions just override base class functions (where they are documented). Did not want to copy-paste the documentation from there in here, might make things harder to maintain, but certainly a point.

> Is a list of samplers documented anywhere or should a user look into the code?

Good point. I will add an example to the sampler study notebook, but maintaining a list of samplers(, optimizers) somewhere certainly makes sense.

dilpath:
> > Some missing docstrings but i guess those functions correspond to functions/docstrings in other samplers. Could be copied here anyway.
> 
> The undocumented functions just override base class functions (where they are documented). Did not want to copy-paste the documentation from there in here, might make things harder to maintain, but certainly a point.

Not sure on convention but then maybe something like `"""See base class."""` would be good?

yannikschaelte:
> > > Some missing docstrings but i guess those functions correspond to functions/docstrings in other samplers. Could be copied here anyway.
> > 
> > 
> > The undocumented functions just override base class functions (where they are documented). Did not want to copy-paste the documentation from there in here, might make things harder to maintain, but certainly a point.
> 
> Not sure on convention but then maybe something like `"""See base class."""` would be good?

done :+1:

dweindl:
> Not sure on convention but then maybe something like `"""See base class."""` would be good?

I think it might be preferable to not add any dummy doctrings and instead let sphinx figure it out (https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_inherit_docstrings)

yannikschaelte:
> > Not sure on convention but then maybe something like `"""See base class."""` would be good?
> 
> I think it might be preferable to not add any dummy doctrings and instead let sphinx figure it out (https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_inherit_docstrings)

good point, we should add that soon

yannikschaelte:
@dilpath @jvanhoefer in light of #607 I added a bounds check here as well (like for the other samplers).

PR #611, opened 2021-03-29
fix #569

### LINKED ISSUE #569: Waterfall plot with multiple results doesn't handle unsorted optimizer results well
[Chen_MSB2009_all_starts_adjoint.pdf](https://github.com/ICB-DCM/pyPESTO/files/6025931/Chen_MSB2009_all_starts_adjoint.pdf)

PaulJonasJost:
Added a sorting of the fvals, problem before was that the points were plotted sorted but the lines were plotted as passed through.


FFroehlich (approved):

FFroehlich:
👍

Issue #567, opened 2021-02-18
HDF5 History saving for MultiProcessEngine


PaulJonasJost:
Still needs to be done, currently only for SingleCoreEngine.

PaulJonasJost:
closed with #620

FFroehlich:
> closed with #620

that PR was closed and not merged though

yannikschaelte:
#639 to be exact

PR #620, opened 2021-04-08
Hdf5History with multiprocessing

PaulJonasJost:
Workflow is like that:
* if engine is not SingleCoreEngine and we want a Hdf5History, we change the filename of the history options to include an `_{id}`
* This creates for each start one hdf5 file. They are saved in a directory that has the name of the original hdf5file in history options
* create an extra hdf5 file with original name that has all the results (to be used later for storage of optimiser and so on)

Do you think that is ok?

Edit: solves #567


dilpath (commented):
Are the partial HDF5 files deleted after use? Maybe make it optional but delete by default? If you  don't go with the `tempfile` suggestion, then it's maybe safest to not delete by default. If using a `TemporaryDirectory`, can just delete the whole directory.

PaulJonasJost:
> 
> 
> Are the partial HDF5 files deleted after use? Maybe make it optional but delete by default? If you don't go with the `tempfile` suggestion, then it's maybe safest to not delete by default. If using a `TemporaryDirectory`, can just delete the whole directory.

No they are not deleted, as they are still needed. There is basically no data in the original file, but links at the rght places pointing to the files in the folder.

pypesto/optimize/util.py
```diff
@@ -0,0 +1,41 @@
...
+        fn, type = os.path.splitext(history_options.storage_file)
+        # creates a folder in which the hdf5 files will be saved
+        if not os.path.exists(fn):
+            os.mkdir(fn)
+        fn = fn + '/' + fn + '_{id}'
+        history_options.storage_file = fn + type
+        # create hdf5 file that gather the other with history group
+        f = h5py.File(filename, mode='a')
+        get_or_create_group(f, "history")
+        return filename
```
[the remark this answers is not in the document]
PaulJonasJost: > filename has no extension
The function is only called if it has an .hdf5/.h5 extension, thus this might not be a real concern?
> filename has parents (resolved by the pathlib suggestion below)
Yes you are right that is an issue, what i want there would then be if `filename = 'a/b/c.txt'` i want `fn = a/b/c/c_id.txt`. Will correct that, thanks.

pypesto/optimize/util.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+import h5py
+
+
+def check_hdf5_mp(history_options: HistoryOptions,
+                  engine: Engine):
+    """
+    This function checks whether we use another engine than
+    `SingleCoreEngine` and whether we use a `Hdf5History`. If
+    that is the case, i rephrases the filename to create a hdf5 file
```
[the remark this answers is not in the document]
PaulJonasJost: so delete the whole sentence and only keep 
```
This function checks whether we use another engine than
    `SingleCoreEngine` and whether we use a `Hdf5History`.  Returns filename for later usage in filling hdf5 file
    otherwise return None.
```

pypesto/optimize/util.py
```diff
@@ -11,31 +13,59 @@ def check_hdf5_mp(history_options: HistoryOptions,
...
-        if not os.path.exists(fn):
-            os.mkdir(fn)
-        fn = fn + '/' + fn + '_{id}'
+        file_path = Path(filename)
+        fn = file_path.parent / (file_path.stem)
+        # create directory with same name as original file
+        fn.mkdir(parents=True, exist_ok=True)
+        fn = fn / (file_path.stem + '_{id}' + file_path.suffix)
+        fn = str(fn)
         history_options.storage_file = fn + type
```
dilpath:
```suggestion
        # create directory with same name as original file stem
        partial_file_path = (
            file_path.parent / file_path.stem /
            (file_path.stem + '_{id}' + file_path.suffix)
        )
        partial_file_path.mkdir(parents=True, exist_ok=True)
        history_options.storage_file = str(partial_file_path)
```

No need to do `mkdir` before including the filename in the path.

Also I guess the `+ type` is a typo

PaulJonasJost: oh yes that is a typo. But I tested it and when I had a path like `a/b/c.txt` it created a folder called `c.txt` that's why that change was included

dilpath: Ah true. Fixed in the suggestion (can use `Path.parent.mkdir` instead), if you prefer this.

PaulJonasJost: Used the suggestion 👍

pypesto/optimize/util.py
```diff
@@ -0,0 +1,68 @@
...
+    If a parallelization engine is not used, `None` is returned.
+    """
+    if not isinstance(engine, SingleCoreEngine):
+        filename = history_options.storage_file
+        file_path = Path(filename)
+        # create directory with same name as original file stem
+        partial_file_path = (
+                file_path.parent / file_path.stem /
+                (file_path.stem + '_{id}' + file_path.suffix)
+        )
```
dilpath, line 35:
```suggestion
        partial_file_path = (
            file_path.parent / file_path.stem /
            (file_path.stem + '_{id}' + file_path.suffix)
        )
```

pypesto/optimize/util.py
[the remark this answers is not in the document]
jvanhoefer: Also, I would not use first person for docs. Rest of the docs doesn't do that either and better stay consistent here...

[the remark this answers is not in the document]
PaulJonasJost: at least in this case it was fixed 👍

pypesto/optimize/util.py
```diff
@@ -0,0 +1,68 @@
...
+        The Engine which is used in the optimization
+    history_options:
+        The HistoryOptions used in the optimization
+
+    Returns
+    -------
+    The filename that will be used to combine the partial HDF5 files later.
+    If a parallelization engine is not used, `None` is returned.
+    """
+    if not isinstance(engine, SingleCoreEngine):
```
jvanhoefer, line 28: Why don't you go for 
```python
if isinstance(engine, SingleCoreEngine):
   return None
filename = ...
```
Would be more intuitive to read for future contributors... :)

pypesto/optimize/util.py
```diff
@@ -0,0 +1,68 @@
...
+        filename = history_options.storage_file
+        file_path = Path(filename)
+        # create directory with same name as original file stem
+        partial_file_path = (
+            file_path.parent / file_path.stem /
+            (file_path.stem + '_{id}' + file_path.suffix)
+        )
```
jvanhoefer, line 35: would go for
```suggestion
        partial_file_path = os.path.join(
            file_path.parent, <usw ...>
        )
```

jvanhoefer: Also then, you do not need the `str(partial_file_path)`.

pypesto/optimize/util.py
```diff
@@ -0,0 +1,68 @@
...
+        partial_file_path.parent.mkdir(parents=True, exist_ok=True)
+        history_options.storage_file = str(partial_file_path)
+        # create hdf5 file that gathers the others within history group
```
jvanhoefer, line 38:
```suggestion

        # create hdf5 file that gathers the others within history group
```

jvanhoefer: empty line before comment structures code

jvanhoefer: (same elsewhere...)

pypesto/optimize/util.py
```diff
@@ -0,0 +1,68 @@
...
+        with h5py.File(filename, mode='a') as f:
+            get_or_create_group(f, "history")
+        return filename
+    return None
+
+
+def fill_hdf5_file(ret,
+                   filename):
```
jvanhoefer, line 46: type hints for inputs missing

jvanhoefer (changes requested):
As far, as I can see, the functionalities are currently not covered by the unit tests. We should have running tests before merging this.

pypesto/optimize/util.py
```diff
@@ -0,0 +1,68 @@
...
+    ---------
+    engine:
+        The Engine which is used in the optimization
+    history_options:
+        The HistoryOptions used in the optimization
+
+    Returns
+    -------
+    The filename that will be used to combine the partial HDF5 files later.
+    If a parallelization engine is not used, `None` is returned.
```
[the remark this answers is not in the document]
dweindl: It's true that this is done elsewhere in pypesto. But how could that be considered helpful? The return values are unnamed, and as a user I would/should not care how the returned variable is named within the called function.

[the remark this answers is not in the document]
PaulJonasJost: agree with Daniel and if one would be very strict it is wrong as when I return `None` I do not return `filename` with the value `None` but just `None`.

pypesto/optimize/util.py
PaulJonasJost: if I do that, I'd loose the possibility to conveniently create the directory where the files would be stored (line below).

pypesto/optimize/util.py
PaulJonasJost: anything I could put into ret? I think it depends on the engine what is returned here.

PaulJonasJost:
ahh something went terribly wrong I think. I will delete this branch open up a new one and create a new PR...

PR #627, opened 2021-04-12
Add pyswarms optimizer

PhilippStaedter:
pull request to merge the pyswarms optimzer into develop


pypesto/optimize/optimizer.py
```diff
@@ -720,6 +725,84 @@ def is_least_squares(self):
...
+        c1: cognitive parameter
+        c2: social parameter
+        w: inertia parameter
+    """
+
+    def __init__(self, par_popsize: float = 10, options: Dict = None):
+        super().__init__()
+
+        if options is None:
+            options = {'maxiter': 1000, 'c1': 0.5, 'c2': 0.3, 'w': 0.9}
```
yannikschaelte, line 745: Is it necessary to specify c1, c2, w? If not, I would prefer to has pyswarms choose its own defaults. maxiter does not seem to be a valid parameter.

pypesto/optimize/optimizer.py
```diff
@@ -720,6 +725,84 @@ def is_least_squares(self):
...
+            result = np.zeros(n_particles)
+            for iPar, par in enumerate(swarm):
+                result[iPar] = problem.objective.get_fval(par)
+
+            return result
+
+
+        cost, pos = optimizer.optimize(
+            successively_working_fval, iters=self.options['maxiter'])
+
```
yannikschaelte, line 793: flake8 will complain about these 2 blank lines -- only one allowed inside functions

pypesto/optimize/optimizer.py
yannikschaelte: better specify iters as a separate parameter?

pypesto/optimize/optimizer.py
```diff
@@ -720,6 +725,84 @@ def is_least_squares(self):
...
+            Parameters:
+            -----------
+            swarm: np.ndarray, shape (n_particales_in_swarm, n_parameters)
+
+            Returns:
+            --------
+            result: np.ndarray, shape (n_particles_in_swarm)
+            """
+
+            n_particles = swarm.shape[0]
```
yannikschaelte, line 783: this inner function looks good to me.

pypesto/optimize/optimizer.py
```diff
@@ -720,6 +725,84 @@ def is_least_squares(self):
...
+        if pyswarms is None:
+            raise ImportError(
+                "This optimizer requires an installation of pyswarms.")
+
+        optimizer = pyswarms.single.global_best.GlobalBestPSO(
+            n_particles=self.par_popsize, dimensions=len(x0), options=self.options,
+            bounds=(lb, ub))
+
+
+        def successively_working_fval(swarm):
```
yannikschaelte, line 771:
```suggestion
        def successively_working_fval(swarm: np.ndarray) -> np.ndarray:
```

pypesto/optimize/optimizer.py
PhilippStaedter: 1. The above written parameter values for c1,c2,w are the defaults, taken directly from pyswarms api. I would leave them there so one has the option to change them

2. I assign the value of 'maxiter' to iters by ```iters=self.options['maxiter']```. Do you want me to rename 'maxiters' into 'iters_value'?

pypesto/optimize/optimizer.py
PhilippStaedter: Done

pypesto/optimize/optimizer.py
PhilippStaedter: Done

pypesto/optimize/optimizer.py
```diff
@@ -720,6 +725,84 @@ def is_least_squares(self):
...
+    Parameters
+    ----------
+    options:
+        Optimizer options that are directly passed on to pyswarms.
+        c1: cognitive parameter
+        c2: social parameter
+        w: inertia parameter
+    """
+
+    def __init__(self, par_popsize: float = 10, options: Dict = None):
```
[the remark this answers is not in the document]
PhilippStaedter: Done

pypesto/optimize/optimizer.py
```diff
@@ -720,6 +725,84 @@ def is_least_squares(self):
         return False
 
 
+class PyswarmsOptimizer(Optimizer):
+    """
+    Global optimization using pyswarms.
```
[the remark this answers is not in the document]
PhilippStaedter: Done

pypesto/optimize/optimizer.py
```diff
@@ -722,6 +727,94 @@ def is_least_squares(self):
...
+
+            Returns:
+            --------
+            result: np.ndarray, shape (n_particles_in_swarm)
+            """
+
+            n_particles = swarm.shape[0]
+            result = np.zeros(n_particles)
+            for iPar, par in enumerate(swarm):
+                result[iPar] = problem.objective.get_fval(par)
```
paulstapor, line 800: This doesn't make sense to me:
In line 798, result is an array which is as long as we have particles. In line 800, it's assumed to have the size of the parameter vector... This looks like a bug.

PhilippStaedter: No, it should be correct, since ```enumerate(swarm)``` has the dimensions ```number_particles_in_swarm x number_parameters```.
In the for-loop, we assign ```iPar = number_particles_in_swarm``` and ```par = parameter```.
Since ```result = np.zeros(n_particles)```, we can index it by all number of particles and assign the value ```probem.objective.get_fval(par)```

paulstapor: Ah, thanks!
So, for my understanding: "swarm" is an N times M matrix. What does enumerate then do? Does it iterate over rows, columns, or elements? I would guess it iterates over columns, correct? Please just add a line of comment, that you actually iterate over columns, like "# iterate over found parameter vector, i.e., the columns in swarm"

For me, iPar always iterates over entries in the parameter vector, this is probably why I got confused. If iPar for you iterates over "individuals in the population/swarm", maybe go for iParVector, or iVector, or iSolution... If that's done, I'm fine!

paulstapor (approved):

pypesto/optimize/optimizer.py
yannikschaelte: enumerate wraps around a `for x in ...`, adding counters. when iterating over numpy arrays, iteration is always over the first dimension (=rows), i.e. over the particles, returning a parameter vector as `par`. Looks correct to me. Paul is right, iPar should be changed to `i_particle` or sth like that (lower case all).

yannikschaelte (approved):
Could you add a test as well? Then it would look all nice to me.

tox.ini
```diff
@@ -76,7 +76,7 @@ description =
 
 [testenv:notebooks1]
 allowlist_externals = bash
-extras = example,amici,petab,pyswarm,pymc3,cmaes,nlopt,fides
+extras = example,amici,petab,pyswarm,pymc3,cmaes,nlopt,fides,pyswarms
```
yannikschaelte, line 79: this is not needed here

tox.ini
```diff
@@ -54,7 +54,7 @@ description =
     Test basic functionality
 
 [testenv:petab]
-extras = test,test_petab,amici,petab,pyswarm
+extras = test,test_petab,amici,petab,pyswarm,pyswarms
```
yannikschaelte, line 57: this is not needed here

test/optimize/test_optimize.py
```diff
@@ -158,6 +159,8 @@ def get_optimizer(library, solver):
     elif library == 'scipydiffevolopt':
         optimizer = optimize.ScipyDifferentialEvolutionOptimizer(
             options=options)
+    elif library == 'pyswarms':
```
yannikschaelte, line 162: the options update in the pyswarms optimizer has to check whether options exist (see there, as otherwise just setting some options as here will fail)

pypesto/optimize/optimizer.py
```diff
@@ -722,6 +727,95 @@ def is_least_squares(self):
...
+    maxiter:
+        used to calculate the maximal number of funcion evaluations.
+        Default: 1000
+    """
+
+    def __init__(self, par_popsize: float = 10, options: Dict = None):
+        super().__init__()
+
+        if options is None:
+            options = {'maxiter': 1000, 'c1': 0.5, 'c2': 0.3, 'w': 0.9}
```
yannikschaelte, line 760:
```suggestion
        all_options = {'maxiter': 1000, 'c1': 0.5, 'c2': 0.3, 'w': 0.9}
        if options is None:
            options = {}
        all_options.update(options)
```

pypesto/optimize/optimizer.py
```diff
@@ -722,6 +727,95 @@ def is_least_squares(self):
...
+        self.options = options
```
yannikschaelte, line 761:
```suggestion
        self.options = all_options
```

tox.ini
PhilippStaedter: Done

tox.ini
PhilippStaedter: Done

pypesto/optimize/optimizer.py
PhilippStaedter: Done

pypesto/optimize/optimizer.py
PhilippStaedter: Done

test/optimize/test_optimize.py
PhilippStaedter: I am not sure I know what you mean with this comment...
what exactly should I change here?

yannikschaelte: here nothing, just in the class as mentioned above, cause otherwise this here doesn't work

PR #629, opened 2021-04-12
Caching Free Integrated Aesara Objective

FFroehlich:
Simplifies the construction of aesara objectives. This implementation allows exploiting integrated evaluation of the wrapped objective.

### DIFF
--- a/pypesto/objective/aesara.py
+++ b/pypesto/objective/aesara.py
@@ -10,7 +10,7 @@ from .base import ObjectiveBase, ResultDict
 from .constants import MODE_FUN, FVAL, GRAD, HESS
 
 
-from typing import Tuple
+from typing import Tuple, Optional
 
 try:
     import aesara
@@ -44,7 +44,7 @@ class AesaraObjective(ObjectiveBase):
                  objective: ObjectiveBase,
                  aet_x: TensorVariable,
                  aet_fun: TensorVariable,
-                 coeff: float):
+                 coeff: Optional[float] = 1.):
         if not isinstance(objective, ObjectiveBase):
             raise TypeError('objective must be an ObjectiveBase instance')
         if not objective.check_mode(MODE_FUN):
@@ -138,7 +138,7 @@ class AesaraObjectiveOp(Op):
 
     def __init__(self,
                  obj: AesaraObjective,
-                 coeff: float = 1.):
+                 coeff: Optional[float] = 1.):
         self._objective: AesaraObjective = obj
         self._coeff: float = coeff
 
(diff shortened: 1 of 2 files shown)


pypesto/objective/aesara.py
```diff
@@ -5,139 +5,235 @@
...
         else:
             self._log_prob_grad = None
 
     def perform(self, node, inputs, outputs, params=None):
-        theta, = inputs
-        log_prob = self._log_prob(theta)
+        log_prob = self._coeff * self._objective.inner_ret[FVAL]
         outputs[0][0] = np.array(log_prob)
 
     def grad(self, inputs, g):
```
yannikschaelte, line 164: ~~how~~by what dark magic is it possible that inputs is not used here? :thinking:

pypesto/objective/aesara.py
```diff
@@ -5,139 +5,235 @@
...
+    evaluation, caching it for later calls.
+    Caching is only enabled after the first time the gradient is asked for
+    and disabled whenever the cached gradient is not used,
+    in order not to increase computation time for derivative-free samplers.
+    Parameters
+    ----------
+    objective:
+        The `pypesto.ObjectiveBase` to wrap.
+    aet_x:
+        tensor variables that that define the variables of `aet_fun`
```
yannikschaelte, line 36: unify uppercase

yannikschaelte (approved):
looks very nice and useful!, don't get one point yet.

pypesto/objective/aesara.py
```diff
@@ -5,139 +5,235 @@
...
+        if 0 in sensi_orders:
+            ret[FVAL] = float(self.afun(x))
+        if 1 in sensi_orders:
+            ret[GRAD] = self.agrad(x)[0]
+        if 2 in sensi_orders:
+            ret[HESS] = self.ahess(x)[0]
+
+        return ret
+
+    def __deepcopy__(self, memodict=None):
```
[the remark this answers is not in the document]
FFroehlich: I think pickling would require some additional work, see https://aesara.readthedocs.io/en/latest/library/compile/function.html#aesara.compile.function.function_dump

pypesto/objective/aesara.py
FFroehlich: Added a note about this. The trick is that we already know what the inputs here are going to look like when computing `inner_ret`, so we can ignore them here and just use the precomputed values. This also has the advantage that we only spend a minimal amount of time in aesara functions, since the bulk of computation is done in the outer `call_unprocessed`, which improves performance in multithreaded settings as we avoid a lot of locking of aesara functions.

pypesto/objective/aesara.py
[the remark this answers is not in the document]
FFroehlich: I haven't tested it though.

PR #630, opened 2021-04-12
Update BioNetGen download URL (fixes CI)

dweindl:


FFroehlich (approved):
How switching to https://pypi.org/project/bionetgen/ at some point?

dweindl:
> How about switching to https://pypi.org/project/bionetgen/ at some point?

No strong opinion. As we'd only use it for downloading an archive, this adds an additional link that could break. On the other hand, it could go into the python dependencies removing the extra handling here.
If changed, I'd rather change it directly in AMICI and add it to the pysb extra-requirements there.

yannikschaelte (approved):

PR #631, opened 2021-04-12
allow ls optimization depending on settings in amici model

FFroehlich:
~Requires https://github.com/AMICI-dev/AMICI/pull/1489~


yannikschaelte (approved):
will likely not work with serialization, but that may be a problem on its own still (i.e. definition of model configurations)

Issue #636, opened 2021-04-16
Handle missing sub-results in `write_result`


dilpath:
Running `pypesto.store.save_to_hdf5.write_result(result)` if `result` has no sample sub-result causes an exception. Could specify `sample=False` but then adding additional sub-result objects (e.g. maybe ensembles) to `write_result` may break backwards compatibility in future, as the additional sub-result objects would be "missing" too.

Possible options:
- warn on missing sub-results (possibly in `SamplingResultHDF5Writer`) instead
- try/except in `write_result` for each sub-result type
- change all options in `write_result` to be `False` by default instead

PaulJonasJost:
Also Noticed a problem when loading files with `read_result()` if some directories are non existent within the file.

jvanhoefer:
@dilpath and @PaulJonasJost This is closed with the new release, right? I am wondering a bit why this was not closed automatically... (otherwise please reopen :) )

PR #637, opened 2021-04-16
Fix 636

PaulJonasJost:
Fixes Issue #636.

I did not include a try...except for each case as the sampling is special in the sense that the Class of `result.sampling_result` changes from `SampleResult` to `McmcPtResult` when sampling is performed. This created the error.

### DIFF
--- a/pypesto/store/read_from_hdf5.py
+++ b/pypesto/store/read_from_hdf5.py
@@ -207,7 +207,11 @@ class SamplingResultHDF5Reader:
             for key in f['/sampling/results'].attrs:
                 sample_result[key] = \
                     f['/sampling/results'].attrs[key]
-        self.results.sample_result = McmcPtResult(**sample_result)
+        try:
+            self.results.sample_result = McmcPtResult(**sample_result)
+        except TypeError:
+            logger.warning("Warning: You tried loading a non existent "
+                           "sampling result.")
 
         return self.results
 
--- a/pypesto/store/read_from_hdf5.py
+++ b/pypesto/store/read_from_hdf5.py
@@ -210,7 +210,7 @@ class SamplingResultHDF5Reader:
         try:
             self.results.sample_result = McmcPtResult(**sample_result)
         except TypeError:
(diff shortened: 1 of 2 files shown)


pypesto/store/read_from_hdf5.py
```diff
@@ -207,7 +207,11 @@ def read(self) -> Result:
             for key in f['/sampling/results'].attrs:
                 sample_result[key] = \
                     f['/sampling/results'].attrs[key]
-        self.results.sample_result = McmcPtResult(**sample_result)
+        try:
+            self.results.sample_result = McmcPtResult(**sample_result)
+        except TypeError:
+            logger.warning("Warning: You tried loading a non existent "
```
dilpath, line 213:
```suggestion
            logger.warning("Warning: You tried loading a non-existent "
```

pypesto/store/save_to_hdf5.py
```diff
@@ -182,6 +185,11 @@ def write(self, result: Result, overwrite: bool = False):
         """
         Write HDF5 sampling file from pyPESTO result object.
         """
+        # if there is no sample available, raise a warning and return
```
dilpath, line 188:
```suggestion
        # if there is no sample available, log a warning and return
```

pypesto/store/save_to_hdf5.py
```diff
@@ -182,6 +185,11 @@ def write(self, result: Result, overwrite: bool = False):
...
+        if(isinstance(result.sample_result, SampleResult)):
```
dilpath, line 189: Maybe add a comment about `SampleResult` being some dummy type that indicates no sampling has already been done (actual sampling is indicated by `McmcPtResult`)

dilpath (approved):
Thanks!

pypesto/store/save_to_hdf5.py
PaulJonasJost:
```suggestion
        # SampleResult is only a dummy class created by the Result class
        if(isinstance(result.sample_result, SampleResult)):
```

PR #638, opened 2021-04-19
Fix 636 (2)

PaulJonasJost:
adds as a fix to #636

### DIFF
--- a/pypesto/store/read_from_hdf5.py
+++ b/pypesto/store/read_from_hdf5.py
@@ -293,6 +293,7 @@ def read_result(filename: str,
     result = Result()
 
     if problem:
+        pypesto_problem_reader = ProblemHDF5Reader(filename)
         result.problem = pypesto_problem_reader.read()
 
     if optimize:
--- a/pypesto/store/read_from_hdf5.py
+++ b/pypesto/store/read_from_hdf5.py
@@ -293,11 +293,7 @@ def read_result(filename: str,
     result = Result()
 
     if problem:
-        pypesto_problem_reader = ProblemHDF5Reader(filename)
-        try:
-            result.problem = pypesto_problem_reader.read()
-        except Exception:
-            logger.warning('Loading the problem failed.')
+        result.problem = pypesto_problem_reader.read()
 
     if optimize:
         pypesto_opt_reader = OptimizationResultHDF5Reader(filename)
--- a/pypesto/store/read_from_hdf5.py
+++ b/pypesto/store/read_from_hdf5.py
@@ -290,21 +290,40 @@ def read_result(filename: str,
 
     if problem:
         pypesto_problem_reader = ProblemHDF5Reader(filename)
-        result.problem = pypesto_problem_reader.read()
+        try:
+            result.problem = pypesto_problem_reader.read()
+        except KeyError:
+            logger.warning('Loading the problem failed. It is'
+                           'highly likely that no problem exists.')
 
     if optimize:
         pypesto_opt_reader = OptimizationResultHDF5Reader(filename)
-        temp_result = pypesto_opt_reader.read()
-        result.optimize_result = temp_result.optimize_result
+        try:
+            temp_result = pypesto_opt_reader.read()
+            result.optimize_result = temp_result.optimize_result
+        except KeyError:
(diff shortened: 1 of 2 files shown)


pypesto/store/read_from_hdf5.py
```diff
@@ -294,21 +294,40 @@ def read_result(filename: str,
...
         pypesto_opt_reader = OptimizationResultHDF5Reader(filename)
-        temp_result = pypesto_opt_reader.read()
-        result.optimize_result = temp_result.optimize_result
+        try:
+            temp_result = pypesto_opt_reader.read()
+            result.optimize_result = temp_result.optimize_result
+        except KeyError:
+            logger.warning(f'Loading the optimization result failed. It is '
+                           f'highly likely that no optimization result exists '
+                           f'within {filename}.')
```
dilpath, line 311:
```suggestion
            logger.warning('Loading the optimization result failed. It is '
                           'highly likely that no optimization result exists '
                           f'within {filename}.')
```

pypesto/store/read_from_hdf5.py
```diff
@@ -294,21 +294,40 @@ def read_result(filename: str,
 
     if problem:
         pypesto_problem_reader = ProblemHDF5Reader(filename)
-        result.problem = pypesto_problem_reader.read()
+        try:
+            result.problem = pypesto_problem_reader.read()
+        except KeyError:
+            logger.warning('Loading the problem failed. It is'
+                           'highly likely that no problem exists.')
```
dilpath, line 301:
```suggestion
            logger.warning('Loading the problem failed. It is'
                           'highly likely that no problem exists '
                           f'within {filename}.')
```

To match the other messages.

pypesto/store/read_from_hdf5.py
```diff
@@ -294,21 +294,40 @@ def read_result(filename: str,
...
     if profile:
         pypesto_profile_reader = ProfileResultHDF5Reader(filename)
-        temp_result = pypesto_profile_reader.read()
-        result.profile_result = temp_result.profile_result
+        try:
+            temp_result = pypesto_profile_reader.read()
+            result.profile_result = temp_result.profile_result
+        except KeyError:
+            logger.warning(f'Loading the profiling result failed. It is '
+                           f'highly likely that no profiling result exists '
```
dilpath, line 320:
```suggestion
            logger.warning('Loading the profiling result failed. It is '
                           'highly likely that no profiling result exists '
```

pypesto/store/read_from_hdf5.py
```diff
@@ -294,21 +294,40 @@ def read_result(filename: str,
...
         pypesto_sample_reader = SamplingResultHDF5Reader(filename)
-        temp_result = pypesto_sample_reader.read()
-        result.sample_result = temp_result.sample_result
+        try:
+            temp_result = pypesto_sample_reader.read()
+            result.sample_result = temp_result.sample_result
+        except KeyError:
+            logger.warning(f'Loading the sampling result failed. It is '
+                           f'highly likely that no sampling result exists '
+                           f'within {filename}')
```
dilpath, line 331:
```suggestion
            logger.warning('Loading the sampling result failed. It is '
                           'highly likely that no sampling result exists '
                           f'within {filename}.')
```

dilpath (approved):
:+1:

pypesto/store/read_from_hdf5.py
PaulJonasJost: As Problem should always be written and read I corrected this to a regular warning that loading the problem failed. As of now I think most problems have amici objectives which have to be created again so that having the problem read in fail is in most cases no hindrance to the code.

dilpath: Hm then it would be handled differently to sample/optimize/profile results. There, only `KeyError` is handled to treat a known issue, and other exceptions would be raised in the usual way. Here, all exceptions are caught and replaced with a `logger.warning`. So it may be better/more consistent to remove the try-except for the `pypesto.Problem`.

If there is a particular exception in mind here, better to use the specific constant, (e.g. `KeyError` instead of `Exception`).

Just a small point, fine to ignore.

PaulJonasJost: Think you are right, just removed it as there is no particular exception👍

PR #639, opened 2021-04-22
Hdf5 History with MulitProcessEngine

PaulJonasJost:
The second try, as the first had complications when I pushed.


test/base/test_history.py
```diff
@@ -471,3 +473,65 @@ def test_trace_subset(history: pypesto.History):
                 assert isinstance(val, float)
             else:
                 assert isinstance(val, np.ndarray) or np.isnan(val)
+
+
+def test_hdf5_history_mp():
+    """Test whether hdf5-History works with a MultiProcessEngine"""
```
PaulJonasJost, line 479:
```suggestion
    """Test whether hdf5-History works with a MultiProcessEngine."""
```

pypesto/optimize/util.py
```diff
@@ -0,0 +1,70 @@
...
+from ..objective import HistoryOptions
+from ..store.save_to_hdf5 import get_or_create_group
+from pathlib import Path
+
+
+import h5py
+
+
+def check_hdf5_mp(history_options: HistoryOptions,
+                  engine: Engine):
```
yannikschaelte, line 11:
```suggestion
def check_hdf5_mp(
    history_options: HistoryOptions,
    engine: Engine,
) -> Union[str, None]:
```
```suggestion
def check_hdf5_mp(history_options: HistoryOptions,
                  engine: Engine):
```

pypesto/optimize/util.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    Create a folder for partial HDF5 files,
+    if a parallelization engine will be used.
+
+    Paramters
+    ---------
```
yannikschaelte, line 17:
```suggestion
    Parameters
    ----------
```
```suggestion
    Paramters
    ---------
```

pypesto/optimize/util.py
```diff
@@ -0,0 +1,70 @@
...
+    history_options.storage_file = str(partial_file_path)
+
+    # create hdf5 file that gathers the others within history group
+    with h5py.File(filename, mode='a') as f:
+        get_or_create_group(f, "history")
+    return filename
+
+
+def fill_hdf5_file(ret: list,
+                   filename: str):
```
yannikschaelte, line 48:
```suggestion
def fill_hdf5_file(
    ret: list,
    filename: str,
) -> None:
```
```suggestion
def fill_hdf5_file(ret: list,
                   filename: str):
```

pypesto/optimize/util.py
```diff
@@ -0,0 +1,70 @@
...
+        The result iterable returned by the optimization.
+    filename:
+        The filename of the hdf5 file in which the histories
+        are to be gathered.
+    """
+    with h5py.File(filename, mode='a') as f:
+        for result in ret:
+            id = result['id']
+            if f'history/{id}' in f:
+                del f[f'history/{id}']
```
yannikschaelte, line 66: this may be dangerous I think ... should this ever happen?

pypesto/optimize/util.py
yannikschaelte: dou you also see two suggestions? no idea what went wrong

yannikschaelte:
the idea is that a folder is created, in which there is then a separate file per start? and at the end all files are combined into one file?

shouldn't it be safe then to remove the previous files?

pypesto/optimize/util.py
PaulJonasJost: yes also see two suggestions. Guess, that you meant the one where something actually changes.

pypesto/optimize/util.py
PaulJonasJost: mhh perhaps i should change that to a warning/error? Or could go with an additional `overwrite` argument?

PaulJonasJost:
> 
> 
> the idea is that a folder is created, in which there is then a separate file per start? and at the end all files are combined into one file?
> 
> shouldn't it be safe then to remove the previous files?

The end file only has external links, meaning that it still points to the previous files. That way it saves all the saving and loading.

pypesto/optimize/util.py
```diff
@@ -0,0 +1,74 @@
+from ..engine import Engine, SingleCoreEngine
+from ..objective import HistoryOptions
+from ..store.save_to_hdf5 import get_or_create_group
+from pathlib import Path
```
PaulJonasJost, line 4:
```suggestion
from pathlib import Path
from typing import Union
```

yannikschaelte:
> > the idea is that a folder is created, in which there is then a separate file per start? and at the end all files are combined into one file?
> > shouldn't it be safe then to remove the previous files?
> 
> The end file only has external links, meaning that it still points to the previous files. That way it saves all the saving and loading.

ah ok. I guess that should work :+1:

pypesto/optimize/util.py
yannikschaelte: I would say error, no overwrite. there shouldn't be a scenario where one wants to overwrite previous results, right? when starting a new optimization, one should always use a new file, or alternatively append to an existing one, which must however be governed via the ids in the optimization already.

yannikschaelte (approved):

pypesto/optimize/util.py
jvanhoefer: What about doing an `overwrite ` flag as input and setting it to `False` as default...? Then it would be possible, but you would get an error, if the result already exists...

yannikschaelte: I think that would make sense. Also a function (in the public API) to merge all history files into the final one would be useful, as this allows to just export a single file in the end.

PaulJonasJost: The `overwrite` flag was exactly what I meant 👍 will do it.
Would suggest adding the function  to merge files in a separate issue? would perhaps be nice either way if you want to merge two different optimization runs/results?

PaulJonasJost: was about to implement the overwrite function. But the `fill_hdf5_file` function is called in minimize. As this is basically only called if we use an hdf5 history with multiprocessengine, I do not think that this justifies the additional argument overwrite in minimize. Would then just remove the deletion part as the scenario, as yannik pointed out, should not really occur.

PR #640, opened 2021-04-22
Add reading Ensemble features from optimization result

PaulJonasJost:
New features:

-  load Ensemble from an Optimization with or without usage of the history.
-  load Ensemble from an Hdf5 File


pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,164 @@ def from_sample(
...
+                break
+
+        if n_starts*max_per_start > max_size:
+            logger.info(f'The number of starts that can contribute an '
+                        f'ensemble vector muliplied with max_per_start '
+                        f'is higher than max_size. Thus we will lower '
+                        f'max_per_start. If you do not want this to '
+                        f'happen consider increasing max_size to '
+                        f'{max_per_start*n_starts} or decrease cutoff.')
+            max_per_start = math.floor(max_size/n_starts)
```
dilpath, line 490: Perhaps all possible vectors from the different starts should be collected first, then trimmed back until `max_size` is reached? e.g. given two history traces `trace_0` and `trace_1`, `max_size == 10` and `max_per_start == 10`, and given `len(sum(trace_0['fval'] <= cutoff)) == 7` and `len(sum(trace_1['fval'] <= cutoff)) == 3`, then I think the current implementation would result in an ensemble with 8 vectors, not 10? (ignoring that the current implementation stops at the first `fval` that is worse than `cutoff`)

dilpath (approved):
i think some tests might help clarify usage (and of course ensure behavior) a bit.

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,164 @@ def from_sample(
...
+    ):
+        """Construct an ensemble from an optimization result.
+
+        Parameters
+        ----------
+        result:
+            A pyPESTO result that contains a sample result.
+        cutoff:
+            Exclude parameters from the optimization if the
+            nllh is higher than the `cutoff`.
```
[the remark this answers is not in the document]
PaulJonasJost: Thought about it as well, have it currently in my idea notes 👍

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,164 @@ def from_sample(
         x_vectors = x_vectors.T
         return Ensemble(x_vectors, **kwargs)
 
+    @staticmethod
+    def from_optimization(
+            result: Result,
+            cutoff: float,
+            max_size: int,
```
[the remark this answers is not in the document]
PaulJonasJost: uh that's nice, thanks, did not think of the np.inf, will include it.

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,164 @@ def from_sample(
...
+        vector_tags = []
+        x_names = result.problem.x_names
+        lb = result.problem.lb_full
+        ub = result.problem.ub_full
+
+        for start in result.optimize_result.list:
+            # add the parameters from the next start as long as we
+            # did not reach maximum size and the next value is still
+            # lower than the cutoff value
+            if start['fval'] < cutoff and len(x_vectors) < max_size:
```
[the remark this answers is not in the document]
PaulJonasJost: for now moved it into an idea notes 👍

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,164 @@ def from_sample(
...
+            trace_fval = \
+                result.optimize_result.list[i]['history'].get_fval_trace()
+
+            # calculate number of candidates
+            n_cand = 0
+            for iter_fval in reversed(trace_fval):
+                if iter_fval < cutoff:
+                    n_cand += 1
+                else:
+                    break
```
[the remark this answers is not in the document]
PaulJonasJost: Did the rewrite, saved it as a discussion point 👍

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,164 @@ def from_sample(
...
+
+            # calculate distance between the vectors included in ensemble:
+            dist = math.floor((n_cand-1)/(max_per_start-1))
+
+            # add vectors to ensemble
+            if dist == 0:
+                x_vectors.append(trace_x[-1])
+                vector_tags.append((i, len(trace_x)))
+                continue
```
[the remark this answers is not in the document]
PaulJonasJost: I think this was an error in my part, I changed the setup now.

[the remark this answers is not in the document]
PaulJonasJost:
```suggestion
            # calculate distance between the vectors included in ensemble:
            dist = max(math.floor((n_cand)/(max_per_start)),1)
            for i in range(max(max_per_start, n_cand)):
```
changed it to this, should be better I think.

pypesto/ensemble/utils.py
```diff
@@ -57,6 +59,55 @@ def read_from_csv(path: str,
...
+                            ):
+    """
+    Function for creating an ensemble from hdf5 storage file.
+
+    Parameters:
+    ----------
+    filename:
+        Name or path of the hdf5 file.
+    type:
+        Which type of ensemble to create. From History, from
```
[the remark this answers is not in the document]
PaulJonasJost: how about `source`?

pypesto/ensemble/utils.py
```diff
@@ -57,6 +59,55 @@ def read_from_csv(path: str,
...
+    """
+    # TODO: add option HISTORY. Need to fix
+    #  reading history from hdf5.
+    if type == OPTIMIZATION:
+        if cutoff is None or max_size is None:
+            raise TypeError('You need to define cutoff and max_size '
+                            'for the ensemble creation to work.')
+        result = read_result(filename=filename,
+                             profile=False,
+                             sample=False)
```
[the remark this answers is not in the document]
PaulJonasJost: It was meant to keep Memory usage lower by only loading what is needed. May be unnecessary as it should get removed after the function but still should save time.
Would say as long as we have only the three (optimise, profile, sample) I would keep it, as you would have to set two values in either case (e.g. `sample = True, all = False` in your example and `profile=False, optimize=False`) when only loading one, but it you would have to set three if you wanted to save two in your case. But good to keep in mind 👍

pypesto/ensemble/ensemble.py
PaulJonasJost: Yes, but in this scenario you would have 5 from `trace_0` and 3 from `trace_1` which is already leaning towards `trace_0`. This would be even stronger if we kept `max_per_start==10` in that case it would be `7:3`. Will put it as a discussion point for next meeting :)

dilpath: Ah, if I specify a `cutoff` and `max_per_start` then I would expect `7:3`. Any bias that arises from the users choices should probably be handled by the user too right? Otherwise the user may wonder why the ensemble contains fewer vectors than expected.

e.g. could add an option like `max_diff_between_starts = 2` to ensure the number of vectors from any start is within two of the number of vectors from the start that contributes the fewest vectors (then `7:3` will be truncated to `5:3`)

paulstapor: I agree that just going for floor here is not ideal... This will result in (maybe much) smaller ensembles than necessary in some cases, which we don't want to happen. Another thing: Why going for `math.floor` rather than `np.floor`? Numpy is imported anyways, no need to import math as well...

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,155 @@ def from_sample(
...
+            max_per_start = math.floor(max_size/n_starts)
+
+        for i in range(n_starts):
+            trace_x = \
+                result.optimize_result.list[i]['history'].get_x_trace()
+            trace_fval = \
+                result.optimize_result.list[i]['history'].get_fval_trace()
+
+            # calculate number of candidates
+            n_cand = 0
```
paulstapor, line 497: I see the nice "fail fast" aspect of this implementation, but personally, I'd go for some list comprehension here...
`n_cand = sum([iter_fval for iter_fval in trace_fval if iter_fval < cutoff])`. That saves four lines...

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,155 @@ def from_sample(
...
+        if n_starts*max_per_start > max_size:
+            logger.info(f'The number of starts that can contribute an '
+                        'ensemble vector multiplied with max_per_start '
+                        'is higher than max_size. Thus we will lower '
+                        'max_per_start. If you do not want this to '
+                        'happen consider increasing max_size to '
+                        f'{max_per_start*n_starts} or decrease cutoff.')
+            max_per_start = math.floor(max_size/n_starts)
+
+        for i in range(n_starts):
```
paulstapor, line 490: Please: i_Start or i_MS or iMS or... not `i`. Never `i`. The reason why not is what just happened a few lines below...

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,155 @@ def from_sample(
...
+
+            # calculate number of candidates
+            n_cand = 0
+            for iter_fval in reversed(trace_fval):
+                if iter_fval > cutoff:
+                    break
+                n_cand += 1
+
+            # calculate distance between the vectors included in ensemble:
+            dist = max(math.floor(n_cand/max_per_start), 1)
```
paulstapor, line 504: This will somewhat truncate the maximum possible distance...
Maybe better:
```
indices = np.round(np.linspace(0, n_cand - 1, np.min(max_per_start, n_cand)))
x_vectors = [trace_x[ind] for ind in indices]
x_tags = [(iMS, len(trace_x) - 1 - ind) for ind in indices]
```

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,155 @@ def from_sample(
...
+            for i in range(max(max_per_start, n_cand)):
```
paulstapor, line 505: You reuse/shadow `i` here. That's why you should never use just `i` as iterator... ;)

paulstapor (approved):

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,155 @@ def from_sample(
...
+            The maximum size the ensemble should be.
+
+        Returns
+        -------
+        The ensemble.
+        """
+        x_vectors = []
+        vector_tags = []
+
+        for start in result.optimize_result.list:
```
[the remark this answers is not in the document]
PaulJonasJost: yes, as the minimize function has an inherent sorting here (https://github.com/ICB-DCM/pyPESTO/blob/625c41911fdee2608e736321d9e9050ac90f7ba4/pypesto/optimize/optimize.py#L119) I assumed that the optimization results should all be sorted.

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,155 @@ def from_sample(
...
+        x_vectors = []
+        vector_tags = []
+        x_names = result.problem.x_names
+        lb = result.problem.lb_full
+        ub = result.problem.ub_full
+
+        # calculate the number of starts whose final nllh is below cutoff
+        n_starts = 0
+        for start in result.optimize_result.list:
+            if start['fval'] > cutoff:
```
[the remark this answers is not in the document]
PaulJonasJost: it should be. The only time this might not be (was an issue some time ago with plotting) is when you combine two different optimization runs. But I would exclude this for now tbh.

pypesto/ensemble/ensemble.py
PaulJonasJost: if I sum that up, wouldn't that be then the sum of the fvals? I think `n_can = len([iter_fval for iter_fval in trace_fval if iter_fval < cutoff])` would be correct here then?

paulstapor: You're right. It should be either something using len or summing over 1!

pypesto/ensemble/utils.py
```diff
@@ -57,6 +59,52 @@ def read_from_csv(path: str,
...
+    ensemble:
+        Ensemble object of parameter vectors
+    """
+    # TODO: add option HISTORY. Need to fix
+    #  reading history from hdf5.
+    if type == OPTIMIZE:
+        result = read_result(filename=filename,
+                             profile=False,
+                             sample=False)
+        return Ensemble.from_optimization_endpoints(result=result,
```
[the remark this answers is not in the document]
PaulJonasJost: that is added as a todo above. Currently if you load the file through read_result, the history is not loaded. I wanted to first fix that and then add the third option `HISTORY`.

pypesto/ensemble/ensemble.py
PaulJonasJost: the issue here is that it will sum over all possible iterations, we have that the iterations might not be seamless, as the optimization is not strictly decreasing.

pypesto/ensemble/utils.py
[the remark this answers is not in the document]
PaulJonasJost: went with input_type 👍

pypesto/ensemble/ensemble.py
dilpath:
```python
>>> import numpy as np
>>> b = np.array([1,2,3,4,5,6,4,3,2,1])
>>> np.sum(b < 5)
8
>>> np.where(np.diff(b < 5))[0][0] + 1
4
```

So setting `n_cand` to the last expression is probably what you want? With some comment to describe what it does since it might be confusing to someone unfamiliar with the code

pypesto/ensemble/utils.py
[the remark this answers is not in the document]
dilpath: Hm agreed, could be
```python
def read_result(
    filename: str,
    optimize: bool = False,
    profile: bool = False,
    sample: bool = False,
):
    if not any([optimize, profile, sample]):
        optimize = True
        profile = True
        sample = True

    # read optimize
    # read profile
    # read sample
```

Then everything is loaded unless the user specifies something in particular.

Or use the `all` argument above, but set its default to `False`, then the user must always specify what to load otherwise nothing is loaded.

I would prefer one of these two options (in particular the option without `all`)

pypesto/ensemble/ensemble.py
yannikschaelte: instead of funny where[0][0], consider also numpy nonzero and numpy flatnonzero ;)

dilpath: Also `np.diff(b<5).argmax()`

pypesto/ensemble/utils.py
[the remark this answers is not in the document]
PaulJonasJost: I like that one, will go for it. Should I do it in this PR or separately as it is quite a different topic (not sure on how to handle that best).

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +371,155 @@ def from_sample(
...
+            else:
+                break
+
+        # print a warning if there are no vectors within the ensemble
+        if len(x_vectors) == 0:
+            raise ValueError('The ensemble does not contain any vectors. '
+                             'Either the cutoff value was too small or the '
+                             'result.optimize_result object might be empty.')
+        elif len(x_vectors) < max_size:
+            logger.info(f'The ensemble contains {len(x_vectors)} parameter '
```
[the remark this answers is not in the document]
PaulJonasJost: would leave that to @dilpath in a democratic vote then 👍

pypesto/ensemble/ensemble.py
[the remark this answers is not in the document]
PaulJonasJost: as I do not fully understand, I would suggest we quickly discuss that at the next meeting :)

pypesto/ensemble/utils.py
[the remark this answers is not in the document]
PaulJonasJost: merged the PR 👍

pypesto/ensemble/ensemble.py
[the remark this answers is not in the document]
PaulJonasJost: outdated and already discussed in another comment.

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +370,153 @@ def from_sample(
...
+                        'vectors, which is less than the maximum size. If '
+                        'you want to include more vectors you can consider '
```
yannikschaelte, line 419:
```suggestion
                        'you want to include more vectors, you can consider '
```

pypesto/ensemble/ensemble.py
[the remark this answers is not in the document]
yannikschaelte: both fine :guitar:

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +370,153 @@ def from_sample(
...
+            cutoff: float = np.inf,
+            max_size: int = np.inf,
+            max_per_start: int = np.inf,
+            **kwargs,
+    ):
+        """Construct an ensemble from the history of an optimization.
+
+        Parameters
+        ----------
+        result:
```
yannikschaelte, line 443: clarify whether this is the original object, or  deep copy.

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +370,153 @@ def from_sample(
...
+        ub = result.problem.ub_full
+
+        # calculate the number of starts whose final nllh is below cutoff
+        n_starts = 0
+        for start in result.optimize_result.list:
+            if start['fval'] > cutoff:
+                break
+            n_starts += 1
+
+        if n_starts*max_per_start > max_size:
```
yannikschaelte, line 480:
```suggestion
        if n_starts * max_per_start > max_size:
```

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +370,153 @@ def from_sample(
...
+            logger.info(f'The number of starts that can contribute an '
```
yannikschaelte, line 481: not sure how this looks on screen, but consider also using line breaks after e.g. 80 chars

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +370,153 @@ def from_sample(
...
+                        'ensemble vector multiplied with max_per_start '
+                        'is higher than max_size. Thus we will lower '
+                        'max_per_start. If you do not want this to '
+                        'happen consider increasing max_size to '
+                        f'{max_per_start*n_starts} or decrease cutoff.')
+            max_per_start = np.floor(max_size/n_starts)
+
+        for i_MS in range(n_starts):
```
yannikschaelte, line 489:
```suggestion
        for i_ms in range(n_starts):
```

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +370,153 @@ def from_sample(
...
+            trace_x = \
+                result.optimize_result.list[i_MS]['history'].get_x_trace()
+            trace_fval = \
+                result.optimize_result.list[i_MS]['history'].get_fval_trace()
+
+            # calculate number of candidates (argmax returns the indice right
```
yannikschaelte, line 495:
```suggestion
            # calculate number of candidates (argmax returns the index right
```

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +370,153 @@ def from_sample(
...
+            # before the first time the fval is higher than the cutoff
```
yannikschaelte, line 496:
```suggestion
            # before the first time the fval is higher than the cutoff)
```

pypesto/ensemble/utils.py
```diff
@@ -57,6 +59,50 @@ def read_from_csv(path: str,
...
+                            input_type: str = OPTIMIZE,
+                            remove_burn_in: bool = True,
+                            chain_slice: slice = None,
+                            cutoff: float = np.inf,
+                            max_size: int = np.inf
+                            ):
+    """
+    Create an ensemble from an HDF5 storage file.
+
+    Parameters:
```
yannikschaelte, line 72:
```suggestion
    Parameters
```

pypesto/ensemble/utils.py
```diff
@@ -57,6 +59,50 @@ def read_from_csv(path: str,
...
+                                                    max_size=max_size)
+    elif input_type == SAMPLE:
+        result = read_result(filename=filename,
+                             sample=True)
+        return Ensemble.from_sample(result=result,
+                                    remove_burn_in=remove_burn_in,
+                                    chain_slice=chain_slice)
+    else:
+        raise ValueError('The type you provided was neither '
+                         '"sample" nor "optimization". Those are '
```
yannikschaelte, line 101: use string constnts here

yannikschaelte: dmn, now on my keybord lso the _ key is broken. rrg tht's wful.

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +370,153 @@ def from_sample(
...
+            single optimization start.
+
+        Returns
+        -------
+        The ensemble.
+        """
+        if not result.optimize_result.list[0].history.options['trace_record']:
+            logger.warning('The optimize result has no trace. The Ensemble '
+                           'will automatically be created through '
+                           'from_optimization().')
```
yannikschaelte, line 462: _endpoints?

yannikschaelte (approved):

pypesto/ensemble/ensemble.py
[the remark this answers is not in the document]
yannikschaelte: I am not sure whether we always make sure that the values are sorted, e.g. when loading from file / appending. it may be useful to, as I think we do now in visualization, always explicitly sort when the n best points or the order is needed.

pypesto/ensemble/utils.py
PaulJonasJost: I can still read it 👍 so define this string in the `constants.py`?

yannikschaelte: It is e.g. `SAMPLE == 'sample'`, right? Then I'd just use `f'{SAMPLE} nor {OPTIMIZATION}. Those are '` here.

pypesto/ensemble/ensemble.py
```diff
@@ -367,6 +370,153 @@ def from_sample(
...
+                        'ensemble vector multiplied with max_per_start '
+                        'is higher than max_size. Thus we will lower '
+                        'max_per_start. If you do not want this to '
+                        'happen consider increasing max_size to '
+                        f'{max_per_start*n_starts} or decrease cutoff.')
+            max_per_start = np.floor(max_size/n_starts)
+
+        for i_MS in range(n_starts):
+            trace_x = \
+                result.optimize_result.list[i_MS]['history'].get_x_trace()
```
PaulJonasJost, line 491:
```suggestion
                result.optimize_result.list[i_ms]['history'].get_x_trace()
```

pypesto/ensemble/utils.py
PaulJonasJost:
```suggestion
                         f'"{SAMPLE}" nor "{OPTIMIZE}". Those are '
```

jvanhoefer:
maybe before merging a more explanatory PR title would be nice (for future references, e.g. when preparing a new release)

pypesto/ensemble/ensemble.py
```diff
@@ -747,3 +749,97 @@ def check_identifiability(self) -> pd.DataFrame:
             parameter_identifiability['parameterId']
 
         return parameter_identifiability
+
+
+def entries_per_start(fval_trace: List['np.ndarray'],
```
dilpath, line 754:
```suggestion
def entries_per_start(fval_traces: List[np.ndarray],
```

pypesto/ensemble/ensemble.py
```diff
@@ -747,3 +749,97 @@ def check_identifiability(self) -> pd.DataFrame:
...
+    -------
+        A list of number of candidates per start that are to
+        be included in the ensemble.
+
+    """
+    # choose possible candidates
+    ens_ind = [np.flatnonzero(fval <= cutoff) for fval in fval_trace]
+
+    # count the number of candidates per start
+    n_per_starts = np.array([len(start) for start in ens_ind])
```
dilpath, line 785:
```suggestion
    n_per_start = np.array([len(start) for start in ens_ind])
```

pypesto/ensemble/ensemble.py
```diff
@@ -747,3 +749,97 @@ def check_identifiability(self) -> pd.DataFrame:
...
+    #  entries from certain starts. This would brng up the
+    #  discussion which starts to choose. One obvious choice
+    #  would be the best starts based on their endpoint.
+
+    return n_per_starts
+
+
+def indices_per_start(trace_start: np.ndarray,
+                      cutoff: float,
+                      n_per_start: int,
```
dilpath, line 815:
```suggestion
                      n_vectors: int,
```

pypesto/ensemble/ensemble.py
```diff
@@ -747,3 +749,97 @@ def check_identifiability(self) -> pd.DataFrame:
...
+    #  a flag which would be `force_max`, that indicates
+    #  whether those remaining free slots should be filled by
+    #  entries from certain starts. This would brng up the
+    #  discussion which starts to choose. One obvious choice
+    #  would be the best starts based on their endpoint.
+
+    return n_per_starts
+
+
+def indices_per_start(trace_start: np.ndarray,
```
dilpath, line 813:
```suggestion
def get_vector_indices(trace_start: np.ndarray,
```

pypesto/ensemble/ensemble.py
```diff
@@ -747,3 +749,97 @@ def check_identifiability(self) -> pd.DataFrame:
...
+    n_per_start:
+        The number of indices to be included from one start.
+    dist_or_best:
+        Boolean flag, whether the best (False) values from the
+        start should be taken or whether the indices should be
+        more evenly distributed.
+
+    Returns
+    -------
+        The indices which to include in the ensemble.
```
dilpath, line 836:
```suggestion
        The indices to include in the ensemble.
```

pypesto/ensemble/ensemble.py
```diff
@@ -747,3 +749,97 @@ def check_identifiability(self) -> pd.DataFrame:
...
+    """
+
+    candidates = np.flatnonzero(trace_start <= cutoff)
+
+    if dist_or_best:
+        indices = np.round(np.linspace(0, len(candidates)-1, n_per_start))
+        return candidates[indices.astype(int)]
```
dilpath, line 843: Maybe convert to set? Otherwise is it possible that e.g. `len(candidates)-1==1` and `n_per_start==6` so `indices==[0, 0, 0, 1, 1, 1]` and so the returned vector indices are duplicates?

test/base/test_ensemble.py
```diff
@@ -0,0 +1,53 @@
+import numpy as np
+import pypesto
+from pypesto.ensemble import Ensemble
+import scipy.optimize as so
+
+
+def test_ensemble_from_optimization():
+    """
+    Test, whether reading ensemble from history works as intended.
```
dilpath, line 9:
```suggestion
    Test reading an ensemble from history.
```

pypesto/ensemble/ensemble.py
PaulJonasJost: no n_per_start will give you a number that will definitely be lower than len(candidates) as per code so this should not be a problem. I do like lists more but that is a personal preference :D

pypesto/ensemble/constants.py
```diff
@@ -23,6 +23,10 @@
 PREDICTION_ARRAYS = 'prediction_arrays'
 PREDICTION_SUMMARY = 'prediction_summary'
 
+HISTORY = 'history'
+OPTIMIZE = 'optimization'
```
dilpath, line 27:
```suggestion
OPTIMIZE = 'optimize'
```

pypesto/ensemble/ensemble.py
dilpath: Ah, if it's not intended for a user to call this method directly, then the name can be prefixed with a `_` e.g. `_indices_per_start`

test/base/test_ensemble.py
PaulJonasJost:
```suggestion
    Test reading an ensemble from optimization result.
```

paulstapor (commented):
Looks good! Only really minor stuff left

pypesto/ensemble/ensemble.py
```diff
@@ -594,3 +749,97 @@ def check_identifiability(self) -> pd.DataFrame:
...
+
+    # trimm down starts that exceed the limit:
+    n_per_start = [min(n, max_per_start) for n in n_per_start]
+
+    # trimm down more until it fits the max size
+    decr = 0
+    while(sum(n_per_start) > max_size):
+        n_per_start = [min(n, max_per_start-decr)
+                       for n in n_per_start]
+        decr += 1
```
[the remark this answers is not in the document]
PaulJonasJost: what would you have in mind? :)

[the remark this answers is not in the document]
paulstapor: Something that exactly matches max_size whenever possible?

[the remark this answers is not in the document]
PaulJonasJost: no I meant what would your little hack be 🙂

PR #644, opened 2021-04-23
Add parameter names to check_grad results

FFroehlich:


pypesto/objective/base.py
```diff
@@ -527,7 +527,8 @@ def check_grad(
             data = {**prefix_data, **data, **postfix_data}
 
         # create dataframe
-        result = pd.DataFrame(data=data)
+        result = pd.DataFrame(data=data,
+                              index=[self.x_names[ix] for ix in x_indices])
```
dilpath, line 531: `self.x_names` defaults to `None`. Not sure if this will ever be the case here though.

dilpath (approved):
:+1:

pypesto/objective/base.py
yannikschaelte: I think whenever parameter names are used, a good default are p0, ... / p_0, ... / x0, ..., or sth like that.

yannikschaelte (approved):

jvanhoefer (approved):

pypesto/objective/base.py
FFroehlich: cannot add default names to objective since dimension of x is not known at initialization

pypesto/objective/base.py
```diff
@@ -527,7 +527,9 @@ def check_grad(
             data = {**prefix_data, **data, **postfix_data}
 
         # create dataframe
-        result = pd.DataFrame(data=data)
+        result = pd.DataFrame(data=data,
+                              index=[self.x_names[ix] for ix in x_indices]
+                              if self.x_names is not None else None)
```
dilpath, line 532:
```suggestion
                              index=[
                                  self.x_names[ix]
                                      if self.x_names is not None
                                      else f'x_{ix}'
                                  for ix in x_indices
                              ])
```

re: Yannik's suggestion

FFroehlich: Yes that could be done, but it doesn't really add any additional information beyond the default pandas range index, right?

dilpath: Yup not much more info, but e.g. if `x_indices = [5, 10, 15]` then the index will be `['x_5', 'x_10', 'x_15']` instead of `[0,1,2]`

FFroehlich: Oooh thats a good point!

PR #645, opened 2021-04-24
Fix #642

### LINKED ISSUE #642: Saving/loading changes profiles
The following snippet should produce the same profile visualization, but instead produces the following images (the first image is the expected figure). Note that the parameter ID in the plots change too.
```python
# plot
pypesto.visualize.profiles(pypesto_result)
plt.savefig(str(output_path / 'plot_before_saving.png'))

# save and load
pypesto.store.write_result(
    result=pypesto_result,
    filename=str(output_path / 'pypesto_result.hdf5'),
    sample=False,
)

loaded_result = pypesto.store.read_result(
    str(output_path / 'pypesto_result.hdf5'),
    sample=False,
)

# plot loaded result
pypesto.visualize.profiles(loaded_result)
plt.savefig(str(output_path / 'plot_after_saving.png'))
```
"plot_before_saving.png"
plot_before_saving
"plot_after_saving.png"
plot_after_saving

Here's the code to reproduce.
```bash
bash mkenv.sh
source venv_bug1/bin/activate
python bug.py
```
[mwe.zip](https://github.com/ICB-DCM/pyPESTO/files/6366159/mwe.zip)

PaulJonasJost:
Fixes #642.

As far as I could see, the problem was that the order of the parameter_id's is not kept. This should fix it.

### DIFF
--- a/pypesto/store/read_from_hdf5.py
+++ b/pypesto/store/read_from_hdf5.py
@@ -247,17 +247,19 @@ class ProfileResultHDF5Reader:
                 problem_reader = ProblemHDF5Reader(self.storage_filename)
                 self.results.problem = problem_reader.read()
             for profile_id in f['/profiling']:
-                profiling_list.append([])
+                profiling_list.append([
+                    None for _ in range(len(f[f'/profiling/{profile_id}']))
+                ])
+                breakpoint()
                 for parameter_id in f[f'/profiling/{profile_id}']:
                     if f[f'/profiling/{profile_id}/'
                          f'{parameter_id}'].attrs['IsNone']:
-                        profiling_list[int(profile_id)].append(None)
+                        continue
                     else:
-                        profiling_list[int(profile_id)]\
-                            .append(
+                        profiling_list[int(profile_id)][int(parameter_id)] = \
                             read_hdf5_profile(f,
                                               profile_id=profile_id,
-                                              parameter_id=parameter_id))
+                                              parameter_id=parameter_id)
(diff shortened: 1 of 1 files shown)


pypesto/store/read_from_hdf5.py
```diff
@@ -247,17 +247,18 @@ def read(self) -> Result:
...
-                profiling_list.append([])
+                profiling_list.append([
+                    None for _ in range(len(f[f'/profiling/{profile_id}']))
+                ])
                 for parameter_id in f[f'/profiling/{profile_id}']:
                     if f[f'/profiling/{profile_id}/'
                          f'{parameter_id}'].attrs['IsNone']:
-                        profiling_list[int(profile_id)].append(None)
+                        continue
                     else:
```
dilpath, line 257: Can remove this `else` since the `if` has a `continue`

pypesto/store/read_from_hdf5.py
```diff
@@ -247,17 +247,18 @@ def read(self) -> Result:
                 problem_reader = ProblemHDF5Reader(self.storage_filename)
                 self.results.problem = problem_reader.read()
             for profile_id in f['/profiling']:
-                profiling_list.append([])
+                profiling_list.append([
+                    None for _ in range(len(f[f'/profiling/{profile_id}']))
```
dilpath, line 251:
```suggestion
                    None for _ in f[f'/profiling/{profile_id}']
```

dilpath (approved):
Thanks!

PR #647, opened 2021-04-29
add docu and slightly refactor latin hypercube sampling...

jvanhoefer:
When using latin hypercube sampling I noticed:

* There is no docu (therefore I had to go through the whole code to understand the meaning if the `smooth` argument...). 
* The sampling itself was slightly more complicated then necessary, so I refactored that...

### DIFF
--- a/pypesto/startpoint/latin_hypercube.py
+++ b/pypesto/startpoint/latin_hypercube.py
@@ -67,7 +67,7 @@ def _latin_hypercube(n_starts: int,
         hypercube [i/n_starts, (i+1)/n_starts] should be chosen
         (`smooth==True`) or the midpoint of the interval (`smoth==False`).
     """
-    xs = np.nan * np.ones((n_starts, dim))
+    xs = np.empty((n_starts, dim))
 
     for i_dim in range(dim):
         xs[:, i_dim] = np.random.permutation(np.arange(n_starts))
--- a/pypesto/startpoint/latin_hypercube.py
+++ b/pypesto/startpoint/latin_hypercube.py
@@ -19,7 +19,7 @@ def latin_hypercube(**kwargs) -> np.ndarray:
         upper bound.
 
     smooth:
-        indicates, if a (uniformly chosen) random starting point within the
+        indicates if a (uniformly chosen) random starting point within the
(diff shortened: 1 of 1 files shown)


pypesto/startpoint/latin_hypercube.py
```diff
@@ -5,6 +5,23 @@
 def latin_hypercube(**kwargs) -> np.ndarray:
```
PaulJonasJost, line 5:
```suggestion
def latin_hypercube(n_starts: int,
        lb: Union[np.ndarray, List[float]],
        ub: Union[np.ndarray, List[float]],
        smooth: bool) -> np.ndarray:
```
Would add the parameters directly. Seems more informative to me then kwargs. Or is there a special reason for `**kwargs`?

PaulJonasJost (approved):

pypesto/startpoint/latin_hypercube.py
jvanhoefer: The interface for the sampling routines are quite various, therefore all sampling routines have the `**kwargs` argument currently (since they are not completely standardized and may use different ones, depending on the methods...). Since this isn't the focus of this PR, I would leave this one to an other refactoring of the starting value sampling...

PaulJonasJost: perfectly fine with that :)

pypesto/startpoint/latin_hypercube.py
```diff
@@ -30,19 +47,30 @@ def latin_hypercube(**kwargs) -> np.ndarray:
...
+        dimension of the optimization problem.
+
+    smooth:
+        indicates, if a (uniformly chosen) random starting point within the
+        hypercube [i/n_starts, (i+1)/n_starts] should be chosen
+        (`smooth==True`) or the midpoint of the interval (`smoth==False`).
     """
-    # uniform points
-    xs = np.random.random((n_starts, dim))
+    xs = np.nan * np.ones((n_starts, dim))
```
FFroehlich, line 70:
```suggestion
    xs = np.empty((n_starts, dim))
```

yannikschaelte (approved):

PR #648, opened 2021-04-29
Use model name in output directory if provided to PEtab importer

dilpath:
Currently described as using a provided `model_name` but doesn't.

https://github.com/ICB-DCM/pyPESTO/blob/41c470bd2ef84258b22eb9f290f954c75c223811/pypesto/petab/importer.py#L33-L42


yannikschaelte (approved):
whoopsie :see_no_evil:

PR #649, opened 2021-04-30
changed default values of read_result

PaulJonasJost:
This way it still loads everything given no arguments, but only loads desired values if anything is specified.


dilpath (approved):
Thanks!

PR #653, opened 2021-05-04
Bugfix: custom timepoints after serialization

dilpath:
The method is renamed from `custom_timepoints` to `set_custom_timepoints`


pypesto/objective/amici.py
```diff
@@ -423,10 +432,27 @@ def custom_timepoints(
...
+            ]
+
+        amici_objective.custom_timepoints = custom_timepoints
+
         for index in range(len(amici_objective.edatas)):
-            if timepoints_global is not None:
-                amici_objective.edatas[index].setTimepoints(timepoints_global)
-            else:
-                amici_objective.edatas[index].setTimepoints(timepoints[index])
+            amici_objective.edatas[index].setTimepoints(
```
FFroehlich, line 454: I think this should be implemented as separate method (`apply_custom_timepoints`) that can also be called from `__setstate__`

FFroehlich (approved):

pypesto/objective/amici.py
```diff
@@ -191,6 +191,8 @@ def __init__(self,
         self.calculator = calculator
         super().__init__(x_names=x_names)
 
+        self.custom_timepoints = None
```
yannikschaelte, line 194: document what this attribute is for

pypesto/objective/amici.py
```diff
@@ -253,6 +255,13 @@ def __setstate__(self, state: Dict):
         solver = self.amici_object_builder.create_solver(model)
         edatas = self.amici_object_builder.create_edatas(model)
 
+        # Set custom timepoints if available
```
yannikschaelte, line 258: move to function (as Fabi said)

yannikschaelte (approved):

PR #657, opened 2021-05-06
do not import predict via import pypesto fixes #656

### LINKED ISSUE #656: Do not import prediction with import pypesto
Currently `import pypesto` only imports basic functionality, that is needed _everywhere_ (e.g. problem/result objects) AND parts of the prediction. 

Would highly be in favor of only importing this only with `import pypesto.predict`, as we do that with everything else, even optimization.

jvanhoefer:


PaulJonasJost (approved):

test/petab/test_amici_predictor.py
```diff
@@ -13,7 +13,7 @@
 import libsbml
 import petab
 
-from pypesto.predict import PredictionResult, PredictionConditionResult
+from pypesto.prediction import PredictionResult, PredictionConditionResult
```
dilpath, line 16:
```suggestion
from pypesto.predict import PredictionResult, PredictionConditionResult
```

PaulJonasJost (approved):

PR #658, opened 2021-05-07
Feature validation profiles closes #94

### LINKED ISSUE #94: Prediction profiles
At some point in the more or less near future, we will want to have prediction profiles in pyPESTO. Pretty much of the parameter profiling code could be reused, I think/hope...

jvanhoefer:
Provides basic functionality to compute validation significances, as in Kreutz et al. 2012.

Basically one has to provide a `pypesto.Problem` encoding the fitting of the full data set as well as `pypesto.result`s for the training and the full data set. The functionality then computes the significance of the validation data set.

For a whole "validation profile" (i.e. a profile of a new data point in the "data space"...) one would need to specify the concept of "data" in more detail. Quite possible to also do this via one more layer around the functionality provided in this PR (i.e. write a loop, that gradually in-/decreases the new measurement value and computes the `validation_profile_significance`...).


pypesto/profile/validation_intervals.py
```diff
@@ -0,0 +1,106 @@
+import logging
```
yannikschaelte, line 1:
```suggestion
"""Validation intervals."""

import logging
```
some file level docstring is always good, more informative would be good

pypesto/profile/validation_intervals.py
```diff
@@ -0,0 +1,106 @@
...
+        problem_full_data: Problem,
+        result_training_data: Result,
+        result_full_data: Optional[Result] = None,
+        n_starts: int = 1,
+        optimizer: Optional[Optimizer] = None,
+        engine: Engine = None,
+        lsq_objective: float = False
+) -> float:
+    """
+    Computes the significance of an validation experiment as described in
```
yannikschaelte, line 24:
```suggestion
    Computes the significance of a validation experiment as described in
```

pypesto/profile/validation_intervals.py
```diff
@@ -0,0 +1,106 @@
...
+    Kreutz et al. BMC Systems Biology 2012.
```
yannikschaelte, line 25: an actual reference would be good to make it easier to find in the future. See https://github.com/ICB-DCM/pyABC/blob/main/pyabc/inference/smc.py#L61 (and maybe use an actual url :see_no_evil: )

pypesto/profile/validation_intervals.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def validation_profile_significance(
+        problem_full_data: Problem,
+        result_training_data: Result,
+        result_full_data: Optional[Result] = None,
+        n_starts: int = 1,
+        optimizer: Optional[Optimizer] = None,
+        engine: Engine = None,
+        lsq_objective: float = False
```
yannikschaelte, line 21:
```suggestion
        lsq_objective: float = False,
```
just playing flake8 ...

pypesto/profile/validation_intervals.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    Computes the significance of an validation experiment as described in
+    Kreutz et al. BMC Systems Biology 2012.
+
+
+    Parameters
+    ----------
+    problem_full_data:
+        pypesto.problem, such that the objective is the
+        negative-log-likelihood of the training and validation data set.
```
yannikschaelte, line 32: what would be if there is a Bayesian prior defined?

pypesto/profile/validation_intervals.py
```diff
@@ -0,0 +1,106 @@
...
+
+    result_training_data:
+        result object from the fitting of the training data set only.
+
+    result_full_data
+        pypesto.result object, that either contains, or (if fitting was not
```
yannikschaelte, line 38:
```suggestion
        pypesto.result object that either contains, or (if fitting was not
```
using "that" in the sense of "which", unlike for which there is usually no comma.

yannikschaelte (changes requested):

pypesto/profile/validation_intervals.py
```diff
@@ -0,0 +1,106 @@
...
+            dim_full=problem_full_data.dim_full,
+            x_fixed_indices=problem_full_data.x_fixed_indices,
+            x_fixed_vals=problem_full_data.x_fixed_vals,
+            x_guesses=x_0,
+            startpoint_method=problem_full_data.startpoint_method,
+            x_names=problem_full_data.x_names,
+            x_scales=problem_full_data.x_scales,
+            x_priors_defs=problem_full_data.x_priors,
+            lb_init=problem_full_data.lb_init,
+            ub_init=problem_full_data.ub_init)
```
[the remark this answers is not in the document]
yannikschaelte: hm ok. in that case it may be good to implement the correct methods in the problem class, either directly setting the required attributes via set functions, or via a copy constructor. This avoids the likely scenario that problem will get additional arguments in the future, and then this call here will be incomplete.

test/profile/test_validation_intervals.py
```diff
@@ -0,0 +1,63 @@
+"""
+This is for testing profile based validation intervals.
+"""
+
+import numpy as np
+import unittest
```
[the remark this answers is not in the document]
yannikschaelte: Haha, my policy would always be to use pytest instead of the class-syntax, which I find hard to read. Would be possible here for sure, but all would be fine for me.

yannikschaelte (approved):
looks good to me

PR #663, opened 2021-05-12
added option for hdf5 History to read_result_from_file

### LINKED ISSUE #259: Fill result from history
When a run crashes, there should be a function to fill a result from history (if such a one has been saved to file via csv/hdf5, that is). Partly exists already in the pypesto.optimize.optimizer.full_result_from_history function, but the rest of the workflow, both from csv and hdf5, should be added, to create the result object and finally store everything in the result hdf5 file.

PaulJonasJost:
closes #259.


pypesto/objective/history.py
```diff
@@ -1115,10 +1118,10 @@ def _compute_vals_from_trace(self):
         # etc
         max_init_iter = 3
         for it in range(min(len(self.history), max_init_iter)):
-            candidate = self.history.get_fval_trace(it)
+            candidate = self.history.get_fval_trace()[it]
```
yannikschaelte, line 1121: why is this necessary? passing `it` into the function was designed to allow for efficient access, e.g. not calculating the full trajectory, but only that specific index/indices. Not sure if csv/hdf5 use it, but could potentially.

PaulJonasJost: csv uses it, but hdf5 did not that's why I changed it for now into this to be able to use the function in both file formats.

yannikschaelte: oh, I see. That's a bug in the History class: It seems like when `it` was introduced, it was not properly propagated down to hdf5, only csv, therefore the function calls there raise a "Signature does not match base class" warning in pycharm. Could you modify the functions in the Hdf5History?

PaulJonasJost: updated

yannikschaelte (approved):

pypesto/optimize/optimizer.py
```diff
@@ -205,17 +206,25 @@ def fill_result_from_objective_history(
...
+        history = Hdf5History.load(
+            id=identifier,
+            file=history_options.storage_file,
+        )
     else:
         raise NotImplementedError()
 
     opt_hist = OptimizerHistory(
-        history, history.get_x_trace(0),
+        history, history.get_x_trace()[0],
```
PaulJonasJost, line 227:
```suggestion
        history, history.get_x_trace(0),
```

pypesto/objective/history.py
```diff
@@ -955,18 +960,18 @@ def _generate_hdf5_group(self, f: h5py.File = None):
         except OSError:
             pass
 
-    def _get_hdf5_entries(self, entry_id: str) -> Sequence:
+    def _get_hdf5_entries(self, entry_id: str,
+                          ix: Union[int, Sequence[int], None]) -> Sequence:
         """
         returns the entries for the key entry_id.
```
jvanhoefer, line 966: doc string for `Parameters`

jvanhoefer (approved):

test/base/test_history.py
```diff
@@ -141,18 +143,25 @@ def check_reconstruct_history(self, start: OptimizerResult, id: str):
...
+                                    f'get_{entry}_trace')()
+            for iteration in range(len(original_trace)):
+                # comparing nan and None difficult
+                if original_trace[iteration] is None or np.isnan(
+                        original_trace[iteration]).all():
+                    continue
+                np.testing.assert_array_almost_equal(
+                    reconst_trace[iteration],
+                    original_trace[iteration],
+                    decimal=15
```
PaulJonasJost, line 163:
```suggestion
                    decimal=10
```

pypesto/objective/history.py
PaulJonasJost:
```suggestion
        """
        Parameters
        ------------
        entry_id:
            The key whose trace is returned.
        ix:
            Index or list of indices of the iterations that will produce
            the trace.
            
        Returns
        --------
        The entries ix for the key entry_id.
```

pypesto/objective/history.py
```diff
@@ -955,18 +960,28 @@ def _generate_hdf5_group(self, f: h5py.File = None):
...
         """
-        returns the entries for the key entry_id.
+        Parameters
+        ------------
+        entry_id:
+            The key whose trace is returned.
+        ix:
+            Index or list of indices of the iterations that will produce
+            the trace.
+
```
PaulJonasJost, line 973:
```suggestion

```

test/base/test_history.py
```diff
@@ -141,18 +147,25 @@ def check_reconstruct_history(self, start: OptimizerResult, id: str):
...
+        for entry in history_entries:
+            original_trace = getattr(start.history,
+                                     f'get_{entry}_trace')()
+            reconst_trace = getattr(reconst_history,
+                                    f'get_{entry}_trace')()
+            for iteration in range(len(original_trace)):
+                # comparing nan and None difficult
+                if original_trace[iteration] is None or np.isnan(
+                        original_trace[iteration]).all():
+                    continue
```
yannikschaelte, line 163: better check that the same then also applies to the reconst_ one?

pypesto/objective/history.py
```diff
@@ -955,18 +961,27 @@ def _generate_hdf5_group(self, f: h5py.File = None):
         except OSError:
             pass
 
-    def _get_hdf5_entries(self, entry_id: str) -> Sequence:
+    def _get_hdf5_entries(self, entry_id: str,
+                          ix: Union[int, Sequence[int], None]) -> Sequence:
```
yannikschaelte, line 965:
```suggestion
    def _get_hdf5_entries(
        self,
        entry_id: str,
        ix: Union[int, Sequence[int], None] = None,
    ) -> Sequence:
```

yannikschaelte (approved):
Looks good to me, thanks! @FFroehlich as this affects some of your code, you may want to have a look, but I think we can merge already.

test/base/test_history.py
```diff
@@ -49,22 +49,28 @@ def check_history(self):
...
-                history_options=self.history_options
-            )
-
-            for istart, start in enumerate(result.optimize_result.list):
-                self.check_reconstruct_history(start, str(istart))
-                self.check_load_from_file(start, str(istart))
-                self.check_history_consistency(start)
+        for storage_type in ['.csv',
+                             '.hdf5',
+                             None]:
```
yannikschaelte, line 54:
```suggestion
        for storage_type in ['.csv', '.hdf5', None]:
```

test/base/test_history.py
PaulJonasJost:
```suggestion
                    assert(original_trace[iteration] is None or 
                        np.isnanoriginal_trace[iteration]).all()
```

PaulJonasJost: like this @yannikschaelte ?

yannikschaelte: maybe separate None and nan, then it should be like the previous check

PR #665, opened 2021-05-12
Fix axes variable overwriting; Make sure problem.x_names are unique

### LINKED ISSUE #488: Non-unique x_names crash history tracing
Should probably ensure that `pypesto.Problem.x_names` only contains unique values.

### LINKED ISSUE #502: Plotting of sampling results
Hey,

the function `sampling_parameters_trace()` defined in visualize/sampling.py has a return statement in line 192. There is just `ax` returned, meaning that you can only access information about the last subplot. I think you could simply return `axes` instead of `ax` to solve this.

Also, the function defines the argument `ax` (line 103). However, the parameter `ax` is overwritten in line 171 without using the supplied `ax`.

Best,
Tobi

yannikschaelte:
fixes #502, fixes #488

### DIFF
--- a/pypesto/problem.py
+++ b/pypesto/problem.py
@@ -142,11 +142,11 @@ class Problem:
             x_names = [f'x{j}' for j in range(0, self.dim_full)]
         if len(set(x_names)) != len(x_names):
             raise ValueError("Parameter names x_names must be unique")
-        self.x_names_full: List[str] = list(x_names)
+        self.x_names: List[str] = list(x_names)
 
         if x_scales is None:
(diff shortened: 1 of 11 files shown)


yannikschaelte:
ok, missed that x_names was formerly used in the form of x_names_full. Is it preferable to have the distinction, or should x_names refer to the full vector?

PaulJonasJost (approved):

yannikschaelte:
leaving x_names and x_scales as-are for the moment. For consistency, it may be good though to have them refer to the reduced vectors via properties in the future, tbd.

PR #666, opened 2021-05-13
Basic finite differences

### LINKED ISSUE #18: Approximate grad and hess
Approximate grad using finite differences as in the matlab version in https://github.com/ICB-DCM/PESTO/blob/master/private/getFiniteDifferences.m,

and approximate hess using any of bfgs, dfp, sr1, as in https://github.com/ICB-DCM/NOODLES/blob/feature_arc/%2Bnoodles/NoodleProblem.m.

The location for this would probably be the pesto.Objective class, which will need to interpret the input for the grad and hess arguments appropriately (e.g. string 'SR1' -> use that method), and memorize maybe some values for the adaptive finite differences scheme, and for the hessian approximations.

yannikschaelte:
Implementation of central finite differences as an objective class. Can be wrapped around existing objectives to provide finite difference approximations to derivatives (grad, hess, sres).

Not implemented yet: 
* ~~Choice of method (forward, backward, centered=only),~~
* step length adaption, 
* efficient handling of fixed parameters.
* one could save a few evaluations here and there, but as those are only O(n_par), I currently don't care ...

closes #18


yannikschaelte:
waiting for #663

pypesto/objective/aggregated.py
```diff
@@ -59,19 +60,29 @@ def __deepcopy__(self, memodict=None):
...
-    def call_unprocessed(self, x, sensi_orders, mode) -> ResultDict:
+    def call_unprocessed(
+        self,
+        x: np.ndarray,
+        sensi_orders: Tuple[int, ...],
+        mode: str,
+        **kwargs,
+    ) -> ResultDict:
         return aggregate_results([
             objective.call_unprocessed(x, sensi_orders, mode)
```
PaulJonasJost, line 87:
```suggestion
            objective.call_unprocessed(x, sensi_orders, mode, **kwargs)
```
Or remove **kwargs three lines above at is not used in this class otherwise?

pypesto/optimize/optimizer.py
```diff
@@ -206,9 +206,7 @@ def fill_result_from_objective_history(
 
 def read_result_from_file(problem: Problem, history_options: HistoryOptions,
                           identifier: str):
-    """
-    Fill a OptimizerResult from history.
-    """
+    """Fill a OptimizerResult from history."""
```
PaulJonasJost, line 209:
```suggestion
    """Fill an OptimizerResult from history."""
```

pypesto/objective/function.py
```diff
@@ -114,32 +114,6 @@ def has_res(self) -> bool:
...
-
-    def check_mode(self, mode: str) -> bool:
-        if mode == MODE_FUN and not self.has_fun:
-            return False
-
-        if mode == MODE_RES and not self.has_res:
-            return False
-
-        return True
-
```
PaulJonasJost, line 142: Why did you remove it? It is still used in e.g. the aesaraObjective.

PaulJonasJost (approved):

pypesto/objective/function.py
yannikschaelte: moved to the base class ObjectiveBase. The functionality of the has_... functions and check_sensi_orders is mutually referencing as a good default, therefore in derived classes it now suffices to replace either.

jvanhoefer:
I currently don't see an example, might be nice to have something like that? :)

yannikschaelte:
> I currently don't see an example, might be nice to have something like that? :)

One could make one, but I think that would be rather artificial at the moment and refactored anyway when we decide on how to structure the pyPESTO tutorial, therefore I would rather wait until then with an example?

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,474 @@
...
+                gm = g
+            elif self.method == FD.BACKWARD:
+                gp = g
+                gm = f_grad(x - delta)
+            else:
+                raise ValueError("Method not recognized.")
+
+            hess[:, ix] = (gp - gm) / delta_val
+        # make it symmetric
+        hess = 0.5 * (hess + hess.T)
```
paulstapor, line 429: :+1:

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,474 @@
...
+                elif self.method == FD.BACKWARD:
+                    fpp = fval
+                    fpm = f_fval(x + 0 - delta2)
+                    fmp = f_fval(x - delta1 + 0)
+                    fmm = f_fval(x - delta1 - delta2)
+                else:
+                    raise ValueError("Method not recognized.")
+
+                hess[ix1, ix2] = hess[ix2, ix1] = \
+                    (fpp - fpm - fmp + fmm) / (delta1_val * delta2_val)
```
paulstapor, line 390: There's also a particular form of 2nd order FD for the diagonal entries of the Hessian, which should be however equivalent to this here... Probably, the particular form is just a simplification, which comes out after plugging in the values from this general form here. However, Might make sense to double check. At least, I would say this here is correct...

paulstapor (approved):
Approved, looks good

paulstapor:
> > I currently don't see an example, might be nice to have something like that? :)
> 
> One could make one, but I think that would be rather artificial at the moment and refactored anyway when we decide on how to structure the pyPESTO tutorial, therefore I would rather wait until then with an example?

Would also agree to add an example. However, also fine with adding one in the tutorial. In this case, please open an issue that this won't be forgotten! :)

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,474 @@
...
+        def f_fval(x):
+            """Short-hand to get a function value."""
+            return self.obj.call_unprocessed(
+                x=x, sensi_orders=(0,), mode=MODE_FUN, **kwargs)[FVAL]
+
+        # calculate value at x only once if needed
+        if fval is None and self.method in [FD.FORWARD, FD.BACKWARD]:
+            fval = f_fval(x)
+
+        grad = np.nan * np.empty(shape=n_par)
```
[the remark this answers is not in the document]
yannikschaelte: I agree that full is the better way of initializing (was too lazy to look it up, thanks). empty works here, but e.g. not multiplying by nan leaves things in a weird state (apparently referencing the last numpy matrix calculated), so not recommended to use it.

pypesto/objective/finite_difference.py
yannikschaelte: note the diagonal is computed a few lines above

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,474 @@
...
+        If the Hessian is to be calculated via finite differences, whether
+        to employ 2nd order FDs via fval even if the objective can provide one,
+        or 1st order FDs from gradients if available.
+    delta_fun:
+        FD step sizes for gradient and Hessian.
+        Can be either a float, or a :class:`np.ndarray` of shape (n_par,)
+        for different step sizes for different coordinates.
+    delta_res:
+        FD step sizes for residual sensitivities.
+        Similar to `delta_fun`.
```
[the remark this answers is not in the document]
yannikschaelte: agreed, adding one

yannikschaelte:
> > > I currently don't see an example, might be nice to have something like that? :)
> > 
> > 
> > One could make one, but I think that would be rather artificial at the moment and refactored anyway when we decide on how to structure the pyPESTO tutorial, therefore I would rather wait until then with an example?
> 
> Would also agree to add an example. However, also fine with adding one in the tutorial. In this case, please open an issue that this won't be forgotten! :)

Added an example in the docstring now at least ...

pypesto/objective/finite_difference.py
paulstapor: ah, indeed... Sorry, seems like I missed that one...

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,508 @@
...
+                elif self.method == FD.FORWARD:
+                    fpp = f_fval(x + delta1 + delta2)
+                    fpm = f_fval(x + delta1 + 0)
+                    fmp = f_fval(x + 0 + delta2)
+                    fmm = fval
+                elif self.method == FD.BACKWARD:
+                    fpp = fval
+                    fpm = f_fval(x + 0 - delta2)
+                    fmp = f_fval(x - delta1 + 0)
+                    fmm = f_fval(x - delta1 - delta2)
```
jvanhoefer, line 405: I mean, there is the possibility to take the `f_fval(x+delta1 + delta2)` here and elsewhere (so drop the `/2` everywhere) and instead go for `(fpp - fpm - fmp + fmm) / (4 * delta1_val * delta2_val)`. This would be consistent with the diagonal elements...?

jvanhoefer: Even though now I am confused, why there is no `4*` in the diagonals...?

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,508 @@
...
+                else:
+                    raise ValueError("Method not recognized.")
```
jvanhoefer, line 407:
```suggestion
                    raise ValueError(f"Method {self.method} is currently not supported.")
```

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,508 @@
...
+
+        return hess
+
+    def _hess_via_fd_grad(self, x: np.ndarray, **kwargs) -> np.ndarray:
+        """Calculate 1st order FD approximation to Hessian via gradients."""
+        # parameter dimension
+        n_par = len(x)
+
+        def f_grad(x):
+            """Short-hand to get a function value."""
```
jvanhoefer, line 420:
```suggestion
            """Short-hand to get a gradient value."""
```

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,508 @@
...
+                gp = f_grad(x + delta / 2)
+                gm = f_grad(x - delta / 2)
+            elif self.method == FD.FORWARD:
+                gp = f_grad(x + delta)
+                gm = g
+            elif self.method == FD.BACKWARD:
+                gp = g
+                gm = f_grad(x - delta)
+            else:
+                raise ValueError("Method not recognized.")
```
jvanhoefer, line 445:
```suggestion
                raise ValueError(f"Method {self.method} is currently not supported.")
```

jvanhoefer (approved):
Looks fine, only minor comments :) (I am a bit curious about the "inconsistency" in the hessian between mixed and non-mixed 2nd derivatives...)

pypesto/objective/base.py
```diff
@@ -218,23 +242,23 @@ def check_sensi_orders(
...
+                0 in sensi_orders and not self.has_fun
+                or 1 in sensi_orders and not self.has_grad
+                or 2 in sensi_orders and not self.has_hess
+            )
+        ) or (
+            mode == MODE_RES
+            and (
+                0 in sensi_orders and not self.has_res
+                or 1 in sensi_orders and not self.has_sres
+            )
```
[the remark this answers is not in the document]
yannikschaelte: not aware of any optimizer that does as most stick to the linear structure ... but could of course be added then.

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,474 @@
...
+        self.hess: Union[bool, None] = hess
+        self.sres: Union[bool, None] = sres
+        self.hess_via_fval: bool = hess_via_fval
+        self.delta_fun: Union[float, np.ndarray] = delta_fun
+        self.delta_res: Union[float, np.ndarray] = delta_res
+        self.method: str = method
+
+        if isinstance(delta_fun, str) or isinstance(delta_res, str):
+            raise NotImplementedError(
+                "Adaptive FD step sizes are not implemented yet.",
```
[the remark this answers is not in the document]
yannikschaelte: haha it is recommended because it allows modifying less lines when e.g. adding additional arguments. true, adding one below.

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,474 @@
...
+    relies on numerical approximations, such as an ODE.
+
+    Parameters
+    ----------
+    grad:
+        Derivative method for the gradient.
+    hess:
+        Derivative method for the Hessian
+    sres:
+        Derivative method for the residual sensitivities.
```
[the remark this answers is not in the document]
yannikschaelte: {True, False, None} here is a ternary variable: None means whatever is available, True means always FD, False never. Documented that a few lines above, will add hints.

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,474 @@
...
+
+    def call_unprocessed(
+        self,
+        x: np.ndarray,
+        sensi_orders: Tuple[int, ...],
+        mode: str,
+        **kwargs,
+    ) -> ResultDict:
+        # This is the main method to overwrite from the base class, it handles
+        #  and delegates the actual objective evaluation.
```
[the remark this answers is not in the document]
yannikschaelte: would rather not use docstring here, as the docstring should be the one from the base class, this here is just a hint to the programmer.

pypesto/objective/finite_difference.py
```diff
@@ -0,0 +1,474 @@
...
+                fc = fval
+                f2m = f_fval(x - delta1)
+            elif self.method == FD.FORWARD:
+                f2p = f_fval(x + 2 * delta1)
+                fc = f_fval(x + delta1)
+                f2m = fval
+            elif self.method == FD.BACKWARD:
+                f2p = fval
+                fc = f_fval(x - delta1)
+                f2m = f_fval(x - 2 * delta1)
```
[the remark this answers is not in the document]
yannikschaelte: it is (f'(x) - f'(x-d)) / d \approx ( (f(x) - f(x-d)) / d - (f(x-d) - f(x-2d)) / d) / d, so yes should be correct.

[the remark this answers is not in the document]
yannikschaelte: it is (f'(x) - f'(x-d)) / d \approx ( (f(x) - f(x-d)) / d - (f(x-d) - f(x-2d)) / d) / d, so yes should be correct.

pypesto/objective/finite_difference.py
yannikschaelte: the step size should always be delta, i.e. from x - delta/2 to x + delta/2 (central), or from x to x + delta (forward). I think that should be consistent ...

yannikschaelte: I think in the matlab implementation, it used a stepsize of delta for the off-diagonals, and delta/2 on the diagonals, which may be not the most consistent ... even though one would save a few evaluations when simultaneously computing gradients.

pypesto/objective/base.py
```diff
@@ -218,23 +241,23 @@ def check_sensi_orders(
...
-        ----------
-        mode:
-            Whether to compute function values or residuals.
-        Returns
-        -------
-        flag:
-            Boolean indicating whether mode is supported
-        """
-        raise NotImplementedError()
+        if (
```
PaulJonasJost, line 244: I was wondering about that in another PR but shouldn't we implement something that checks that the `sensi_orders` are even valid? right now if we would call `check_sensi_orders(sensi_orders=(3,))` it would return `True`.  How about adding an `if(max(sensi_orders)>2): raise ValueError`?

yannikschaelte: would make sense

PR #667, opened 2021-05-14
pypesto screen logger to info by default.

### LINKED ISSUE #367: pypesto log level
Often, we have useful information to print on screen. Currently, all is done via the `logging` module. The default is to only show `warning` and `error`, `info` is not shown. I would rather not print "information" on "warning" level. One could in pypesto centrally enable showing `info` as well. Then the level `debug` would remain for everything that is only to be shown when e.g. searching for bugs. Opinions?

Alternatively, one could employ `print` for info, but that I would rather not do, as this has a bunch of disadvantages.

PaulJonasJost:
I think that is what was requested in #367.


yannikschaelte:
It is good to set the level to INFO here, but this does not solve the issue yet, as this function is never called. To by default always log everything from INFO upwards, one would need to call the function once in module import. This can be done similarly to https://github.com/ICB-DCM/pyABC/blob/main/pyabc/__init__.py#L120 (with ABC -> pypesto).

PaulJonasJost:
that should do the trick @yannikschaelte?

pypesto/logging.py
```diff
@@ -31,7 +31,7 @@ def log(name: str = 'pypesto',
         If specified, messages are logged to a file with this name.
     """
     logger = logging.getLogger(name)
-    logger.setLevel(level)
+    logging.basicConfig(level=level)
```
yannikschaelte, line 34: it shouldn't be necessary to call basicConfig, as this messes with global debuggers ...

pypesto/__init__.py
```diff
@@ -36,3 +36,5 @@
 from . import startpoint
 from . import store
 from . import visualize
+
+logging.log(name=__name__)
```
yannikschaelte, line 40:
```suggestion
logging.log()
```
should suffice

pypesto/logging.py
PaulJonasJost: well i am not too sure but when i kept it at `logger.setLevel(level)` the level was set to info (so everything should have been okay there) but the info messages still were not displayed...

PaulJonasJost: Found the error.

yannikschaelte (approved):

PR #669, opened 2021-05-16
Storage of objective representation to hdf5

### LINKED ISSUE #350: Result: Do not store objective
The objective function itself cannot be save (except serialization, which makes no sense), so it should not be part of the result. I.e. not the full problem should be part of the result. What's needed from the problem are anyway only the vectors, e.g. parameter names and fixed parameters.

Suggesting to save a "ProblemInfo" (or a nicer name) to the result, which contains all vectors, but instead of the objective just a string representation.

### LINKED ISSUE #573: Problems with ProblemHDF5Reader/Writer
I noticed two problem with the saving and reading of the problem in and from hdf5.

1.  If you have x_guesses in the original problem, they do not get transported into the problem you read from hdf5
2. the objective function seems to not work after reading it from hdf5

i included a minimal example:
``` python
"""This file serves as an error example"""
import pypesto
import numpy as np
import scipy as sp
import pypesto.optimize as optimize
from pypesto.store import (ProblemHDF5Reader,
                           ProblemHDF5Writer)


def setup_rosen_problem(n_starts: int = 2):
    """Set up the Rosenbrock problem and return
    a pypesto.Problem"""
    objective = pypesto.Objective(fun=sp.optimize.rosen,
                                  grad=sp.optimize.rosen_der,
                                  hess=sp.optimize.rosen_hess)

    dim_full = 10
    lb = -5 * np.ones((dim_full, 1))
    ub = 5 * np.ones((dim_full, 1))

    # fixing startpoints
    startpoints = pypesto.startpoint.latin_hypercube(n_starts=n_starts,
                                                     lb=lb,
                                                     ub=ub)
    problem = pypesto.Problem(objective=objective, lb=lb, ub=ub,
                              x_guesses=startpoints)
    return problem

file_name = 'test'
problem = setup_rosen_problem()
problem_writer = ProblemHDF5Writer(file_name)
problem_writer.write(problem)
problem_reader = ProblemHDF5Reader(file_name)
problem_read = problem_reader.read()

# error one: No passing of startpoints
print(problem_read.x_guesses)
# error two: objective function does not work
result = optimize.minimize(problem=problem_read, n_starts=2)
```

PaulJonasJost:
Closes #350, closes #573


pypesto/objective/__init__.py
```diff
@@ -10,6 +10,7 @@
 from .priors import NegLogPriors, NegLogParameterPriors
 from .util import res_to_chi2, sres_to_schi2
 from .base import ObjectiveBase
+from .aesara import AesaraObjective
```
yannikschaelte, line 13: I think this is not desirable, as it requires an installation of aesara

PaulJonasJost: ok, should i then remove it as well from read/write? could also do a `try...except ImportError`.

pypesto/problem.py
```diff
@@ -128,8 +128,6 @@ def __init__(self,
             for idx, x in enumerate(x_fixed_vals)
         ]
 
-        self._x_free_indices: Union[List[int], None] = None
```
yannikschaelte, line 131: :+1:

pypesto/objective/history.py
```diff
@@ -778,7 +778,7 @@ def __init__(self,
         self._generate_hdf5_group()
 
     def __len__(self):
-        with h5py.File(self.file, 'a') as f:
+        with h5py.File(self.file, 'r') as f:
```
yannikschaelte, line 781: :+1:

pypesto/store/hdf5.py
```diff
@@ -29,6 +29,8 @@ def write_array(f: h5py.Group,
     elif all(isinstance(x, str) for x in values):
         write_string_array(f, path,
                            values)
+    else:
```
yannikschaelte, line 32: Not sure, but this looks potentially problematic. @jvanhoefer @plakrisenko @MerktSimon What was the reason to explicitly use the "write_x_array" functions?

pypesto/store/read_from_hdf5.py
```diff
@@ -111,13 +111,24 @@ def read(self, objective: ObjectiveBase = None) -> Problem:
         problem:
             A problem instance with all attributes read in.
         """
+        # raise warning that problem is not to be used but
+        # only stores information
+        logger.warning('WARNING: You are loading a problem.\nThis problem'
+                       ' is not to be used without a separatly created'
+                       ' objective.\nIn place of problem.objective you can'
+                       ' access the Information\nstored on the objective.\n')
```
yannikschaelte, line 119: don't get this sentence ...

pypesto/store/read_from_hdf5.py
```diff
@@ -111,13 +111,24 @@ def read(self, objective: ObjectiveBase = None) -> Problem:
         problem:
             A problem instance with all attributes read in.
         """
+        # raise warning that problem is not to be used but
+        # only stores information
+        logger.warning('WARNING: You are loading a problem.\nThis problem'
+                       ' is not to be used without a separatly created'
```
yannikschaelte, line 117:
```suggestion
                       ' is not to be used without a separately created'
```

pypesto/store/read_from_hdf5.py
```diff
@@ -111,13 +111,24 @@ def read(self, objective: ObjectiveBase = None) -> Problem:
         problem:
             A problem instance with all attributes read in.
         """
+        # raise warning that problem is not to be used but
+        # only stores information
+        logger.warning('WARNING: You are loading a problem.\nThis problem'
```
yannikschaelte, line 116: I think a warning is a bit harsh here. Documenting in the docstring should be enough. Potentially warn if no objective is passed.

pypesto/store/read_from_hdf5.py
```diff
@@ -111,13 +111,24 @@ def read(self, objective: ObjectiveBase = None) -> Problem:
...
+                       ' access the Information\nstored on the objective.\n')
+
         # create empty problem
         if objective is None:
             objective = Objective()
         problem = Problem(objective, [], [])
 
         with h5py.File(self.storage_filename, 'r') as f:
             for problem_key in f['/problem']:
+                if problem_key == 'objective_info':
```
yannikschaelte, line 128: I would rather not at all read the "objective info" into the problem.objective, but rather offer a different function reader.get_objective_info() for the moment. The problem.objective should only hold a proper objective.

pypesto/store/read_from_hdf5.py
```diff
@@ -331,3 +342,22 @@ def read_result(filename: str,
                            f'within {filename}.')
 
     return result
+
+
+def load_objective_info(f: h5py.Group):
+    """
+    Loads the objective Information stored in f
```
yannikschaelte, line 349:
```suggestion
    """Load the objective information stored in f.
```

pypesto/store/save_to_hdf5.py
```diff
@@ -315,3 +320,110 @@ def write_result(result: Result,
...
+
+
+def save_objective_infos(
+        objective: Union[AmiciObjective,
+                         AggregatedObjective,
+                         Objective],
+        f: h5py.File
+):
+    """
+    Saves the Objective Infos in a Dictonary.
```
yannikschaelte, line 332:
```suggestion
    """Save the Objective Infos in a Dictonary.
```
infinitive/imperative

pypesto/store/save_to_hdf5.py
```diff
@@ -315,3 +320,110 @@ def write_result(result: Result,
...
+        The objective whose information is to be stored.
+    f:
+        The h5py.File in which the infos are stored.
+    """
+    if isinstance(objective, AmiciObjective):
+        save_amici_objective_info(objective, f)
+    elif isinstance(objective, AggregatedObjective):
+        save_aggregated_objective_info(objective, f)
+    else:
+        save_function_objective_info(objective, f)
```
yannikschaelte, line 346: if ... else statements are generally bad design in such a context, as they are inflexible to extend. rather, the objective function should have a method to return such information, see e.g. https://github.com/ICB-DCM/pyABC/blob/main/pyabc/distance/distance.py#L106. this function should then be extended/modified in subclasses.

yannikschaelte (changes requested):

pypesto/objective/base.py
```diff
@@ -241,6 +241,8 @@ def check_sensi_orders(
             Boolean indicating whether combination of sensi_orders and mode
             is supported
         """
+        if(max(sensi_orders) > 2):
```
yannikschaelte, line 244: if checking this, then better also for fun and res separately (the latter can only <=1)

pypesto/objective/aggregated.py
```diff
@@ -92,6 +92,13 @@ def initialize(self):
         for objective in self._objectives:
             objective.initialize()
 
+    def get_config(self):
+        info = {}
+        info['type'] = str(type(self))
```
yannikschaelte, line 97: rather use `self.__class__.__name__` instead of `type`

pypesto/objective/aggregated.py
```diff
@@ -92,6 +92,13 @@ def initialize(self):
         for objective in self._objectives:
             objective.initialize()
 
+    def get_config(self):
```
yannikschaelte, line 95:
```suggestion
    def get_config(self) -> dict:
```

pypesto/objective/amici.py
```diff
@@ -197,6 +197,17 @@ def __init__(
         # `set_custom_timepoints` method for more information.
         self.custom_timepoints = None
 
+    def get_config(self):
```
yannikschaelte, line 200: the `get_config` method should also exist in the base class with docstring and a basic implementation, which derived classes can then choose to override.

pypesto/objective/function.py
```diff
@@ -114,6 +114,17 @@ def has_res(self) -> bool:
     def has_sres(self) -> bool:
         return callable(self.sres) or self.sres is True
 
+    def get_config(self):
+        info = {}
```
yannikschaelte, line 118: it makes sense to always start by `config = super().get_config()` and then extend it.

pypesto/store/hdf5.py
```diff
@@ -96,3 +98,35 @@ def write_int_array(f: h5py.Group,
     """
     dset = f.create_dataset(path, (len(values),), dtype=dtype)
     dset[:] = values
+
+
+def write_dict(
```
yannikschaelte, line 103: I actually think it would be much easier to have that config just be a str, then one has absolutely no trouble of storing it ... what do you think?

pypesto/store/read_from_hdf5.py
```diff
@@ -111,13 +111,21 @@ def read(self, objective: ObjectiveBase = None) -> Problem:
         problem:
             A problem instance with all attributes read in.
         """
+        # raise warning that problem is not to be used but
```
yannikschaelte, line 114: only makes sense to say something if objective is None.

yannikschaelte (changes requested):

pypesto/store/hdf5.py
PaulJonasJost: you mean instead of 
```
dict = {'key1': value1, 'key2' : value2}
```
rather 
```
dict = "key1: value1\nkey2: value2"
```
?

PaulJonasJost: could perhaps be a bit "problematic" with the aggregated objective but yes in principle that should work.

yannikschaelte: you could also have them here always return dicts (either to append to), and just convert to string before saving.

pypesto/objective/base.py
PaulJonasJost: :+1:

pypesto/store/read_from_hdf5.py
PaulJonasJost: why only then? If we have e.g. an amici objective, we cannot load this by loading the problem from hdf5 don't we?

yannikschaelte: sorry, I don't get it :sweat_smile:

PaulJonasJost: Well the info/warning was meant to tell the user that he cannot use the problem loaded from the hdf5 file for optimization etc. The reason for this is (as far as I understand that is the only reason) that the objective is not loaded correctly (and is also not intended to be loaded). Therefore the warning says to separately create the objective function (or the problem in general) if one aims to do optimization.

yannikschaelte: yes, but `if objective is not None`, then everything should be fine, right? As then the user has provided an objective.

But there may be a problem as this objective is likely not accounting for fixed parameters yet, therefore maybe one should call problem.objective.update_from_problem() afterwards ...

PaulJonasJost: ahhh yeah I think now I get what you meant :D you are right 👍

pypesto/store/read_from_hdf5.py
```diff
@@ -331,3 +339,23 @@ def read_result(filename: str,
                            f'within {filename}.')
 
     return result
+
+
+def load_objective_config(filename: str):
+    """
+    Load the objective Information stored in f
```
yannikschaelte, line 346:
```suggestion
    Load the objective information stored in f
```

yannikschaelte (approved):
looks good to me!

PR #671, opened 2021-05-18
Implement FD step size adaptation

yannikschaelte:
Step sizes adaptation for finite differences, realized via a `pypesto.objective.finite_difference.FDDelta` class.

* "distance": Update when sufficiently far away from last update point (as in PESTO)
* "steps": Update when gone a number of steps.
* "constant": Only update initially.

Also, some simplifications of grad+hess+sres derivative calculations (i.e. rather interpret those as 1st or 2nd order derivatives with code reuse).

Reference: https://github.com/ICB-DCM/PESTO/blob/master/private/getStepSizeFD.m


pypesto/objective/finite_difference.py
```diff
@@ -1,13 +1,215 @@
...
+    max_steps:
+        Number of steps after which to update in the `Delta.STEPS` update
+        condition.
+    """
+
+    # update conditions
+    CONSTANT = "constant"
+    SPACE = "space"
+    STEPS = "steps"
+    UPDATE_CONDITIONS = [CONSTANT, SPACE, STEPS]
```
paulstapor, line 48: -> constants.py

pypesto/objective/finite_difference.py
```diff
@@ -1,13 +1,215 @@
...
+    update_condition:
+        A "good" step size may be a local property. Thus, this class allows
+        updating the step size if certain criteria are met, in the
+        :func:`pypesto.objective.finite_difference.Delta.update`
+        function.
+        Delta.CONSTANT means that the step size is only initially selected.
+        Delta.SPACE means that the step size is updated if the current
+        evaluation point is sufficiently far away from the last training point.
+        Delta.STEPS means that the step size is updated `max_steps` evaluations
+        after the last update.
```
paulstapor: If `STEPS` are chosen, the `max_steps` is used to assess, whether or not Delta is recomputed. For `SPACE`, one would need an input such as `max_dist`.

pypesto/objective/finite_difference.py
```diff
@@ -1,13 +1,215 @@
...
+        test_deltas: np.ndarray = None,
+        update_condition: str = CONSTANT,
+        max_steps: int = 30,
+    ):
+        if not isinstance(delta, (np.ndarray, float)) and delta is not None:
+            raise ValueError(f"Unexpected type {type(delta)} for delta")
+        self.delta: Union[np.ndarray, float, None] = delta
+
+        if test_deltas is None:
+            test_deltas = np.array([10**(-i) for i in range(2, 9)])
```
paulstapor, line 62: Maybe... one might want to include also -1 here, i.e., `range(1, 9)` rather than `range(2,9)`. If you don't want to increase cpu time, rather kick out the -8, i.e., go for `range(1,8)`

pypesto/objective/finite_difference.py
```diff
@@ -1,13 +1,215 @@
...
+    UPDATE_CONDITIONS = [CONSTANT, SPACE, STEPS]
+
+    def __init__(
+        self,
+        delta: Union[np.ndarray, float, None] = None,
+        test_deltas: np.ndarray = None,
+        update_condition: str = CONSTANT,
+        max_steps: int = 30,
+    ):
+        if not isinstance(delta, (np.ndarray, float)) and delta is not None:
```
PaulJonasJost, line 57: as they delta vector should only ever be of size `(n_par,)` would be good to add a check for this as well?

test/base/test_objective.py
```diff
@@ -251,10 +269,14 @@ def test_fds(fd_method):
         val_fd_grad = getattr(obj_fd_grad, f"get_{attr}")(p)
         val_fd_fake = getattr(obj_fd_fake, f"get_{attr}")(p)
 
-        for v in [val_fd, val_fd_grad]:
-            assert np.allclose(val, v, atol=atol, rtol=rtol), attr
-            # cannot completely coincide
-            assert (val != v).any(), attr
+        assert np.allclose(val, val_fd, atol=atol, rtol=rtol), attr
```
paulstapor, line 272: Can we maybe double check, whether we achieve better accuracies when using adaptive FDs than when using static FDs? Maybe even down to 1e-5?

pypesto/objective/finite_difference.py
```diff
@@ -1,13 +1,215 @@
...
+        if isinstance(self.delta, float):
+            self.delta: np.ndarray = self.delta * np.ones(shape=x.shape)
+
+        self.steps += 1
+
+        # return if no update needed
+        if self.delta is not None:
+            if (
+                self.update_condition == FDDelta.SPACE and
+                np.sum((x - self.x0)**2) <= 0.5 * np.sqrt(len(x))
```
paulstapor, line 112: Would be nice, if this 0.5 was a hyperparameter.

pypesto/objective/finite_difference.py
```diff
@@ -1,13 +1,215 @@
...
+            ):
+                return
+            if (
+                self.update_condition == FDDelta.STEPS and
```
PaulJonasJost, line 116: I might be wrong, but as `self.steps==0` initially, update condition `STEPS` would update the deltas at the first update making the initial deltas obsolete?

pypesto/objective/finite_difference.py
```diff
@@ -1,13 +1,215 @@
...
+        #  entries towards smaller and larger deltas, thus indicating the
+        #  change in the approximation when changing delta.
+        # This is done separately for each parameter. Then, for each the delta
+        #  with the minimal entry and thus the most stable behavior
+        #  is selected.
+        stab_vec = np.full(shape=nablas.shape, fill_value=np.nan)
+        stab_vec[1:-1] = 0.5 * (
+            np.abs(nablas[2:] - nablas[1:-1]) +
+            np.abs(nablas[1:-1] - nablas[:-2])
+        )
```
paulstapor, line 171:
```
stab_vec[1:-1] = np.mean(np.abs(
    [nablas[2:] - nablas[1:-1], 
     nablas[1:-1] - nablas[:-2]]
))
```
?

paulstapor: Should hopefully do the same, and makes it clearer what actually happens

paulstapor: alternatively, you add a nan-vector in the first and last position and just use np.nanmean over the whole thing, whic avoids handling the corner cases separately

pypesto/objective/finite_difference.py
```diff
@@ -1,13 +1,215 @@
...
+        # on the edge, just take the single neighbor
+        stab_vec[0] = np.abs(nablas[1] - nablas[0])
+        stab_vec[-1] = np.abs(nablas[-1] - nablas[-2])
+
+        # if the function is tensor-valued, consider the mean over all entries
+        # there may be more intelligent ways
+        if stab_vec.ndim > 2:
+            # flatten all dimensions > 1
+            stab_vec = stab_vec.reshape(
+                stab_vec.shape[0], stab_vec.shape[1], -1).mean(axis=2)
```
paulstapor, line 181: In what situation would that happen? If we do the Hessian? ... Maybe then a Max-Norm rather than the mean would be better, right? We want to constrain the worst deviation, not the mean of it over n parameters, right?

pypesto/objective/finite_difference.py
```diff
@@ -1,13 +1,215 @@
...
+
+        # minimum delta index for each parameter
+        min_ixs = np.argmin(stab_vec, axis=0)
+
+        # extract optimal deltas per parameter
+        delta_opt = np.array([self.test_deltas[ix] for ix in min_ixs])
+        self.delta = delta_opt
```
paulstapor, line 188: Personally, I'm a fan of having things explicit and not craming everything into one line or LC. However, one could think of writing `self.delta = np.array([self.test_deltas[ix] for ix in min_ixs])` ... ;)

PaulJonasJost (approved):

paulstapor (approved):
Overall good, some minor stuff left

pypesto/objective/finite_difference.py
yannikschaelte: I'd rather leave these here, s.t. one has to call them `pypesto.FDDelta.CONSTANT`, as they are only used in this specific context.

pypesto/objective/finite_difference.py
yannikschaelte: true, added "and self.steps > 1"

pypesto/objective/finite_difference.py
yannikschaelte: exactly, if it is the hessian or sres. Agreed, max may make more sense here. changed .mean to .max

test/base/test_objective.py
yannikschaelte: The default is 1e-6, which already gives quite a good approximation (automatically chosen are 1e-7 to 1e-8), so on this problem not much gain, rather a "automatic detection does not screw things up" check.

PR #678, opened 2021-05-27
add petab linter to petab import closes #677

### LINKED ISSUE #677: Run PEtab linter in PEtab import
currently, the PEtab linter is not run when running the PEtab import. This can lead to an incorrect PEtab model being imported, which then causes (potentially non-informative) problems downstream, which might be especially an issue for new users...

jvanhoefer:


yannikschaelte (approved):
makes sense

PR #679, opened 2021-05-28
Read-only HDF5 file access for Hdf5History.__len__

dweindl:
That function doesn't need write access to the history file and unnecessarily prevents parallel access.
Changing to read-only.

### DIFF
--- a/pypesto/objective/history.py
+++ b/pypesto/objective/history.py
@@ -778,7 +778,7 @@ class Hdf5History(History):
         self._generate_hdf5_group()
 
     def __len__(self):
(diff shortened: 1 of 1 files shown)


PaulJonasJost (approved):
Yes sorry that is on me. Thanks for the fix 👍

yannikschaelte (approved):

yannikschaelte:
@PaulJonasJost had fixed it swhere else already, but twice is better

PaulJonasJost:
> @PaulJonasJost had fixed it swhere else already, but twice is better

Oh yes you are right :D but the PR is still open.

PR #688, opened 2021-06-30
Fixed ensemble from optimization

PaulJonasJost:
Noticed a mistake in the implementation, as `x_names, ub, lb` where from the full parameter vector, while the vectors from optimization only contain the non fixed. Adjusted it so that it is consistent with the optimization parameters.


dilpath (approved):
:+1:

PR #690, opened 2021-07-01
Adding a helper method to check model gradients before optimization.

stephanmg:
Check for the gradients of the model equations and returns `False` if gradients are incompatible / not well-suited for gradient-based optimizers otherwise returns `True`.

### DIFF
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -107,12 +107,8 @@ class PetabImporter(AmiciObjectBuilder):
             except (RuntimeError, ValueError):
                 dfs.append(dfs[0])
 
-        return np.all((dfs[0].rel_err.values < RTOL) |
-                      (dfs[0].abs_err.values < ATOL) |
-                      (dfs[1].rel_err.values < RTOL) |
-                      (dfs[1].abs_err.values < ATOL) |
-                      (dfs[2].rel_err.values < RTOL) |
-                      (dfs[2].abs_err.values < ATOL))
+        return all([np.all(df.rel_err.values < RTOL |
+                   df.abs_err.values < ATOL) for df in dfs])
 
     def create_model(self,
                      force_compile: bool = False,
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -87,18 +87,19 @@ class PetabImporter(AmiciObjectBuilder):
         rtol: float = 1e-2,
         atol: float = 1e-3,
         mode: Literal = None,
-        objective: Objective = None,
-        objAbsoluteTolerance: float = 1e-10,
-        objRelativeTolerance: float = 1e-12,
+        objective: Optional[Objective] = None,
+        obj_absolute_tolerance: float = 1e-10,
+        obj_relative_tolerance: float = 1e-12,
         multi_eps=None,
-        **kwargs
-    ):
+        **kwargs,
+    ) -> bool:
         """Check if gradients match finite differences (FDs)
+
         Parameters
         ----------
         rtol: relative error tolerance
         atol: absolute error tolerance
-        mode: unction values or residuals
+        mode: function values or residuals
         objective: objective function
         objAbsoluteTolerance: absolute tolerance in sensitivity calculation
         objRelativeTolerance: relative tolerance in sensitivity calculation
@@ -116,8 +117,8 @@ class PetabImporter(AmiciObjectBuilder):
             objective = problem.objective
             objective.amici_solver.setSensitivityMethod(
                 amici.SensitivityMethod_forward)
-            objective.amici_solver.setAbsoluteTolerance(objAbsoluteTolerance)
-            objective.amici_solver.setRelativeTolerance(objRelativeTolerance)
+            objective.amici_solver.setAbsoluteTolerance(obj_absolute_tolerance)
+            objective.amici_solver.setRelativeTolerance(obj_relative_tolerance)
 
         free_indices = par[problem.x_free_indices]
         dfs = []
@@ -129,7 +130,7 @@ class PetabImporter(AmiciObjectBuilder):
             modes = [mode]
 
         if multi_eps is None:
-            multi_eps = [1e-3, 1e-4, 1e-5]
+            multi_eps = np.array([10**(-i) for i in range(3, 9)])
 
         for mode in modes:
             try:
@@ -139,12 +140,12 @@ class PetabImporter(AmiciObjectBuilder):
             except (RuntimeError, ValueError):
                 # Might happen in case PEtab problem not well defined or
                 # fails for specified tolerances in forward sensitivities
-                pass
+                return False
 
-        return any([
+        return all([
             any([
-                np.all(mode_df.rel_err.values < rtol),
-                np.all(mode_df.abs_err.values < atol),
+                np.all((mode_df.rel_err.values < rtol) |
+                       (mode_df.abs_err.values < atol))
             ])
             for mode_df in dfs
         ])
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -110,9 +110,7 @@ class PetabImporter(AmiciObjectBuilder):
         return np.all((dfs[0].rel_err.values < RTOL) |
                       (dfs[0].abs_err.values < ATOL) |
                       (dfs[1].rel_err.values < RTOL) |
-                      (dfs[1].abs_err.values < ATOL) |
-                      (dfs[2].rel_err.values < RTOL) |
-                      (dfs[2].abs_err.values < ATOL))
+                      (dfs[1].abs_err.values < ATOL))
 
     def create_model(self,
                      force_compile: bool = False,
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -90,6 +90,7 @@ class PetabImporter(AmiciObjectBuilder):
         objective: Objective = None,
         objAbsoluteTolerance: float = 1e-10,
         objRelativeTolerance: float = 1e-12,
+        multi_eps=None,
         **kwargs
     ):
         """Check if gradients match finite differences (FDs)
@@ -101,6 +102,7 @@ class PetabImporter(AmiciObjectBuilder):
         objective: objective function
         objAbsoluteTolerance: absolute tolerance in sensitivity calculation
         objRelativeTolerance: relative tolerance in sensitivity calculation
+        multi_eps: multiple test step width for FDs
 
         Returns
         -------
@@ -120,16 +122,20 @@ class PetabImporter(AmiciObjectBuilder):
         free_indices = par[problem.x_free_indices]
         dfs = []
         modes = []
+
         if mode is None:
             modes = [MODE_FUN, MODE_RES]
         else:
             modes = [mode]
 
+        if multi_eps is None:
+            multi_eps = [1e-3, 1e-4, 1e-5]
+
         for mode in modes:
             try:
                 dfs.append(objective.check_grad_multi_eps(
                             free_indices, *args, **kwargs,
-                            mode=mode))
+                            mode=mode, multi_eps=multi_eps))
             except (RuntimeError, ValueError):
                 # Might happen in case PEtab problem not well defined or
                 # fails for specified tolerances in forward sensitivities
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -133,7 +133,7 @@ class PetabImporter(AmiciObjectBuilder):
         return all([
             any([
                 np.all((mode_df.rel_err.values < rtol) |
-                       (mode_df.abs_err.values < atol))
+                       (mode_df.abs_err.values < atol)),
             ])
             for mode_df in dfs
         ])
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -1,4 +1,3 @@
-from pypesto.objective.constants import MODE_FUN, MODE_RES
 import pandas as pd
 import numpy as np
 import os
@@ -7,7 +6,7 @@ import importlib
 import shutil
 import logging
 import tempfile
-from typing import Iterable, List, Optional, Sequence, Union, Callable
+from typing import Iterable, List, Literal, Optional, Sequence, Union, Callable
 
 from ..problem import Problem
 from ..objective import AmiciObjective, AmiciObjectBuilder, AggregatedObjective
@@ -15,6 +14,8 @@ from ..predict import AmiciPredictor, PredictionResult
 from ..predict.constants import CONDITION_SEP
 from ..objective.priors import NegLogParameterPriors, \
     get_parameter_prior_dict
+from ..objective.function import Objective
+from ..objective.constants import MODE_FUN, MODE_RES
 
 try:
     import petab
@@ -80,37 +81,67 @@ class PetabImporter(AmiciObjectBuilder):
             output_folder=output_folder,
             model_name=model_name)
 
-    def check_and_log_gradients(self, RTOL: float = 1e-2, ATOL: float = 1e-3):
-        """Check and log if gradients are compatible with gradient-based optimizers
+    def check_gradients(
+        self,
+        *args,
+        rtol: float = 1e-2,
+        atol: float = 1e-3,
+        mode: Literal = None,
+        objective: Objective = None,
+        objAbsoluteTolerance: float = 1e-10,
+        objRelativeTolerance: float = 1e-12,
+        **kwargs
+    ):
+        """Check if gradients match finite differences (FDs)
         Parameters
         ----------
-        RTOL: relative error tolerance (default 1e-2)
-        ATOL: absolute error tolerance (default 1e-3)
+        rtol: relative error tolerance
+        atol: absolute error tolerance
+        mode: unction values or residuals
+        objective: objective function
+        objAbsoluteTolerance: absolute tolerance in sensitivity calculation
+        objRelativeTolerance: relative tolerance in sensitivity calculation
+
+        Returns
+        -------
+        bool
+            Indicates whether gradients match (True) FDs or not (False)
         """
         par = np.asarray(self.petab_problem.x_nominal_scaled)
         problem = self.create_problem()
-        objective = problem.objective
-        objective.amici_solver.setSensitivityMethod(
-            amici.SensitivityMethod_forward)
-        objective.amici_solver.setAbsoluteTolerance(1e-10)
-        objective.amici_solver.setRelativeTolerance(1e-12)
+
+        if objective is None:
+            objective = problem.objective
+            objective.amici_solver.setSensitivityMethod(
+                amici.SensitivityMethod_forward)
+            objective.amici_solver.setAbsoluteTolerance(objAbsoluteTolerance)
+            objective.amici_solver.setRelativeTolerance(objRelativeTolerance)
+
         free_indices = par[problem.x_free_indices]
         dfs = []
+        modes = []
+        if mode is None:
+            modes = [MODE_FUN, MODE_RES]
+        else:
+            modes = [mode]
 
-        modes = [MODE_FUN, MODE_RES]
         for mode in modes:
             try:
                 dfs.append(objective.check_grad_multi_eps(
-                            free_indices,
-                            multi_eps=[1e-3, 1e-4, 1e-5],
+                            free_indices, *args, **kwargs,
                             mode=mode))
             except (RuntimeError, ValueError):
-                dfs.append(dfs[0])
+                # Might happen in case PEtab problem not well defined or
+                # fails for specified tolerances in forward sensitivities
+                pass
 
-        return np.all((dfs[0].rel_err.values < RTOL) |
-                      (dfs[0].abs_err.values < ATOL) |
-                      (dfs[1].rel_err.values < RTOL) |
-                      (dfs[1].abs_err.values < ATOL))
+        return any([
+            any([
+                np.all(mode_df.rel_err.values < rtol),
+                np.all(mode_df.abs_err.values < atol),
+            ])
+            for mode_df in dfs
+        ])
 
     def create_model(self,
                      force_compile: bool = False,
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -107,8 +107,12 @@ class PetabImporter(AmiciObjectBuilder):
             except (RuntimeError, ValueError):
                 dfs.append(dfs[0])
 
-        return all([np.all(df.rel_err.values < RTOL |
-                   df.abs_err.values < ATOL) for df in dfs])
+        return np.all((dfs[0].rel_err.values < RTOL) |
+                      (dfs[0].abs_err.values < ATOL) |
+                      (dfs[1].rel_err.values < RTOL) |
+                      (dfs[1].abs_err.values < ATOL) |
+                      (dfs[2].rel_err.values < RTOL) |
+                      (dfs[2].abs_err.values < ATOL))
 
     def create_model(self,
                      force_compile: bool = False,
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -1,4 +1,6 @@
+from pypesto.objective.constants import MODE_FUN, MODE_RES
 import pandas as pd
+import numpy as np
 import os
 import sys
 import importlib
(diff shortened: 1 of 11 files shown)


pypesto/petab/importer.py
```diff
@@ -78,6 +80,40 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        dfs.append(objective.check_grad_multi_eps(
+                    free_indices,
+                    multi_eps=[1e-3, 1e-4, 1e-5]))
+
+        modes = [MODE_FUN, MODE_RES]
+        for mode in modes:
+            try:
+                dfs.append(objective.check_grad(free_indices, mode=mode))
+            except (RuntimeError, ValueError):
+                dfs.append(dfs[0])
```
PaulJonasJost, line 108:
```suggestion
        dfs = []

        modes = [MODE_FUN, MODE_RES]
        for mode in modes:
            try:
                dfs.append(objective.check_grad_multi_eps(
                            free_indices,
                            multi_eps=[1e-3, 1e-4, 1e-5],
                            mode = mode))
            except (RuntimeError, ValueError):
                dfs.append(dfs[0])
```
I think you could just leave out `check_grad`.

stephanmg:
Added your suggestion @PaulJonasJost - however two checks still fail, most likely not due to my changes. Could you check? Thanks!

pypesto/petab/importer.py
```diff
@@ -78,6 +80,38 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        dfs = []
+
+        modes = [MODE_FUN, MODE_RES]
+        for mode in modes:
+            try:
+                dfs.append(objective.check_grad_multi_eps(
+                            free_indices,
+                            multi_eps=[1e-3, 1e-4, 1e-5],
+                            mode=mode))
+            except (RuntimeError, ValueError):
```
dweindl, line 107: Short comment why/when this is expected to happen?

pypesto/petab/importer.py
```diff
@@ -78,6 +80,38 @@ def from_yaml(yaml_config: Union[dict, str],
             output_folder=output_folder,
             model_name=model_name)
 
+    def check_and_log_gradients(self, RTOL: float = 1e-2, ATOL: float = 1e-3):
```
dweindl, line 83: Please use lower-case argument names

pypesto/petab/importer.py
```diff
@@ -78,6 +80,38 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        """Check and log if gradients are compatible with gradient-based optimizers
+        Parameters
```
dweindl, line 85:
```suggestion
        """Check and log if gradients match finite differences
        
        Parameters
```

dweindl (commented):
Generally useful. I am only wondering if it might be better to move it to a different place . It doesn't really depend on PEtab. `pypesto.Problem`? @yannikschaelte ?

dilpath:
> Generally useful. I am only wondering if it might be better to move it to a different place . It doesn't really depend on PEtab. `pypesto.Problem`? @yannikschaelte ?

Perhaps the [`ObjectiveBase`](https://github.com/ICB-DCM/pyPESTO/blob/main/pypesto/objective/base.py#L17) class, where the gradient check methods are implemented? However, the convenience of automatically using `petab_problem.x_nominal_scaled` as the test vector would then be lost. This could still be possible by redefining `check_and_log_gradients` in `AmiciObjective`, as e.g.:
```python
def check_and_log_gradients(self, x: np.ndarray = None, *args, **kwargs):
    if 'petab_problem' in dir(self.amici_object_builder):
    return super(ObjectiveBase, self).check_and_log_gradients(x=x, *args, **kwargs)
```

pypesto/petab/importer.py
```diff
@@ -78,6 +80,38 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        objective.amici_solver.setRelativeTolerance(1e-12)
+        free_indices = par[problem.x_free_indices]
+        dfs = []
+
+        modes = [MODE_FUN, MODE_RES]
+        for mode in modes:
+            try:
+                dfs.append(objective.check_grad_multi_eps(
+                            free_indices,
+                            multi_eps=[1e-3, 1e-4, 1e-5],
```
dilpath, line 105: Could remove `multi_eps` from here (`check_grad_multi_eps` already has a default `multi_eps`) and replace with `*args` and `**kwargs`, so that the user has more control over the gradient check.

pypesto/petab/importer.py
```diff
@@ -78,6 +80,38 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        objective = problem.objective
+        objective.amici_solver.setSensitivityMethod(
+            amici.SensitivityMethod_forward)
+        objective.amici_solver.setAbsoluteTolerance(1e-10)
+        objective.amici_solver.setRelativeTolerance(1e-12)
+        free_indices = par[problem.x_free_indices]
+        dfs = []
+
+        modes = [MODE_FUN, MODE_RES]
+        for mode in modes:
```
dilpath, line 101: This loop over `modes` could be by default, but also allow users to only specify one of `MODE_FUN` or `MODE_RES`, to reduce computation (perhaps not a significant issue except for large problems).

pypesto/petab/importer.py
```diff
@@ -78,6 +80,38 @@ def from_yaml(yaml_config: Union[dict, str],
...
+            try:
+                dfs.append(objective.check_grad_multi_eps(
+                            free_indices,
+                            multi_eps=[1e-3, 1e-4, 1e-5],
+                            mode=mode))
+            except (RuntimeError, ValueError):
+                dfs.append(dfs[0])
```
dilpath, line 108: Not sure, what does this achieve? Also, it will raise an error if `dfs == []` too.

pypesto/petab/importer.py
```diff
@@ -78,6 +80,38 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        RTOL: relative error tolerance (default 1e-2)
+        ATOL: absolute error tolerance (default 1e-3)
+        """
+        par = np.asarray(self.petab_problem.x_nominal_scaled)
+        problem = self.create_problem()
+        objective = problem.objective
+        objective.amici_solver.setSensitivityMethod(
+            amici.SensitivityMethod_forward)
+        objective.amici_solver.setAbsoluteTolerance(1e-10)
+        objective.amici_solver.setRelativeTolerance(1e-12)
```
dilpath, line 96: These hard-coded problem-dependent settings should instead be default values that the user can customize (some models will fail at these tolerances, or may be overly computationally-expensive with forward sensitivities). Alternatively, the function could accept an objective that is already customized by the user.

pypesto/petab/importer.py
```diff
@@ -78,6 +80,38 @@ def from_yaml(yaml_config: Union[dict, str],
...
             model_name=model_name)
 
+    def check_and_log_gradients(self, RTOL: float = 1e-2, ATOL: float = 1e-3):
+        """Check and log if gradients are compatible with gradient-based optimizers
+        Parameters
+        ----------
+        RTOL: relative error tolerance (default 1e-2)
+        ATOL: absolute error tolerance (default 1e-3)
+        """
+        par = np.asarray(self.petab_problem.x_nominal_scaled)
```
dilpath, line 90: Could be by default but also accept a user-supplied vector.

pypesto/petab/importer.py
dilpath: If this is to ensure `len(dfs) == 2` for the `return` statement, then see my suggestion there for an alternative.

pypesto/petab/importer.py
```diff
@@ -78,6 +80,38 @@ def from_yaml(yaml_config: Union[dict, str],
...
+                            free_indices,
+                            multi_eps=[1e-3, 1e-4, 1e-5],
+                            mode=mode))
+            except (RuntimeError, ValueError):
+                dfs.append(dfs[0])
+
+        return np.all((dfs[0].rel_err.values < RTOL) |
+                      (dfs[0].abs_err.values < ATOL) |
+                      (dfs[1].rel_err.values < RTOL) |
+                      (dfs[1].abs_err.values < ATOL))
```
dilpath, line 113:
```suggestion
        # Gradients are correct if any of the error types are within their
        # respective tolerance, for any of the modes.
        return any([
            any([
                np.all(mode_df.rel_err.values < RTOL),
                np.all(mode_df.abs_err.values < ATOL),
            ])
            for mode_df in dfs
        ])
```

For another suggestion.

test/petab/test_petab_import.py
```diff
@@ -76,6 +76,13 @@ def test_3_optimize(self):
             self.assertTrue(np.isfinite(
                 result.optimize_result.get_for_key('fval')[0]))
 
+    def test_4_gradients(self):
```
dilpath, line 79:
```suggestion
    def test_check_and_log_gradients(self):
```

pypesto/petab/importer.py
```diff
@@ -1,4 +1,6 @@
+from pypesto.objective.constants import MODE_FUN, MODE_RES
```
dilpath, line 1:
```suggestion
from ..objective.constants import MODE_FUN, MODE_RES
```

Relative imports are a few lines below, could move this there.

pypesto/petab/importer.py
```diff
@@ -78,6 +80,38 @@ def from_yaml(yaml_config: Union[dict, str],
             output_folder=output_folder,
             model_name=model_name)
 
+    def check_and_log_gradients(self, RTOL: float = 1e-2, ATOL: float = 1e-3):
+        """Check and log if gradients are compatible with gradient-based optimizers
```
dilpath, line 84: What is the `log` functionality of this method?

dilpath (commented):
:+1:

Mostly suggestions, feel free to ignore, thanks for the helper method!

pypesto/petab/importer.py
stephanmg: Added in commit, see below.

pypesto/petab/importer.py
stephanmg: Corrected.

pypesto/petab/importer.py
stephanmg: Changed the documentation brief.

pypesto/petab/importer.py
stephanmg: Done.

pypesto/petab/importer.py
stephanmg: Okay, default is to consider both modes but give the user control over it.

pypesto/petab/importer.py
stephanmg: Made these hardcoded values (1e-10 and 1e-12) a user input, allow for pre-configured objective function as user input as well.

pypesto/petab/importer.py
stephanmg: Made as default and allow user defined tolerances.

pypesto/petab/importer.py
stephanmg: Merged.

test/petab/test_petab_import.py
stephanmg: Ah I thought you want to enumerate the test in the class, as the previous tests are enumerated. Changed!

pypesto/petab/importer.py
stephanmg: Cleaned up imports.

pypesto/petab/importer.py
stephanmg: You are right, the logging is implictly done if log level is specified to `check_grad_multi_eps`.

stephanmg:
@dilpath @dweindl @PaulJonasJost please compare if I added all your suggestions.

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        self,
+        *args,
+        rtol: float = 1e-2,
+        atol: float = 1e-3,
+        mode: Literal = None,
+        objective: Objective = None,
+        objAbsoluteTolerance: float = 1e-10,
+        objRelativeTolerance: float = 1e-12,
+        multi_eps=None,
+        **kwargs
```
yannikschaelte, line 94:
```suggestion
        **kwargs,
```

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        dfs = []
+        modes = []
+
+        if mode is None:
+            modes = [MODE_FUN, MODE_RES]
+        else:
+            modes = [mode]
+
+        if multi_eps is None:
+            multi_eps = [1e-3, 1e-4, 1e-5]
```
yannikschaelte, line 132: in ODE problems, we often find FDs of step size between 1e-1 and 1e-8 stable (see also https://github.com/ICB-DCM/pyPESTO/blob/develop/pypesto/objective/finite_difference.py#L66), so one could define a broader default regime.

yannikschaelte (commented):
> > Generally useful. I am only wondering if it might be better to move it to a different place . It doesn't really depend on PEtab. `pypesto.Problem`? @yannikschaelte ?
> 
> Perhaps the [`ObjectiveBase`](https://github.com/ICB-DCM/pyPESTO/blob/main/pypesto/objective/base.py#L17) class, where the gradient check methods are implemented? However, the convenience of automatically using `petab_problem.x_nominal_scaled` as the test vector would then be lost. This could still be possible by redefining `check_and_log_gradients` in `AmiciObjective`, as e.g.:
> 
> ```python
> def check_and_log_gradients(self, x: np.ndarray = None, *args, **kwargs):
>     if x is None and 'petab_problem' in dir(self.amici_object_builder):
>         x = self.amici_object_builder.petab_problem.x_nominal_scaled
>     return super(ObjectiveBase, self).check_and_log_gradients(x=x, *args, **kwargs)
> ```

sounds good to me, i.e. defining generically in the base class and for amici/petab allowing to use the reference parameter.

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        atol: float = 1e-3,
+        mode: Literal = None,
+        objective: Objective = None,
+        objAbsoluteTolerance: float = 1e-10,
+        objRelativeTolerance: float = 1e-12,
+        multi_eps=None,
+        **kwargs
+    ):
+        """Check if gradients match finite differences (FDs)
+        Parameters
```
dweindl, line 97:
```suggestion
        """Check if gradients match finite differences (FDs)
        
        Parameters
```

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        ----------
+        rtol: relative error tolerance
+        atol: absolute error tolerance
+        mode: unction values or residuals
```
dweindl, line 101:
```suggestion
        mode: function values or residuals
```

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
...
 
+    def check_gradients(
+        self,
+        *args,
+        rtol: float = 1e-2,
+        atol: float = 1e-3,
+        mode: Literal = None,
+        objective: Objective = None,
+        objAbsoluteTolerance: float = 1e-10,
+        objRelativeTolerance: float = 1e-12,
```
dweindl, line 92:
```suggestion
        obj_absolute_tolerance: float = 1e-10,
        obj_relative_tolerance: float = 1e-12,
```

We use snake_case for everything but class names (PEP8). Please also adapt below.

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
...
+        atol: absolute error tolerance
+        mode: unction values or residuals
+        objective: objective function
+        objAbsoluteTolerance: absolute tolerance in sensitivity calculation
+        objRelativeTolerance: relative tolerance in sensitivity calculation
+        multi_eps: multiple test step width for FDs
+
+        Returns
+        -------
+        bool
```
dweindl, line 109: I'd prefer using return type hints in the function signature instead of specifying it in the docstring.

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
             output_folder=output_folder,
             model_name=model_name)
 
+    def check_gradients(
+        self,
+        *args,
+        rtol: float = 1e-2,
+        atol: float = 1e-3,
+        mode: Literal = None,
+        objective: Objective = None,
```
dweindl, line 90:
```suggestion
        objective: Optional[Objective] = None,
```

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
             output_folder=output_folder,
             model_name=model_name)
 
+    def check_gradients(
+        self,
+        *args,
+        rtol: float = 1e-2,
+        atol: float = 1e-3,
+        mode: Literal = None,
```
dweindl, line 89:
```suggestion
        mode: Optional[Literal] = None,
```

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
...
+            try:
+                dfs.append(objective.check_grad_multi_eps(
+                            free_indices, *args, **kwargs,
+                            mode=mode, multi_eps=multi_eps))
+            except (RuntimeError, ValueError):
+                # Might happen in case PEtab problem not well defined or
+                # fails for specified tolerances in forward sensitivities
+                pass
+
+        return any([
```
dweindl, line 144:
```suggestion
        return all([
```

Shouldn't that be `all`? If it fails for one mode, I'd consider that a failure.

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
...
+
+        for mode in modes:
+            try:
+                dfs.append(objective.check_grad_multi_eps(
+                            free_indices, *args, **kwargs,
+                            mode=mode, multi_eps=multi_eps))
+            except (RuntimeError, ValueError):
+                # Might happen in case PEtab problem not well defined or
+                # fails for specified tolerances in forward sensitivities
+                pass
```
dweindl, line 142: What will the function return in this case? Or will it raise during the check? I wouldn't not want it return True.

pypesto/petab/importer.py
stephanmg: Changed.

pypesto/petab/importer.py
stephanmg: Okay.

pypesto/petab/importer.py
```diff
@@ -78,6 +81,74 @@ def from_yaml(yaml_config: Union[dict, str],
...
+
+        return any([
+            any([
+                np.all(mode_df.rel_err.values < rtol),
+                np.all(mode_df.abs_err.values < atol),
```
dweindl, line 147: Hm, if one row passes only the relative tolerance and another row passes only the absolute tolerance, this report failure, right? This would be rather unexpected for me.

pypesto/petab/importer.py
stephanmg: Yes it should be all, you are right.

pypesto/petab/importer.py
stephanmg: So you suggest to make this an `and` statement?

pypesto/petab/importer.py
stephanmg: Returning `False` now instead of pass.

pypesto/petab/importer.py
stephanmg: Can't use plain `Literal` in `Optional`.

stephanmg:
To which class should the method be moved? `ObjectiveBase`?

pypesto/petab/importer.py
dweindl:
```suggestion
                np.all((mode_df.rel_err.values < rtol) | (mode_df.abs_err.values < atol)),
```

stephanmg: Yes,  that's what I just added. :)

dweindl:
> To which class should the method be moved? `ObjectiveBase`?

Meanwhile it has an AMICI dependency, so `ObjectiveBase` is rather not suitable anymore. Either that part needs to be thrown out, or it has to go somewhere pypesto.objective.amici*. I think others have a better sense of where it fits than I do.

stephanmg:
> > To which class should the method be moved? `ObjectiveBase`?
> 
> Meanwhile it has an AMICI dependency, so `ObjectiveBase` is rather not suitable anymore. Either that part needs to be thrown out, or it has to go somewhere pypesto.objective.amici*. I think others have a better sense of where it fits than I do.

Okay, maybe @yannikschaelte could clarify? Then I will make the changes accordingly. Thank you!

yannikschaelte:
> > > To which class should the method be moved? `ObjectiveBase`?
> > 
> > 
> > Meanwhile it has an AMICI dependency, so `ObjectiveBase` is rather not suitable anymore. Either that part needs to be thrown out, or it has to go somewhere pypesto.objective.amici*. I think others have a better sense of where it fits than I do.
> 
> Okay, maybe @yannikschaelte could clarify? Then I will make the changes accordingly. Thank you!

What amici dependency do you mean? The amici specific forward sensi + tolerance instruction obviously only makes sense for amici, so the base method would not have those, but the rest should work?

dweindl:
> What amici dependency do you mean? The amici specific forward sensi + tolerance instruction obviously only makes sense for amici, so the base method would not have those, but the rest should work?

Exactly those. Breaks encapsulation and adds a cyclic dependency.

stephanmg:
> > What amici dependency do you mean? The amici specific forward sensi + tolerance instruction obviously only makes sense for amici, so the base method would not have those, but the rest should work?
> 
> Exactly those. Breaks encapsulation and adds a cyclic dependency.

Yes that's what I noticed (cyclic dependency) as well.

dweindl:
P.S.: I'd rather remove the AMICI part. In my opinion, these settings should have been applied consciously by the user before running and checks here.

stephanmg:
So remove the `AMICI` part and add an required argument for an `AMICIObjective`  or `ObjectiveBase` instead to the method?

test/petab/test_petab_import.py
dilpath: Ah, didn't notice sorry, fine as before then.

dilpath:
> > > What amici dependency do you mean? The amici specific forward sensi + tolerance instruction obviously only makes sense for amici, so the base method would not have those, but the rest should work?
> > 
> > 
> > Exactly those. Breaks encapsulation and adds a cyclic dependency.
> 
> Yes that's what I noticed (cyclic dependency) as well.

This suggestion resolves this right?

https://github.com/ICB-DCM/pyPESTO/pull/690#pullrequestreview-697915125

stephanmg:
Removed the AMICI dependencies, still have to move functionality to `BaseObjective`.

stephanmg:
Functionality now moved to BaseObjective, added delegate to `amici.py` for convenient usage.

pypesto/petab/importer.py
yannikschaelte: agreed, call it "match" or something

test/petab/test_petab_import.py
```diff
@@ -76,6 +76,21 @@ def test_3_optimize(self):
             self.assertTrue(np.isfinite(
                 result.optimize_result.get_for_key('fval')[0]))
 
+    def test_check_gradients(self):
+        # check gradients of simple model (should always be a true positive)
```
yannikschaelte, line 80:
```suggestion
        """Test objective FD-gradient check function."""
        # Check gradients of simple model (should always be a true positive)
```

test/petab/test_petab_import.py
```diff
@@ -76,6 +76,21 @@ def test_3_optimize(self):
...
+        petab_problem = pypesto.petab.PetabImporter.from_yaml(
+            os.path.join(folder_base, model_name, model_name + '.yaml'))
+
+        objective = petab_problem.create_objective()
+        objective.amici_solver.setSensitivityMethod(
+                 amici.SensitivityMethod_forward)
+        objective.amici_solver.setAbsoluteTolerance(1e-10)
+        objective.amici_solver.setRelativeTolerance(1e-12)
+
+        self.assertFalse(petab_problem.check_gradients(
```
yannikschaelte, line 91: why is this False again? maybe add a scenario where it's true? this https://github.com/ICB-DCM/pyPESTO/blob/develop/test/util.py#L122 should be stable enough no matter what.

yannikschaelte (approved):
looks good to me, thanks! feel free to merge when you think it's fine.

PR #699, opened 2021-07-09
updated store.ipynb.

PaulJonasJost:
Included HDF5 history and updated saving of results.


doc/example/store.ipynb
```diff
@@ -218,25 +310,123 @@
...
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "Just as in CSV, writing the history to HDF5 takes a considerable amount of time. Note that HDF5 in combination with\n",
+    "`MultiProcessEngine()`will automatically create one file with filename `filename_{id}` for each start, put them in a\n",
+    "folder namen `filename` and create one HDF5 file with name `filename`, which combines all those.The `filename`-file is\n",
+    "intended to be also useable as a savefile for the end results. Note however that `filename` will **cease to function**\n",
+    "if you delete the folder."
```
dilpath, line 323:
```suggestion
    "Just as in CSV, writing the history to HDF5 takes a considerable amount of time. Note that HDF5 history in combination with\n",
    "the parallelization engine `MultiProcessEngine()` will result in one file with filename `filename_{id}` for each start, which are stored in a\n",
    "folder named `filename`, then after optimization an HDF5 file named `filename` is created, which combines all start files. The combined file is\n",
    "intended to be also usable as a savefile for the end results. Note however that the combined file references the start files in the folder, hence the combined file will **cease to function**\n",
    "if you delete the folder of start files."
```

The way it's currently written, the following files would be generated if an output file named `results.hdf5` is specified:
- folder containing start files: `results.hdf5/`
- start files: `results.hdf5/results.hdf5_{n_start}`
- combined file: `results.hdf5`

Maybe instead illustrating with an example would make things easier to understand? e.g.:

If a user specifies a HDF5 output file named `my_results.hdf5` and uses a parallelization engine, then:
- a folder is created to contain partial results, named `my_results/` (the stem of the output filename)
- files are created to store the results of each start, named `my_results/my_results_{START_INDEX}.hdf5`
- a file is created to store the combined result from all starts, named `my_results.hdf5`. Note that this file depends on the files in the `results/` directory, so will become unusable if `my_results/` is deleted.

doc/example/store.ipynb
```diff
@@ -218,25 +310,123 @@
...
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
    "source": [
     "## Result storage\n",
     "\n",
-    "Result objects can be stored to HDF5 files. When appliable, this is preferable to just pickling results, which is not guaranteed to be reproducible in the future."
+    "Result objects can be stored to HDF5 files. When appliable, this is preferable to just pickling results, which is\n",
+    "not guaranteed to be reproducible in the future."
```
dilpath, line 374:
```suggestion
    "Result objects can be stored as HDF5 files. When applicable, this is preferable to just pickling results, which is\n",
    "not guaranteed to be reproducible in the future."
```

doc/example/store.ipynb
```diff
@@ -246,136 +436,115 @@
...
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
    "source": [
     "### Save optimization result as HDF5 file\n",
     "\n",
-    "The optimization result can be saved via a `pypesto.store.OptimizationResultHDF5Writer`."
+    "The optimization result can be saved via a `pypesto.store.write_result()`. This will write the problem, the\n",
+    "optimization-,profile- and sampling_result to hdf5. All of them can be disabled through boolean values."
```
dilpath, line 514:
```suggestion
    "The optimization result can be saved with `pypesto.store.write_result()`. This will write the problem and the\n",
    "optimization result, and the profiling and sampling results if available, to HDF5. All of them can be disabled with boolean flags."
```

Can add a link to the documentation for the function.

doc/example/store.ipynb
```diff
@@ -384,70 +553,45 @@
...
-   "metadata": {},
-   "outputs": [],
+   "execution_count": 12,
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "WARNING: You are loading a problem.\n",
+      "This problem is not to be used without a separately created objective.\n",
```
dilpath, line 563: Is this still correct or fixed in `develop`? #669

doc/example/store.ipynb
```diff
@@ -218,25 +310,123 @@
...
+    "visualize.optimizer_history(result, ax=ax[1])\n",
+    "fig.set_size_inches((15, 5))"
+   ],
+   "execution_count": 7,
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "History type:  <class 'pypesto.objective.history.CsvHistory'>\n"
```
dilpath, line 348: This cell is identical to the previous cell? History type in the output is also `CsvHistory`.

dilpath (approved):
:+1:

doc/example/store.ipynb
PaulJonasJost: it is still correct, as we do not save the objective but rather just information of the objective.

yannikschaelte (approved):

PR #702, opened 2021-07-12
added option for weighted mean in ensemble

PaulJonasJost:
This for now is only a draft.
It assumes, that the post processor either returns only a np.ndarray or a dict/dict like object.

Perhaps want to add another default post processor, that would only need to receive `AMICI_outputs` do creat a return dict of things that should be returned (like 'x', 'y', 'llh' etc).

Also need to add sigmas.

Currently the weighting is done in terms of llh of each condition separately. should I go with the llh for one parameter across all conditions?


pypesto/predict/amici_predictor.py
```diff
@@ -296,7 +296,9 @@ def _default_output(amici_outputs):
             outputs_sensi = self.post_processor_sensi(amici_outputs)
         if self.post_processor_time is not None:
             timepoints = self.post_processor_time(amici_outputs)
-
+        # return the likelihood value additionally if it is requested
+        if 'llh' in amici_outputs:
+            return timepoints, outputs, outputs_sensi, amici_outputs['llh']
```
PaulJonasJost, line 301:
```suggestion
        
```

pypesto/predict/amici_predictor.py
```diff
@@ -296,7 +296,7 @@ def _default_output(amici_outputs):
             outputs_sensi = self.post_processor_sensi(amici_outputs)
         if self.post_processor_time is not None:
             timepoints = self.post_processor_time(amici_outputs)
-
+
```
PaulJonasJost, line 299:
```suggestion

```

pypesto/ensemble/ensemble.py
```diff
@@ -138,7 +140,8 @@ def condense_to_arrays(self):
         }
 
     def compute_summary(self,
-                        percentiles_list: Sequence[int] = (5, 20, 80, 95)
+                        percentiles_list: Sequence[int] = (5, 20, 80, 95),
+                        weighting: bool = True
```
dilpath, line 144:
```suggestion
                        weigh: bool = True
```
Option can be `weigh` and values can be `weights`

pypesto/ensemble/ensemble.py
```diff
@@ -160,6 +163,11 @@ def compute_summary(self,
         if not self.prediction_results:
             raise ArithmeticError('Cannot compute summary statistics from '
                                   'empty prediction results.')
+        # check if weightings shall be used or not depending on whether llh
+        # is part of the prediciton output
```
dilpath, line 167:
```suggestion
        # Check whether LLH-based weights can be used.
```

pypesto/ensemble/ensemble.py
dilpath: Perhaps this should be `False` by default?

pypesto/ensemble/ensemble.py
```diff
@@ -160,6 +163,11 @@ def compute_summary(self,
...
+        if AMICI_LLH not in self.prediction_results[0].conditions[0].output:
+            weighting = False
```
dilpath, line 169: A user might explicitly want weights but not realize they're not used because of this. A warning might help?

pypesto/ensemble/ensemble.py
```diff
@@ -192,14 +205,31 @@ def _stack_outputs_sensi(ic: int):
...
             Computes means, standard deviation, median, and requested
             percentiles for a set of stacked simulations
             """
             summary = {}
-            summary[MEAN] = np.mean(tmp_array, axis=-1)
-            summary[STANDARD_DEVIATION] = np.std(tmp_array, axis=-1)
+            summary[MEAN] = np.average(tmp_array, axis=-1, weights=weights)
+            summary[STANDARD_DEVIATION] = np.average(
+                (tmp_array.T-summary[MEAN].T).T**2,
+                axis=-1, weights=weights)
```
dilpath, line 232: Variance?

dilpath (commented):
I think a test would be good

pypesto/ensemble/ensemble.py
PaulJonasJost: is just a matter of taste I think. As `LLH` is not included in a predictor by default, this will get turned to `False` in line 168. The idea behind that for me was that if you include `llh` in predictor for ensembles the default you probably would want is `True`.

pypesto/ensemble/ensemble.py
PaulJonasJost: oh yes, there needs to be an additional np.sqrt() 👍

pypesto/ensemble/ensemble.py
dilpath: So in the default setting, `weighting==True` but no weights are used? Might be confusing for users. Could call it something more descriptive, e.g. `weights_if_llh_provided`, or make it `False` by default?

pypesto/ensemble/ensemble.py
dilpath: Then equivalent to `np.average(np.abs(tmp_array - summary[MEAN]), axis=-1, weights=weights)` I guess

pypesto/ensemble/ensemble.py
```diff
@@ -172,8 +180,13 @@ def _stack_outputs(ic: int):
...
                 return None
             # stack predictions
-            output_list = [prediction.conditions[ic].output
-                           for prediction in self.prediction_results]
+            if isinstance(self.prediction_results[0].conditions[ic].output,
+                          np.ndarray):
+                output_list = [prediction.conditions[ic].output
+                               for prediction in self.prediction_results]
+            else:
+                output_list = [prediction.conditions[ic].output[AMICI_Y]
```
paulstapor, line 188: I just saw below, that it seems you intend to change the type of `prediction.conditions[ic].output`. However, this is a data member of the class PredictionConditionResult. I'd strongly argue that data members of classes should not switch type, if possible, and if inevitable: Only during major refactorings, as this will typically break backwards compatibility. Rather add a new data member for the weights...

pypesto/ensemble/ensemble.py
```diff
@@ -192,14 +205,31 @@ def _stack_outputs_sensi(ic: int):
...
+            Group weights for different parameter vectors of one ensemble
+            together, if they belong to the same simulation condition, and
+            stacks them in one array
+            """
+            # Were outputs computed
+            if self.prediction_results[0].conditions[ic].output is None:
+                return None
+            # stack predictions
+            output_list = [prediction.conditions[ic].output[AMICI_LLH]
+                           for prediction in self.prediction_results]
```
paulstapor, line 219: Hm... I have the impression that this will be problematic. By default, a PredictionConditionResult has a member "output", which is supposed to be an ndarray. You're making a dict out of it, which might brake quite some dependencies...
I would rather suggest you add a new member "output_weight" to the PredictionConditionResult. This way, you can just use this new member here, and an additional field will not affect the functioning of old script, hence not destroy backwards compatibility.

pypesto/ensemble/ensemble.py
PaulJonasJost: no, at least that should not be the case mathematically...

jvanhoefer (commented):
Since I am not really into that part of the code, I did only review code style (otherwise I would have to get an overview of the rest of the ensemble code first). So please let someone else review regarding functionality/structrue, but from style point of view: Looks fine to me :)

pypesto/ensemble/ensemble.py
```diff
@@ -219,10 +275,19 @@ def _compute_summary(tmp_array, percentiles_list):
             # create a temporary array with all the outputs needed and wanted
             tmp_output = _stack_outputs(i_cond)
             tmp_output_sensi = _stack_outputs_sensi(i_cond)
+            tmp_weights = np.ones(tmp_output.shape[-1])
+            if weighting:
+                tmp_weights = np.exp(_stack_weights(i_cond))
```
dilpath, line 280: Add a comment to explain why `np.exp` is used here (log-likelihood to likelihood?). If the method is described elsewhere, can reference

pypesto/ensemble/ensemble.py
```diff
@@ -149,6 +155,11 @@ def compute_summary(self,
         ----------
         percentiles_list:
             List or tuple of percent numbers for the percentiles
+        weighting:
+            Whether weights should be used for trajectory.
+        compute_chi2:
+            Whether weighted standard deviation of the ensemble mean trajectory
+            should be computed. Defaults to False.
```
dilpath, line 162: `compute_weighted_sigma`?

pypesto/ensemble/ensemble.py
```diff
@@ -256,6 +321,37 @@ def _compute_summary(tmp_array, percentiles_list):
...
+    def compute_chi2(self, amici_objective: AmiciObjective):
+        if (self.prediction_summary[MEAN] is None) or (
+                self.prediction_summary[WEIGHTED_SIGMA] is None):
+            try:
+                self.compute_summary(
+                        weighting=True,
+                        compute_weighted_sigma=True)
+            except TypeError:
+                raise ValueError('Computing a summary failed.')
+        n_conditions = len(self.prediction_results[0].conditions)
```
dilpath, line 333: Can conditions differ between predictions? If so, can simply raise a `NotImplementedError` if there are different conditions between the prediction results.

pypesto/ensemble/ensemble.py
```diff
@@ -256,6 +321,37 @@ def _compute_summary(tmp_array, percentiles_list):
...
+        chi_2 = []
+        for i_cond in range(n_conditions):
+            # get measurements and put into right form
+            y_meas = amici_objective.edatas[i_cond].getObservedData()
+            y_meas = np.array(y_meas)
+            # bring into shape (n_t,n_y)
+            y_meas = y_meas.reshape(
+                (int(len(y_meas)/amici_objective.edatas[0].nytrue()),
```
dilpath, line 341: Can add a comment e.g. `# assumes all observables are provided at all time points`.

Could you replace this with e.g. `amici_objective.edatas[0].nt()`?
https://github.com/AMICI-dev/AMICI/blob/509c850b57dd4197253d0cf30ffb4ba04c752d05/include/amici/edata.h#L145

pypesto/ensemble/ensemble.py
```diff
@@ -256,6 +321,37 @@ def _compute_summary(tmp_array, percentiles_list):
...
+                 amici_objective.edatas[0].nytrue())
+            )
+            mean_traj = \
+                self.prediction_summary[MEAN].conditions[i_cond].output
+            weighted_sigmas = \
+                self.prediction_summary[
+                    WEIGHTED_SIGMA].conditions[i_cond].output
+            if y_meas.shape != mean_traj.shape:
+                raise ValueError('Shape of trajectory and shape '
+                                 'of measurements does not match.')
```
dilpath, line 351: I guess this assumes equal number of timepoints -> same timepoints. However, given the `pypesto.objective.AmiciObjective.set_custom_timepoints` method, maybe this can't be assumed. Could check timepoints explicitly.

pypesto/ensemble/ensemble.py
```diff
@@ -256,6 +321,37 @@ def _compute_summary(tmp_array, percentiles_list):
         # also return the object
         return self.prediction_summary
 
+    def compute_chi2(self, amici_objective: AmiciObjective):
```
dilpath, line 324: docstrings

pypesto/ensemble/constants.py
```diff
@@ -32,6 +32,9 @@
 STANDARD_DEVIATION = 'std'
 PERCENTILE = 'percentile'
 SUMMARY = 'summary'
+CHI2 = 'chi2'
+WEIGHTED_SIGMA = 'weighted_sigma'
+CONDITIONS = 'conditions'
```
dilpath, line 37: Is this used?

dilpath (approved):
Looks good :+1:

Maybe an example in a notebook?

pypesto/ensemble/constants.py
PaulJonasJost: only weighed sigma as it seems. deleted the others.

pypesto/ensemble/ensemble.py
```diff
@@ -41,7 +43,8 @@ def __init__(
             prediction_id: str = None,
             prediction_results: Sequence[PredictionResult] = None,
             lower_bound: Sequence[np.ndarray] = None,
-            upper_bound: Sequence[np.ndarray] = None):
+            upper_bound: Sequence[np.ndarray] = None,
+            x_vectors: np.ndarray = None):
```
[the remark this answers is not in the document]
PaulJonasJost: deleted 👍

pypesto/ensemble/ensemble.py
```diff
@@ -160,6 +171,12 @@ def compute_summary(self,
         if not self.prediction_results:
             raise ArithmeticError('Cannot compute summary statistics from '
                                   'empty prediction results.')
+        # if weightings shall be used, check whether weights are there
+        if weighting:
+            if not self.prediction_results[0].conditions[0].output_weight:
+                logger.warning('WARNING: There are no weights in the '
+                               'prediction results. Will use 1 as weight.')
```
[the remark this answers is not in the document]
PaulJonasJost: what do you mean by that? you mean do a try... except loop? how would that benefit? this is only triggered if you actually want to use weightings but none are existent so I would not count that as a default?

pypesto/ensemble/ensemble.py
```diff
@@ -192,22 +209,61 @@ def _stack_outputs_sensi(ic: int):
             # stack into one numpy array
             return np.stack(output_sensi_list, axis=-1)
 
-        def _compute_summary(tmp_array, percentiles_list):
+        def _stack_weights(ic: int):
+            """
+            Group weights for different parameter vectors of one ensemble
+            together, if they belong to the same simulation condition, and
+            stacks them in one array
+            """
```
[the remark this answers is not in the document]
PaulJonasJost: should I then add it in all the other older functions as well? the return type hints as well?

pypesto/ensemble/ensemble.py
```diff
@@ -192,22 +209,61 @@ def _stack_outputs_sensi(ic: int):
...
             """
             Computes means, standard deviation, median, and requested
             percentiles for a set of stacked simulations
             """
             summary = {}
-            summary[MEAN] = np.mean(tmp_array, axis=-1)
-            summary[STANDARD_DEVIATION] = np.std(tmp_array, axis=-1)
+            summary[MEAN] = np.average(tmp_array, axis=-1, weights=weights)
+            summary[STANDARD_DEVIATION] = np.sqrt(np.average(
+                (tmp_array.T-summary[MEAN].T).T**2,
```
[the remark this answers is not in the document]
PaulJonasJost: tested it again and if I remove the .T's it does actually change the result... I think that has something to do with tmp_array being more then two dimensional...

pypesto/ensemble/ensemble.py
PaulJonasJost: they should not be able to, as the ensembles currently are created from the same amici_model which should generate the prediction the same no matter what the actual parameter values are.

Might be interesting way down the train, when we allow for different model structures to contribute to the same ensemble.

pypesto/ensemble/ensemble.py
PaulJonasJost: not sure, but shouldn't the whole optimization fail if we set custom timepoints as we could not compute a likelihood anymore? 🤔 correct me if I am wrong.

pypesto/ensemble/ensemble.py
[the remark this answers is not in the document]
dilpath: Ah I mean, because the default here is `weighting=False`, then `weighting=True` suggests the user explicitly requested weights (or is using another method that does). If weights are explicitly requested, then this suggests that the user thinks weights are available. It would then be an error (i.e. `Exception`, not warning) if weights are not available (as the user can simply provide `weighting=False` to use 1 as weights).

pypesto/ensemble/ensemble.py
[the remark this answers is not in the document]
dilpath: Would be great, but also fine to just update for your PR

pypesto/ensemble/ensemble.py
dilpath: Conditions are provided via `amici.ExpData` (a single `amici.Model` can simulate different sets of conditions).

pypesto/ensemble/ensemble.py
dilpath: Ah, often used for predictions rather than methods that require a likelihood.

pypesto/ensemble/ensemble.py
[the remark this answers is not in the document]
PaulJonasJost: ahhh yes, now I get it, very good idea :)

pypesto/ensemble/ensemble.py
PaulJonasJost: in that case you wouldn't compute the chi^2 value either I'd say.

PR #704, opened 2021-07-15
Mark test_preeq_guesses and test_optimization as flaky

dweindl:


yannikschaelte (approved):

yannikschaelte (approved):
5 feels a bit much, in that case one may want to investiagte causes ... but ok surely

dweindl:
> 5 feels a bit much, in that case one may want to investiagte causes ... but ok surely

How much feels right? :)

yannikschaelte:
> > 5 feels a bit much, in that case one may want to investiagte causes ... but ok surely
> 
> How much feels right? :)

three and three quarters?

jvanhoefer:
@dweindl @yannikschaelte What is the status here? Is this ready to merge? :)

yannikschaelte:
sí

PR #705, opened 2021-07-16
added option to trimm the trace to a monotonically decreasing one

PaulJonasJost:
I wrote a function to get only a monotonically decreasing history trace. This is meant for comparisons with e.g. test and training data but can obviously also be used for more.

It does not change any functionality from before.

### DIFF
--- a/pypesto/objective/history.py
+++ b/pypesto/objective/history.py
@@ -338,13 +338,7 @@ class HistoryBase(abc.ABC):
         decreasing history.
         """
         fval_trace = self.get_fval_trace()
-        indices = []
-        fval_current = np.inf
-        for iter_i, fval_i in enumerate(fval_trace):
-            if fval_i <= fval_current:
-                indices.append(iter_i)
-                fval_current = fval_i
-        return indices
+        return np.where(fval_trace <= np.minimum.accumulate(fval_trace))
 
 
 class History(HistoryBase):
--- a/pypesto/objective/history.py
+++ b/pypesto/objective/history.py
@@ -338,7 +338,7 @@ class HistoryBase(abc.ABC):
         decreasing history.
         """
         fval_trace = self.get_fval_trace()
-        return np.where(fval_trace <= np.minimum.accumulate(fval_trace))
+        return np.where(fval_trace <= np.minimum.accumulate(fval_trace))[0]
 
 
 class History(HistoryBase):
--- a/pypesto/objective/history.py
+++ b/pypesto/objective/history.py
@@ -24,13 +24,16 @@ def trace_wrap(f):
     index list, and reduces for integer `ix` the output to a single value.
     """
     def wrapped_f(
-            self, ix: Union[Sequence[int], int, None] = None
+            self, ix: Union[Sequence[int], int, None] = None,
+            trimm: bool = False
     ) -> Union[Sequence[Union[float, MaybeArray]], Union[float, MaybeArray]]:
         # whether to reduce the output
         reduce = isinstance(ix, numbers.Integral)
         # default: full list
         if ix is None:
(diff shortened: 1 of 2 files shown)


pypesto/objective/history.py
```diff
@@ -24,13 +24,16 @@ def trace_wrap(f):
     index list, and reduces for integer `ix` the output to a single value.
     """
     def wrapped_f(
-            self, ix: Union[Sequence[int], int, None] = None
+            self, ix: Union[Sequence[int], int, None] = None,
+            trimm: bool = False
```
dilpath, line 28:
```suggestion
            trim: bool = False
```

pypesto/objective/history.py
```diff
@@ -319,6 +331,20 @@ def get_time_trace(
...
+        decreasing history.
+        """
+        fval_trace = self.get_fval_trace()
+        indices = []
+        fval_current = np.inf
+        for iter_i, fval_i in enumerate(fval_trace):
+            if fval_i <= fval_current:
+                indices.append(iter_i)
+                fval_current = fval_i
+        return indices
```
dweindl, line 346:
```suggestion
        return np.where(fval_trace <= np.minimum.accumulate(fval_trace))
```

dweindl (commented):
Add a test and use in some notebook?

pypesto/objective/history.py
PaulJonasJost: wow that looks really elegant :)

pypesto/objective/history.py
```diff
@@ -24,13 +24,16 @@ def trace_wrap(f):
...
+            self, ix: Union[Sequence[int], int, None] = None,
+            trim: bool = False
     ) -> Union[Sequence[Union[float, MaybeArray]], Union[float, MaybeArray]]:
         # whether to reduce the output
         reduce = isinstance(ix, numbers.Integral)
         # default: full list
         if ix is None:
             ix = np.arange(0, len(self), dtype=int)
+            if trim:
+                ix = self.get_trimmed_indices()
```
dweindl, line 36:
```suggestion
            if trim:
                ix = self.get_trimmed_indices()
            else:
                ix = np.arange(0, len(self), dtype=int)
```

test/base/test_history.py
```diff
@@ -620,3 +620,30 @@ def test_hdf5_history_mp():
...
+    """
+    problem = CRProblem()
+    pypesto_problem = problem.get_problem()
+
+    optimizer = pypesto.optimize.ScipyOptimizer()
+    history_options = pypesto.HistoryOptions(trace_record=True)
+    result = pypesto.optimize.minimize(
+        problem=pypesto_problem, optimizer=optimizer,
+        n_starts=2, history_options=history_options,
+        engine=MultiProcessEngine()
```
dweindl, line 638: Why launch two optimizations if only one is used later on?

dweindl (approved):

test/base/test_history.py
PaulJonasJost: you have a point there, I had a reason but I cannot remember what it was :D

test/base/test_history.py
```diff
@@ -620,3 +620,30 @@ def test_hdf5_history_mp():
...
+    decreasing.
+    """
+    problem = CRProblem()
+    pypesto_problem = problem.get_problem()
+
+    optimizer = pypesto.optimize.ScipyOptimizer()
+    history_options = pypesto.HistoryOptions(trace_record=True)
+    result = pypesto.optimize.minimize(
+        problem=pypesto_problem, optimizer=optimizer,
+        n_starts=2, history_options=history_options,
```
PaulJonasJost, line 637:
```suggestion
        n_starts=1, history_options=history_options,
```

PR #707, opened 2021-07-21
Feature temporarily remove literal fixes #706

### LINKED ISSUE #706: To use or not to use typing.Literal
**Bug description**
Since recently we use `Literal` in some type hints in the objective base class. I.e.

https://github.com/ICB-DCM/pyPESTO/blob/3e0cc23075f35ca07850afda752bd0891a6654b9/pypesto/objective/base.py#L6

Since `Literal` is only available for python 3.8 onwards, we would loose python 3.6 & 3.7 support (completely. One can't even import `pypesto` any more...)

**Possible solution:**

1) Get rid of the type hint. Even though it is wise to always use a current python version, we do not want to scare off users for the sake of a type hint.

2) Leave the type hint, delete python 3.6/7 support e.g. in the `setup.cfg`
https://github.com/ICB-DCM/pyPESTO/blob/72488fbb3eaa91dd163f88bac71a1a165a0da70f/setup.cfg#L37

Currently I am in favor of solution 1 (since we do not even use the `Literal` in the way I would guess it is intended to be used for, even though it could be useful at many places). Any opinions?

@yannikschaelte @stephanmg @dweindl

jvanhoefer:
hot fix for making pypesto usable on python 3.7 again...


dweindl (commented):
I guess this pull request was meant to come with some code changes? :thinking:

jvanhoefer:
@stephanmg @yannikschaelte ping? ;)

stephanmg:
There is a merge conflict, this should be addressed first.

dweindl (commented):
If the scope of this PR is to make pypesto usable with Python 3.7, I would suggest to adapt GHA workflows to include that version.

yannikschaelte (approved):
looks fine. and what the others said ;)

jvanhoefer:
> If the scope of this PR is to make pypesto usable with Python 3.7, I would suggest to adapt GHA workflows to include that version.

Since we currently do our tests anyhow only on the newest python version (i.e. 3.9) I just made this also consistent in `deploy.yml` (which still used `3.8`)

PR #621, opened 2021-04-09
account for plist in edata when slicing amici results

FFroehlich:
requires https://github.com/AMICI-dev/AMICI/pull/1487

PR #715, opened 2021-07-27
Fixes steadystate guesses

FFroehlich:
- reinitialize objective during startpoint resampling to avoid using steadystate guesses.
- fix shape bug introduced via https://github.com/AMICI-dev/AMICI/pull/1487 and https://github.com/ICB-DCM/pyPESTO/pull/621
- fix check for elementwise changes in linear updates


yannikschaelte (approved):
looks good to me if tests pass. re-initializing in startpoint method certainly makes sense.

PR #718, opened 2021-08-04
subset x_names according to preprocessor

FFroehlich:


yannikschaelte (approved):
looks good, an issue may be the proper handling of not-available values (-> propagate None?), which seems to also occur in the tests.

PR #726, opened 2021-09-24
Feature load optimization result from history

PaulJonasJost:


stephanmg (changes requested):
In `read_from_hdf5.py` the parameter `filename` is not documented. Return type annotation is missing as well. Besides this I think that the changes are fine.

stephanmg (approved):
OK.

PR #730, opened 2021-09-30
added default values to `ensemble.from_sample`

### LINKED ISSUE #727: Use information as default from problem in `ensemble.from_sample`
When calling `ensemble.from_sample(result=result)`, we could take `x_names`, and parameter bounds as default value for the ensemble, in case no other values for these arguments are defined...

@PaulJonasJost what do you think about that?

https://github.com/ICB-DCM/pyPESTO/blob/fb2be819b41411dc1686d8429ce5efa6c535b70b/pypesto/ensemble/ensemble.py#L343

PaulJonasJost:
fixes #727.

### DIFF
--- a/pypesto/ensemble/ensemble.py
+++ b/pypesto/ensemble/ensemble.py
@@ -481,6 +481,8 @@ class Ensemble:
         The ensemble.
         """
         x_vectors = result.sample_result.trace_x[0]
+        x_names = [result.problem.x_names[i]
+                   for i in result.problem.x_free_indices]
         if remove_burn_in:
             if result.sample_result.burn_in is None:
                 geweke_test(result)
@@ -489,7 +491,11 @@ class Ensemble:
         if chain_slice is not None:
             x_vectors = x_vectors[chain_slice]
         x_vectors = x_vectors.T
-        return Ensemble(x_vectors, **kwargs)
+        return Ensemble(x_vectors=x_vectors,
+                        x_names=x_names,
+                        lower_bound=result.problem.lb,
+                        upper_bound=result.problem.ub,
+                        **kwargs)
 
     @staticmethod
     def from_optimization_endpoints(


jvanhoefer (changes requested):
wouldn't we need some checks, if `x_names`, `lower_bound`, ... are in `**kwargs`?

PaulJonasJost:
> wouldn't we need some checks, if `x_names`, `lower_bound`, ... are in `**kwargs`?

Yes definitely currently on it. 👍

dilpath:
The docstring for `lower_bound` in `EnsemblePrediction` looks incorrect (also seems completely different to `upper_bound`). Is `lower_bound` used with different meaning in different places or is this just a typo?

https://github.com/ICB-DCM/pyPESTO/blob/e1d5640f848a5e0821476df31b777bdd3f56c4df/pypesto/ensemble/ensemble.py#L59-L65

yannikschaelte:
it's just shortened, right? arguably too shortened, should at least say that the similar to lower bound.

yannikschaelte:
ok thx for pointing out it's predictions vs parameters. agreed, while conceptually both may make sense, from usage it should probably be parameters.

PaulJonasJost:
Given the return statement of Ensemble.predict(), I agree, it should be parameters if anything, but does it even make sense? As far as I can see, the EnsemblePredictor does not contain any info about the parameters anymore and is just meant as simulation result. But in that case lower bounds for parameters do not make sense? Should I remove it entirely?

dilpath:
> Given the return statement of Ensemble.predict(), I agree, it should be parameters if anything, but does it even make sense? As far as I can see, the EnsemblePredictor does not contain any info about the parameters anymore and is just meant as simulation result. But in that case lower bounds for parameters do not make sense? Should I remove it entirely?

Seems like the intention is for bounds to be on parameters in `Ensemble` (see usage in `Ensemble.check_identifiability`: https://github.com/ICB-DCM/pyPESTO/blob/f37f050a367a84597c090f1ecd0ae1bd6ca7e2d8/pypesto/ensemble/ensemble.py#L699), and on predictions in `EnsemblePrediction` (currently no usage...). Hence it doesn't make sense to pass bounds from `Ensemble` to `EnsemblePrediction`, as done in `Ensemble.predict`: https://github.com/ICB-DCM/pyPESTO/blob/f37f050a367a84597c090f1ecd0ae1bd6ca7e2d8/pypesto/ensemble/ensemble.py#L666-L667

As @yannikschaelte said, both make sense, but only bounds on parameters are currently used. Maybe add a `NotImplementedError` to `EnsemblePrediction.__init__` if bounds are passed? The names `lower_bound` and `upper_bound` could also be renamed in `EnsemblePrediction` to avoid confusion with the meaning of these names elsewhere in pyPESTO (and in `Ensemble`).

Issue #728, opened 2021-09-29
Parameter bounds cannot be passed as a list to visualize.parameters_lowlevel


C-Peiter:
Hey, I think I found an issue here:
https://github.com/ICB-DCM/pyPESTO/blob/fb2be819b41411dc1686d8429ce5efa6c535b70b/pypesto/visualize/parameters.py#L290-L295

I passed `lb` and `ub` to `pypesto.visualize.parameters_lowlevel` as a `list` but got an Attribute error ```AttributeError: 'list' object has no attribute 'flatten'```.
According to the documentation, one should be able to use a list, correct?
https://github.com/ICB-DCM/pyPESTO/blob/fb2be819b41411dc1686d8429ce5efa6c535b70b/pypesto/visualize/parameters.py#L194-L198

I am using Python v3.8.11 with pyPESTO v0.2.7.

yannikschaelte:
agreed. the documentation suggests both ndarray and list are fine. as the implementation seems to use ndarray functions like flatten, the easiest thing may be to internally transform lb, ub to ndarray at the beginning. do you want to do that?

C-Peiter:
Sure, I can do that some time this week.

PR #731, opened 2021-10-01
Minor issues in parameters_lowlevel and waterfall_lowlevel

C-Peiter:
This pull request is partially linked to issue https://github.com/ICB-DCM/pyPESTO/issues/728.

In `visualize.parameters_lowlevel` lower bounds and upper bounds could not be passed to the function as a list:
- lists will now be transformed to arrays first

`visualize.parameters_lowlevel` and `visualize.waterfall_lowlevel` were both sometimes using interactive plotting instead of object oriented plotting. For example, in the waterfall plot, the `ylim` is set with `plt.ylim`. Unfortunately, when one first creates a figure with axes objects, the `ylim` will always be set on the last axes object:
- plotting is now object oriented


C-Peiter:
Sorry about the back and forth. I actually only wanted to open a pull request for the updates on parameters.py and waterfall.py, but am unsure on how to do that. The update to importer.py had already been part of another pull-request.

yannikschaelte (approved):
looks good to me, thanks for spotting these! diff only contains parameters.py and waterfall.py, which should be correct.

yannikschaelte:
> @jvanhoefer I don't think I can do that. I should see the option to merge below this pull request, correct? I don't see it though. Am I missing something here? It also says "Only those with write access to this repository can merge pull requests".

You should have gotten a mail with invitation to write access now.

PR #760, opened 2021-11-02
Trim results

FFroehlich:
Allows not storing sres/hess which significantly reduces memory consumption for large scale problems with a high number of starts.


pypesto/optimize/task.py
```diff
@@ -37,6 +40,7 @@ def __init__(
             Options object applying to optimization.
         history_options:
             Optimizer history options.
+        report_hess
```
yannikschaelte, line 43: todo

yannikschaelte (approved):
not sure whether there is 1) a better place for handling such flags, and 2) whether flags should be grouped in case more follow, but should be fine I think.

FFroehlich:
> not sure whether there is 1) a better place for handling such flags, and 2) whether flags should be grouped in case more follow, but should be fine I think.

Conceptually, this would be directly handled by the minimize function of each optimizer, but this would involve passing arguments through many functions and a decent amount of code duplication. I don't think these options will be necessary for other fields, even for thousands of starts and medium sized models, the storage requirement is a couple of MB (whereas it can be GB when sres/hess are stored).

yannikschaelte:
> > not sure whether there is 1) a better place for handling such flags, and 2) whether flags should be grouped in case more follow, but should be fine I think.
> 
> Conceptually, this would be directly handled by the minimize function of each optimizer, but this would involve passing arguments through many functions and a decent amount of code duplication. I don't think these options will be necessary for other fields, even for thousands of starts and medium sized models, the storage requirement is a couple of MB (whereas it can be GB when sres/hess are stored).

agreed, forwarding arguments here down multiple layers is something we definitely don't want to do, so should be fine.

Issue #345, opened 2020-05-19
Use np.asarray, not np.array


yannikschaelte:
In a few places, we may use np.array, but np.asarray should be enough. The latter should be more efficient as it does not necessarily create a copy. Just go over and check.

stephanmg:
Closed by PR #765

PR #765, opened 2021-11-23
Array copy

stephanmg:
Trying to address issue #345


yannikschaelte:
Could you shortly also address https://github.com/ICB-DCM/pyPESTO/runs/4297387617?check_suite_focus=true#step:6:23905 ? This is most likely due to a matplotlib update, which now requires actual integers in certain places. Then the tests should pass again.

pypesto/optimize/result.py
```diff
@@ -79,18 +79,18 @@ def __init__(self,
                  message: str = None):
         super().__init__()
         self.id = id
-        self.x: np.ndarray = np.array(x) if x is not None else None
+        self.x: np.ndarray = np.asarray(x) if x is not None else None
```
yannikschaelte, line 82: to be on the safe side results are not overwritten, I think I would enforce `np.array` in this function everywhere.

yannikschaelte (approved):

PR #769, opened 2021-12-05
Refactor startpoints; add gradient check

yannikschaelte:
* Allow checking gradients and resampling of startpoints.
* Refactor startpoint generation with clear class task assignments.

Breaking:
* Removal of option `startpoint_resample` and `assign_startpoints`, as these do not scale with potentially more flexible options demanded. The new implementation encapsulates all startpoint functionality in the corresponding class.
* `Problem.startpoint_method` got removed, as this just introduces duplicated functionality.
* `PetabImporter.create_problem` no longer inserts the startpoint method (see 2), so if the petab sampling function should be used directly, this would need to be manually done via passing `importer.create_startpoint_method()` to the minimize function. I think this is cleaner, and allows better configuration of how startpoints are generated. One should however consider passing `{lb|ub}_init` from petab to the problem.


yannikschaelte:
(sorry for the big diff, a little black may have hopped over problem.py and importer.py) :black_cat:

pypesto/petab/importer.py
```diff
@@ -130,29 +132,43 @@ def check_gradients(
...
+                    [
+                        np.all(
+                            (mode_df.rel_err.values < rtol)
+                            | (mode_df.abs_err.values < atol)
+                        ),
+                    ]
+                )
+                for mode_df in dfs
+            ]
+        )
```
dweindl, line 165: :sweat_smile:

pypesto/problem.py
```diff
@@ -95,7 +98,6 @@ def __init__(
         x_fixed_indices: Optional[SupportsIntIterableOrValue] = None,
         x_fixed_vals: Optional[SupportsFloatIterableOrValue] = None,
         x_guesses: Optional[Iterable[float]] = None,
-        startpoint_method: Optional[Callable] = None,
```
dweindl, line 98: I think it makes sense to have the startpoint outside of `Problem`, as it's rather a part of the algorithm for solving, than of the problem itself.

pypesto/startpoint/base.py
```diff
@@ -18,31 +19,189 @@ class StartpointMethod:
...
+            ret = objective(x, sensi_orders=sensi_orders, return_dict=True)
+            fvals[ix] = ret.get(FVAL, np.nan)
+
+            # loop until all requested sensis are finite
+            while True:
+                # discontinue if all requested sensis are finite
+                if (
+                    0 not in sensi_orders or ret[FVAL] < np.inf
+                ) and (
+                    1 not in sensi_orders or np.all(ret[GRAD] < np.inf)
```
dweindl, line 182:
```suggestion
                    0 not in sensi_orders or np.isfinite(ret[FVAL])
                ) and (
                    1 not in sensi_orders or np.all(np.finite(ret[GRAD]))
```

If we accept negative infinity, we should also accept positive infinity. Or rather neither?

test/base/test_startpoint.py
```diff
@@ -55,3 +56,74 @@ def test_ubounded_startpoints(spmethod):
...
+def check_fval(request) -> bool:
+    return request.param
+
+
+@pytest.fixture(params=[True, False])
+def check_grad(request) -> bool:
+    return request.param
+
+
+def test_resampling(check_fval: bool, check_grad: bool):
```
dweindl, line 71: Intended for use in additional tests? Otherwise `pytest.mark.parameterize` would be more concise.

dweindl (approved):
Looks good and cleaner than before :+1: 

> * Allow checking gradients and resampling of startpoints.

Thanks, very helpful!

pypesto/startpoint/base.py
dweindl: On second thought, probably makes sense to leave as is, unless somebody is optimizing f(x) = inf.

yannikschaelte: I guess requesting finite values should be fine, i.e. excluding {nan, inf, -inf}

pypesto/petab/importer.py
FFroehlich: Agree with Daniel here, is there a way to tell black that we want `all([` and `any([`?

pypesto/startpoint/base.py
```diff
@@ -18,31 +19,189 @@ class StartpointMethod:
...
+
+        Parameters
+        ----------
+        n_starts: Number of starts.
+        problem: Problem specifying e.g. dimensions, bounds, and guesses.
+
+        Returns
+        -------
+        xs: Startpoints, shape (n_starts, n_par).
+        """
```
FFroehlich, line 34: `NotImplementedError()`?

FFroehlich (commented):
lot's of scrolling :( but implenetation looks good! 🥇

FFroehlich (approved):

pypesto/startpoint/base.py
dweindl: To raise that, you'd have to make the method non-abstract. As this is an interface class never expected to provide an implementation, I'd prefer abstract.

FFroehlich: https://docs.python.org/3/library/exceptions.html#NotImplementedError ? I thought thats the standard python way of marking the method abstract.

yannikschaelte: afaik: Conceptually, `NotImplementedError` indicates that something has not (yet) been implemented, which can have various reasons, including lazyness of the developer, or abstracness. If the whole class however is to be considered abstract, `@abstractmethod` is preferable, which prevents any instantiation of the class in the first place until all abstract methods have been implemented (in derived classes).

Issue #743, opened 2021-10-14
Flaky test_result_from_hdf5_history?


dweindl:
Fails here https://github.com/ICB-DCM/pyPESTO/runs/3893112578?check_suite_focus=true

I don't think that's related to that commit.

```
________________________ test_result_from_hdf5_history _________________________

    def test_result_from_hdf5_history():
        problem = create_petab_problem()
        with tempfile.TemporaryDirectory(dir=".") as tmpdirname:
            _, fn = tempfile.mkstemp(".hdf5", dir=f"{tmpdirname}")
    
            history_options_hdf5 = pypesto.HistoryOptions(trace_record=True,
                                                          storage_file=fn)
            # optimize with history saved to hdf5
            result = pypesto.optimize.minimize(
                problem=problem, n_starts=1,
                history_options=history_options_hdf5)
    
            result_from_hdf5 = optimization_result_from_history(fn)
    
            # Currently 'exitflag', 'time' and 'message' are not loaded.
            arguments = [ID, X, FVAL, GRAD, HESS, RES, SRES,
                         N_FVAL, N_GRAD, N_HESS, N_RES, N_SRES, X0, FVAL0]
            for key in arguments:
                if result.optimize_result.list[0][key] is None:
                    assert result_from_hdf5.optimize_result.list[0][key] is None
                elif isinstance(result.optimize_result.list[0][key], np.ndarray):
>                   np.testing.assert_almost_equal(
                        result.optimize_result.list[0][key],
                        result_from_hdf5.optimize_result.list[0][key]
                    )
E                   AssertionError: 
E                   Arrays are not almost equal to 7 decimals
E                   
E                   Mismatched elements: 2 / 2 (100%)
E                   Max absolute difference: 0.0001968
E                   Max relative difference: 0.00014251
E                    x: array([-0.915745 , -7.9470828])
E                    y: array([-0.9156145, -7.946886 ])

test/base/test_store.py:424: AssertionError
```

PaulJonasJost:
yeah not too sure as to why this happens tbh, it should read and write the same values... normally works if you just run it again I think.

yannikschaelte:
I hope this has been fixed in 0.2.11. Something weird may or may not have remained, so let's keep an eye on whether tests fail again.

dweindl:
> Something weird may or may not have remained, so let's keep an eye on whether tests fail again.

Something weird has remained...

[...]

PR #770, opened 2021-12-06
Fix autosave function for single core

yannikschaelte:
* Assert {id} template in CSV histories.
* Fix behavior of filename=None for single-core by rewriting pre- and post-processing (filename was ignored / overwritten by None, and thus the default output file always generated)

Actual fix of #743 not done yet.


yannikschaelte:
(applied black to history.py. should stop doing that :black_cat: )

pypesto/objective/history.py
```diff
@@ -117,9 +143,31 @@ def __getattr__(self, key):
     __setattr__ = dict.__setitem__
     __delattr__ = dict.__delitem__
 
+    def _sanity_check(self):
```
yannikschaelte, line 146: this is the only actual change in this file.

pypesto/objective/history.py
```diff
@@ -154,18 +204,17 @@ def create_history(
...
-        elif type in ['.hdf5', '.h5']:
+        if type_ == '.csv':
+            return CsvHistory(x_names=x_names, file=storage_file, options=self)
+        elif type_ in ['.hdf5', '.h5']:
             return Hdf5History(id=id, file=storage_file, options=self)
         else:
             raise ValueError(
-                "Currently, only history storage to '.csv' and '.hdf5'"
-                "is supported")
+                "Only history storage to '.csv' and '.hdf5' is supported, got "
```
yannikschaelte, line 215: and a tiny change here

dweindl, line 215:
```suggestion
                "Only history storage to '.csv', '.h5' and '.hdf5' is supported, got "
```

test/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+
+@pytest.fixture
+def hdf5_file():
+    """Generate a temporary hdf5 file."""
+    store_file = tempfile.mkstemp(suffix=".hdf5")[1]
+    try:
+        yield store_file
+    finally:
+        if REMOVE_STORE and os.path.exists(store_file):
+            os.remove(store_file)
```
dweindl, line 18: I guess this is to avoid extra indentation by a tempfile context?

Could generalize that and pass the suffix as argument. Can also be used in other places then.

test/base/test_store.py
```diff
@@ -404,31 +427,32 @@ def test_storage_objective_config():
             os.remove(fn)
 
 
-def test_result_from_hdf5_history():
+def test_result_from_hdf5_history(hdf5_file):
     problem = create_petab_problem()
-    with tempfile.TemporaryDirectory(dir=".") as tmpdirname:
```
dweindl, line 409: Note that there are issues on Windows with reopening temporary files, and `TemporaryDirectory` is therefore often preferable. Is pypesto tested on Windows?

yannikschaelte:
> Not entirely sure what you did in the testing files but looks like a more elegant way of creating temporary hdf5 files +1

you mean that `hdf5_file` argument? That was just as a cleaner solution to those `with ...` statements, and to fight claustrophia.

test/base/test_store.py
yannikschaelte: nope, there are currently not tests on windows. one could just rewrite the fixture as 
```python
with tempfile.TemporaryDirectory() as tmpdir:
    file = os.path.join(tmpdir, "file.hdf5")
    yield file
```
?

dweindl: True. Fine then.

pypesto/optimize/optimize.py
```diff
@@ -27,7 +27,7 @@ def minimize(
     progress_bar: bool = True,
     options: OptimizeOptions = None,
     history_options: HistoryOptions = None,
-    filename: Union[str, None] = "Auto"
+    filename: str = "Auto"
```
FFroehlich, line 30: ☝️ #771

pypesto/optimize/optimize.py
```diff
@@ -119,13 +120,14 @@ def minimize(
     if engine is None:
         engine = SingleCoreEngine()
 
+    # maybe change to one hdf5 storage file per start if parallel
```
FFroehlich, line 123:
```suggestion
    # change to one hdf5 storage file per start if parallel
```

pypesto/objective/history.py
```diff
@@ -117,9 +143,31 @@ def __getattr__(self, key):
     __setattr__ = dict.__setitem__
     __delattr__ = dict.__delitem__
 
+    def _sanity_check(self):
+        """Apply basic sanity checks."""
+        if self.storage_file is None:
+            return
+
+        # extract storage type
+        _, type_ = os.path.splitext(self.storage_file)
```
FFroehlich, line 152: also use `pathlib.Path.suffix` here?

pypesto/optimize/util.py
```diff
@@ -1,88 +1,118 @@
...
+        Whether history storage post-processing is required.
     """
+    storage_file = history_options.storage_file
+
+    # nothing to do if no history stored
+    if storage_file is None:
+        return False
+
+    # extract storage type
+    path = Path(storage_file)
```
FFroehlich, line 42: 👍

FFroehlich (approved):

test/base/test_history.py
```diff
@@ -79,13 +86,13 @@ def check_history(self):
                 if storage_type is not None:
                     read_results_from_file(
                         self.problem, self.history_options,
-                        len(result.optimize_result.list) + 2
+                        n_starts=n_starts,
```
FFroehlich, line 89: As indicated by the docstring here, having n_starts+2 was very much intentional

yannikschaelte: I guess silently reading the same file multiple times bc no id is present, is not desirable, unless explicitly requested, thus the {id} check.

FFroehlich: but the `{id}` was there before. This is supposed to check whether read_results_from_file can handle missing files (optimization stopped before file was written, for whatever reason). If we do not wan't this behavior we should add a check whether `read_results_from_file` fails when the file is not present.

yannikschaelte: how did the {id} come in before? It's not in the original file name in https://github.com/ICB-DCM/pyPESTO/commit/6d3226d6ccf841ad4c8f5cb77ca59f0fe79cd514#diff-80b64c062a6e89e507e7518b03d2cf45dc9518a824f39bae232b565e8e5016afL52, does it come in in processing somewhere?

yannikschaelte: I thought it just read in the same file filename.csv 3 times in the previous test implementation, therefore all files existed. when adding {id}, then there was a trace is Null error or something.

FFroehlich: will fix this in a separate PR

yannikschaelte: thanks :) it confused me too then. so in conclusion I guess yes we want to be able to merge histories also when some ids are not present (maybe warn then).

PR #771, opened 2021-12-06
Fix typehint for 'filename' in optimize.minimize

dweindl:
`None` is required to switch off auto-saving, but not allowed as per typehint


PaulJonasJost (approved):
Thanks 👍

PR #777, opened 2021-12-13
make AmiciObjective only report whats being asked for

### LINKED ISSUE #776: AggregateObjective evaluation fails for `MODE_RES` and `sensi_orders=1`
AmiciObjective returns CHI2, RES, SRES, but prior (correctly) only returns SRES, leading to failure when aggregating.

FFroehlich:
fixes #776

### DIFF
--- a/pypesto/objective/amici_calculator.py
+++ b/pypesto/objective/amici_calculator.py
@@ -1,5 +1,5 @@
 import numpy as np
-from typing import Dict, List, Sequence, Union
+from typing import Dict, List, Sequence, Union, Tuple
 
 from .constants import (
     MODE_FUN, MODE_RES, FVAL, GRAD, HESS, RES, SRES, RDATAS, CHI2
@@ -33,7 +33,7 @@ class AmiciCalculator:
 
     def __call__(self,
                  x_dct: Dict,
-                 sensi_order: int,
+                 sensi_orders: Tuple[int],
                  mode: str,
                  amici_model: AmiciModel,
                  amici_solver: AmiciSolver,
@@ -50,8 +50,8 @@ class AmiciCalculator:
         ----------
         x_dct:
             Parameters for which to compute function value and derivatives.
-        sensi_order:
-            Maximum sensitivity order.
+        sensi_orders:
+            Tuple of requested sensitivity orders.
         mode:
             Call mode (function value or residual based).
         amici_model:
@@ -71,6 +71,7 @@ class AmiciCalculator:
             requested).
         """
         # set order in solver
+        sensi_order = max(sensi_orders)
         if sensi_order == 2 and fim_for_hess:
             # we use the FIM
             amici_solver.setSensitivityOrder(sensi_order-1)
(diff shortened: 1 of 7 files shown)


pypesto/objective/amici_util.py
```diff
@@ -286,7 +286,7 @@ def get_error_output(
         amici_model: AmiciModel,
         edatas: Sequence['amici.ExpData'],
         rdatas: Sequence['amici.ReturnData'],
-        sensi_order: int,
+        sensi_orders: Tuple[int],
```
yannikschaelte, line 289: I think `Tuple[int, ...]` would be more correct https://docs.python.org/3/library/typing.html#typing.Tuple

yannikschaelte (approved):

pypesto/objective/amici_calculator.py
```diff
@@ -71,6 +71,11 @@ def __call__(self,
             requested).
         """
         # set order in solver
+        if sensi_orders:
+            sensi_order = max(sensi_orders)
+        else:
+            sensi_order = 0
+
```
PaulJonasJost, line 78: If sensi_order is only used in the amici_solver.setSensitivityOrder, would perhaps do it like in the other changes with 
```python
if 2 in sense_orders:
...
```

pypesto/objective/amici_calculator.py
```diff
@@ -108,14 +113,14 @@ def __call__(self,
...
-            rdatas=rdatas, sensi_order=sensi_order, mode=mode,
+            rdatas=rdatas, sensi_orders=sensi_orders, mode=mode,
             amici_model=amici_model, amici_solver=amici_solver, edatas=edatas,
             x_ids=x_ids, parameter_mapping=parameter_mapping,
             fim_for_hess=fim_for_hess)
 
 
 def calculate_function_values(rdatas,
-                              sensi_order: int,
+                              sensi_orders: Tuple[int],
```
PaulJonasJost, line 123: Here then just as yannik suggested also `Tuple[int, ...]`?

pypesto/objective/amici_util.py
```diff
@@ -320,7 +320,10 @@ def get_error_output(
     return filter_return_dict(ret)
 
 
-def init_return_values(sensi_order, mode, dim, error=False):
+def init_return_values(sensi_orders: Tuple[int],
```
PaulJonasJost, line 323: Here `Tuple[int, ...]`as well then.

PaulJonasJost (approved):
I am not too sure how convenient it is, having to pass on all wanted sensitivities together. Is there a case where I would want second order sensitivities but not first order?

pypesto/objective/amici_calculator.py
FFroehlich: Amici will always only compute lower sensitivity orders, as they come more or less for free. For other objectives that might not be the case, hence the distinction.

FFroehlich:
> I am not too sure how convenient it is, having to pass on all wanted sensitivities together. Is there a case where I would want second order sensitivities but not first order?

For AmiciObjective I agree it does not make too much sense, but that's the way we implemented the API for general objectives and thats how, e.g., priors behave.

PR #779, opened 2021-12-17
Feature separate likelihood and prior

jvanhoefer:
- [x] add `get_negloglikelihood` and `get_neglogposterior` in all the different objectives.
- [x] Go through samplers, and adapt them, if null & prior are needed separately
- [x] remove `problem.x_priors`
- [x] Adapt tests...


pypesto/objective/aesara.py
```diff
@@ -90,6 +91,41 @@ def __init__(self,
...
+        **kwargs,
+    ) -> Union[float, np.ndarray, Tuple, ResultDict]:
+        """
+        Negative logprior.
+
+        As the Objective is interpreted as negative loglikelihood, the prior
+        is the "trivial" prior.
+        """
+        if return_dict:
+            get_zero_result_dict(x, sensi_orders, mode)
```
FFroehlich, line 125: the base objective may have a prior, so I think this function should compute input to the base function and then call `get_neglogprior` on the base function. Same for likelihood. Might be easiest to implement this through an additional argument that can be passed to `__call__` to avoid code duplication.

jvanhoefer: good point. :+1: Had a closer look again, and just calling the corresponding methods on `self.base_objective` should resolve that issue.

FFroehlich: but `self.base_objective` needs `self.infun(x)` as input, not `x`. Also `inner_ret` needs to be populated. Happy to provide an implementation.

pypesto/objective/amici.py
```diff
@@ -371,6 +373,41 @@ def call_unprocessed(
...
+        """
+        Negative logprior.
+
+        As the AMICI objective returns the negative loglikelihood, the prior
+        is the "trivial" prior.
+        """
+        if return_dict:
+            get_zero_result_dict(x, sensi_orders, mode)
+        else:
+            return 0.0
```
dilpath, line 409:
```suggestion
        if return_dict:
            return get_zero_result_dict(x, sensi_orders, mode)
        return 0.0
```

pypesto/objective/finite_difference.py
```diff
@@ -377,6 +378,40 @@ def call_unprocessed(
...
+        """
+        Negative logprior.
+
+        As the objective is interpreted as negative loglikelihood, the prior
+        is the "trivial" prior.
+        """
+        if return_dict:
+            get_zero_result_dict(x, sensi_orders, mode)
+        else:
+            return 0.0
```
dilpath, line 413:
```suggestion
        if return_dict:
            return get_zero_result_dict(x, sensi_orders, mode)
        return 0.0
```

pypesto/objective/function.py
```diff
@@ -149,6 +150,40 @@ def call_unprocessed(
...
+        """
+        Negative logprior.
+
+        As the function is interpreted as negative loglikelihood, the prior
+        is the "trivial" prior.
+        """
+        if return_dict:
+            return get_zero_result_dict(x, sensi_orders, mode)
+        else:
+            return 0.0
```
dilpath, line 185:
```suggestion
        if return_dict:
            return get_zero_result_dict(x, sensi_orders, mode)
        return 0.0
```

pypesto/objective/priors.py
```diff
@@ -108,6 +109,31 @@ def call_unprocessed(
...
+        sensi_orders: Tuple[int, ...] = (0, ),
+        mode: str = MODE_FUN,
+        return_dict: bool = False,
+        **kwargs,
+    ) -> Union[float, np.ndarray, Tuple, ResultDict]:
+        """Negative loglikelihood, returns 0."""
+        if return_dict:
+            get_zero_result_dict(x, sensi_orders, mode)
+        else:
+            return 0.0
```
dilpath, line 124:
```suggestion
        if return_dict:
            return get_zero_result_dict(x, sensi_orders, mode)
        return 0.0
```

pypesto/objective/aesara.py
jvanhoefer: Thank you for your offer, that would be great! :)

FFroehlich: done

jvanhoefer: Thank you :)

PaulJonasJost:
Closing this since due to prolonged inactivity. Feel free to restore branch, if further work is planned.

PR #782, opened 2021-12-22
Black+Isort

yannikschaelte:
given the number of contributors, and to avoid having to handle minor flake8- or own-style inconsistencies in prs, shall we have an external party (=black) decide about code style?

black has been integrated via a pre-commit hook (e.g. run via `pre-commit run --all-files` or automatically in every commit via once `pre-commit install`)

similarly, isort has been integrated for import sorting

manually changed files:
* `pyproject.toml`
* `.flake8`
* `.pre-commit-config.yaml`
* `__init__.py`

this is the continuation of #757, which included too many changes


yannikschaelte:
we have a little problem with circular imports ... not sure if I wanna solve them now

PR #783, opened 2021-12-23
allow loading of partial results from history file

FFroehlich:


yannikschaelte (approved):
So that additional file will just not exist and it therefore raise a warning. :+1:

PR #784, opened 2021-12-23
Create result sub-module; shift storage+result related functionality around

yannikschaelte:
"Result" and any sub-object is a data exchange format and should thus be in its own upstream module. Otherwise depends on right sorting or unchecking of imports.


pypesto/optimize/io.py
```diff
@@ -0,0 +1,221 @@
+"""Load/reconstitute results from history."""
```
FFroehlich, line 1: wouldn't it be more consistent to also have this in `pypesto/result/load.py` or even `pypesto/result/optimize.py`. Technically no `o` here, only `i`, so I would still consider renaming even if it stays in this dir.

FFroehlich (approved):
👍

pypesto/optimize/io.py
yannikschaelte: yeah, one could also move all io functionality to store/result. will leave for the moment as it does not disturb, renaming to "load.py" (was called that before actually).

PR #785, opened 2021-12-23
Apply isort

yannikschaelte:
* Apply isort via pre-commit hook to sort all imports

(i.e. just automatic formatting changes, more or less no actual changes in this pr)

* limit number of function evaluations in tests via #787


pypesto/C.py
```diff
@@ -0,0 +1,128 @@
+"""
+Constants
+=========
+
+Package-wide consistent constant definitions.
+
+"""
+
+from enum import Enum
```
FFroehlich, line 9: how about turning all of these into proper enums? I think with multiple inheritance from `str` and `Enum` we should maintain backwards compatibility.

Would keep imports smaller and enable better input checking.

FFroehlich (changes requested):
image

can we keep things in separate (minimal) PRs if >100 files are touched? This is an absolute pain to review.

pypesto/C.py
yannikschaelte: What for example would you mean? Regarding brevity of imports, I had in mind anyhow only importing `from pypesto import C` and then using e.g. `C.MODE_FUN`, which should be nice and readable already.

yannikschaelte: > image
> 
> can we keep things in separate (minimal) PRs if >100 files are touched? This is an absolute pain to review.

Yeah ... the C + refactoring of the ensemble module became necessary along the way, but will try to extract the actual code changes there, to separate from the mere sorting.

FFroehlich (approved):

PR #788, opened 2021-12-28
Refactor Ensemble module; introduce general C constants file

yannikschaelte:
* Introduce generic module C for package-wide constants
* Fix cyclic dependency in ensemble module (ensemble <-> util)
* Move prediction results to result module

Reason:

Should now be not randomly breaking due to cyclic imports
C.{CONSTANT} is easier to call (java/android convention)
Some submodules were redefining half of the constants to the same value, which is undesirable.

This has been extracted from #785


pypesto/C.py
```diff
@@ -0,0 +1,127 @@
...
+
+TIME = 'time'  # time
+N_FVAL = 'n_fval'  # number of function evaluations
+N_GRAD = 'n_grad'  # number of gradient evaluations
+N_HESS = 'n_hess'  # number of Hessian evaluations
+N_RES = 'n_res'  # number of residual evaluations
+N_SRES = 'n_sres'  # number of residual sensitivity evaluations
+CHI2 = 'chi2'  # chi2 value
+SCHI2 = 'schi2'  # chi2 value gradient
+X = 'x'
```
dilpath, line 33: Might conflict with usage of `X` as a vector of `x` in some scripts

dilpath: If done like PEtab constants, then this might be renamed more informatively as e.g. `PARAMETER_VECTOR` (but currently consistent with the other shortened constant names, so fine as is / out of scope...)

pypesto/C.py
```diff
@@ -0,0 +1,127 @@
...
+
+PREDICTOR = 'predictor'
+PREDICTION_ID = 'prediction_id'
+PREDICTION_RESULTS = 'prediction_results'
+PREDICTION_ARRAYS = 'prediction_arrays'
+PREDICTION_SUMMARY = 'prediction_summary'
+
+HISTORY = 'history'
+OPTIMIZE = 'optimize'
+SAMPLE = 'sample'
```
dilpath, line 80: Perhaps belongs in a different section

FFroehlich (approved):

pypesto/C.py
```diff
@@ -0,0 +1,127 @@
...
+"""
+
+from enum import Enum
+
+from typing import Tuple, Union
+
+###############################################################################
+# OBJECTIVE
+
+MODE_FUN = 'mode_fun'  # mode for function values
```
FFroehlich, line 14: Might be worth reformatting this way:

```
class ObjectiveMode(enum.Enum, str):
    MODE_FUN = 'mode_fun' # function values
    MODE_RES = 'mode_res' # residuals
```

makes import easier, better input checking, etc.
should also address some of the issues pointed out below.

pypesto/C.py
yannikschaelte: definitely a good point, we should consider

pypesto/C.py
```diff
@@ -0,0 +1,127 @@
...
+
+COLOR_HIT_BOTH_BOUNDS = [0.6, 0., 0., 0.9]
+COLOR_HIT_ONE_BOUND = [0.95, 0.6, 0., 0.9]
+COLOR_HIT_NO_BOUNDS = [0., 0.8, 0., 0.9]
+
+
+class EnsembleType(Enum):
+    """Specifies different ensemble types."""
+
+    ensemble = 1
```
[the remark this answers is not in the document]
yannikschaelte: didn't want to mess here, didn't check what it's used for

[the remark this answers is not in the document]
yannikschaelte: if easy to change, definitely sounds good

pypesto/C.py
yannikschaelte: would mess with a couple more files here, but sound good for another pr

PR #792, opened 2022-01-11
Tiny refactoring of priors; fix github repo download in tests

yannikschaelte:
* No actual changes
* Fix github authentication error in tests


pypesto/objective/priors.py
```diff
@@ -1,8 +1,9 @@
 from copy import deepcopy
-from typing import Callable, Dict, List, Sequence, Tuple
+from typing import Callable, Dict, List, Sequence, Tuple, Union
 
 import numpy as np
 
+from .. import C
 from ..C import CHI2, FVAL, GRAD, HESS, MODE_FUN, MODE_RES, RES, SRES
```
PaulJonasJost, line 7: Would either do only the first or only the second kind. Not both.

PaulJonasJost (approved):

PR #799, opened 2022-01-18
disable progress bar in tests

### LINKED ISSUE #791: Disable progress bar in most tests
Output tracking does not gracefully deal with line rollback apparently, and adds another line upon each update. To avoid resulting overly long logfiles, it may be preferable to disable progress bars per default / in tests.

PaulJonasJost:
closes #791

### DIFF
--- a/pypesto/sample/parallel_tempering.py
+++ b/pypesto/sample/parallel_tempering.py
@@ -51,6 +51,7 @@ class ParallelTemperingSampler(Sampler):
             'max_temp': 5e4,
             'exponent': 4,
             'temper_log_posterior': False,
+            'show_progress': True,
         }
 
     def initialize(
@@ -69,8 +70,10 @@ class ParallelTemperingSampler(Sampler):
 
     def sample(self, n_samples: int, beta: float = 1.0):
         """Sample and swap in between samplers."""
+        show_progress = self.options['show_progress']
         # loop over iterations
-        for i_sample in tqdm(range(int(n_samples))):  # TODO test
+        for i_sample in tqdm(range(int(n_samples)), disable=not show_progress):
+            # TODO test
             # sample
             for sampler, beta in zip(self.samplers, self.betas):
                 sampler.sample(n_samples=1, beta=beta)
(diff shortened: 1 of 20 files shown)


test/optimize/test_optimize.py
```diff
@@ -258,7 +258,7 @@ def get_optimizer(library, solver):
     elif library == 'fides':
         options[fides.Options.SUBSPACE_DIM] = solver[1]
         optimizer = optimize.FidesOptimizer(
-            options=options, hessian_update=solver[0], verbose=0
+            options=options, hessian_update=solver[0], verbose=40
```
yannikschaelte, line 261: I guess we'll want `verbose=logging.WARNING` or `logging.ERROR` or whatever 40 is.

PaulJonasJost: yes I think we changed it at the same time :D

pypesto/sample/parallel_tempering.py
```diff
@@ -69,8 +70,10 @@ def initialize(
 
     def sample(self, n_samples: int, beta: float = 1.0):
         """Sample and swap in between samplers."""
+        show_progress = self.options['show_progress']
         # loop over iterations
-        for i_sample in tqdm(range(int(n_samples))):  # TODO test
+        for i_sample in tqdm(range(int(n_samples)), disable=not show_progress):
+            # TODO test
```
jvanhoefer, line 76: Is this still ToDo?

jvanhoefer:
looks good!

pypesto/sample/parallel_tempering.py
PaulJonasJost: not sure, was there before. Would probably have to ask @yannikschaelte

test/optimize/test_optimize.py
yannikschaelte: ah, I meant literally writing the logging.XYZ, which is just an int of value 40, but improves readability.

pypesto/sample/parallel_tempering.py
yannikschaelte: yeah I guess we wanted to add tests there, tbd independent of current pr

PR #803, opened 2022-01-19
removed ftol in favor of tol. Added it to leas squares to keep same f…

PaulJonasJost:
removed ftol from script optimizer options as many solvers are not able to make use of it. Instead used `tol`.


jvanhoefer (approved):

yannikschaelte (approved):
I would change to `tol: float = None` in l. 270 too. @FFroehlich opinions?

FFroehlich:
> I would change to `tol: float = None` in l. 270 too. @FFroehlich opinions?

sounds good

PR #804, opened 2022-01-20
Refactor sample result

jvanhoefer:
The sample result was never used, while the `McmcPtResult` did not ensure a stable interface to the outside world. Fixed that. Might not be the final solution to the problem, but should definitely be an improvement


pypesto/result/sample.py
```diff
@@ -5,19 +5,90 @@
...
+        """Allow usage of keys like attributes."""
+        if key == 'n_samples':
+            raise ValueError(
+                "n_samples can not be set, as they are computed "
+                "from self.trace_x."
+            )
+        self[key] = value
+
+    def __iter__(self):
+        self._index_iter = self.burn_in
```
yannikschaelte, line 64: should this variable be initialized in init? (e.g. if pycharm flags it)

pypesto/result/sample.py
```diff
@@ -5,19 +5,90 @@
...
+
+    def __next__(self):
+        # return next sample
+        if self._index_iter < self.n_samples:
+            sample = self.trace_x[self._index_iter, :]
+            self._index_iter += 1
+            return sample
+        else:
+            # clean up and terminate
+            del self._index_iter
```
yannikschaelte, line 75: should not be necessary, just set to None or something (see prev comment)

pypesto/result/sample.py
```diff
@@ -48,18 +119,31 @@ def __init__(
...
+            trace_neglogpost,
+            trace_neglogprior,
+            burn_in,
+            time,
+            auto_correlation,
+            effective_sample_size,
+            message,
+        )
+
+        self._check_trace_dimensions(
```
yannikschaelte, line 133: yep you call it as a non-static method (see above)

pypesto/result/sample.py
```diff
@@ -100,19 +184,16 @@ def __init__(
...
     __setattr__ = dict.__setitem__
     __delattr__ = dict.__delitem__
 
-
-class SampleResult:
-    """Result of the sample() function."""
-
-    def __init__(self):
-        pass
+    def __next__(self):
```
yannikschaelte, line 190: this looks like a duplicate

pypesto/result/sample.py
```diff
@@ -5,19 +5,90 @@
 import numpy as np
 
 
-class McmcPtResult(dict):
+class SampleResultBase(dict):
```
yannikschaelte, line 8: add abc.Iterable as base class

pypesto/result/sample.py
```diff
@@ -5,19 +5,90 @@
...
+        self[key] = value
+
+    def __iter__(self):
+        self._index_iter = self.burn_in
+        return self
+
+    def __next__(self):
+        # return next sample
+        if self._index_iter < self.n_samples:
+            sample = self.trace_x[self._index_iter, :]
```
yannikschaelte, line 70: I am not sure whether this simple iteration makes sense -- what if one wants to iterate over function values? which will happen often.

yannikschaelte, line 70: I am not sure whether this simple iteration makes sense -- what if one wants to iterate over function values? which will happen often.

yannikschaelte (changes requested):

pypesto/result/sample.py
```diff
@@ -5,19 +5,90 @@
...
+        trace_x: np.ndarray = None,
+        trace_neglogpost: np.ndarray = None,
+        trace_neglogprior: np.ndarray = None,
+        burn_in: int = None,
+        time: float = 0.0,
+        auto_correlation: float = None,
+        effective_sample_size: float = None,
+        message: str = None,
+        **kwargs,
+    ):
```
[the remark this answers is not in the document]
jvanhoefer: yep, will do that via a doctoring for the whole class (similar as it is currently in `McmcPtResult`)

pypesto/result/sample.py
jvanhoefer: would say no, as this is only used during iterating and deleted afterwards... ?

pypesto/result/sample.py
jvanhoefer: Nope, as you have another indexing due to different chains... :/

pypesto/result/sample.py
jvanhoefer: True, also a possible use case. But for this use case this is also no disadvantage either...

yannikschaelte: I would say that specialized iteration for x is not necessary then. the default use case would probably be anyway that one assesses the whole matrix, e.g. to compute expectations, for which vectorization on the full object is better.

pypesto/result/sample.py
```diff
@@ -5,19 +5,113 @@
...
+    burn_in: int
+        The burn in index.
+    time: float
+        The computation time.
+    auto_correlation:
+        The estimated autcorrelation.
+    effective_sample_size:
+        The estimated effective sample size.
+    message: str
+        Textual comment on the profile result.
```
yannikschaelte, line 32:
```suggestion
        Textual comment on the sampling result.
```

pypesto/result/sample.py
```diff
@@ -5,19 +5,113 @@
...
+    def __setattr__(self, key, value):
+        """Allow usage of keys like attributes."""
+        if key == 'n_samples':
+            raise ValueError(
+                "n_samples cannot be set, as they are computed "
+                "from self.trace_x."
+            )
+        self[key] = value
+
+    def __iter__(self):
```
yannikschaelte, line 86: would remove `__iter__` and `__next__` as I don't see the special usecase of iterating over x, but not fval, and using one x at a time (in contrast to slicing to a single parameter, e.g. for plotting, which would also be done via `trace_x[:, ix]`, but also fine to leave if preferred.

yannikschaelte (approved):

pypesto/result/sample.py
```diff
@@ -48,71 +142,81 @@ def __init__(
         effective_sample_size: float = None,
         message: str = None,
     ):
-        super().__init__()
+        super().__init__(
+            trace_x,
```
yannikschaelte, line 146: I usually prefer kwargs in long arg lists, as otherwise assuring correct order may be prone to future errors.

pypesto/result/sample.py
jvanhoefer: :D copied from the original code, which apparently copied it :D

PaulJonasJost (approved):

yannikschaelte:
@jvanhoefer what's the status here?

PaulJonasJost:
@jvanhoefer what is the status on this?

PaulJonasJost:
Closing this since due to prolonged inactivity. Feel free to restore branch, if further work is planned.

PR #807, opened 2022-01-24
Make Fides work with objectives, that do not have a hessian closes #805

### LINKED ISSUE #805: Fides default hessian updated makes optimization fail, if objective does not provide a hessian
**Bug description**
If one uses the default fides optimizer from `fides_optimizer = optimize.FidesOptimizer()` with an objective, that does not provide a Hessian, optimization (or rather the Hessian update) fails.

**Expected behavior**
Set a default value, that works also without gradient.

jvanhoefer:


FFroehlich:
BFGS simply does not perform quite as well on many problems, thus I don't think it is a good default. Maybe only set BFGS as default when `has_hess` is False?

jvanhoefer:
@FFroehlich Good point, fixed that. Would you be fine with that solution? Or prefer an entirely different approximation?

pypesto/optimize/optimizer.py
```diff
@@ -1105,11 +1105,13 @@ def __init__(
         """
         super().__init__()
 
-        if hessian_update == 'Hybrid':
-            hessian_update = fides.HybridFixed()
+        if hessian_update == 'default':
+            hessian_update = 'default'
```
dilpath, line 1109:
```suggestion
```

FFroehlich:
> @FFroehlich Good point, fixed that. Would you be fine with that solution? Or prefer an entirely different approximation?

No looks, good! May be worthwhile documenting the behavior in the docstr and/or issuing a warning when BFGS is used.

FFroehlich (approved):

pypesto/optimize/optimizer.py
```diff
@@ -1087,7 +1087,7 @@ class FidesOptimizer(Optimizer):
 
     def __init__(
         self,
-        hessian_update: Optional['HessianApproximation'] = 'Hybrid',
+        hessian_update: Optional['HessianApproximation'] = 'default',
```
yannikschaelte, line 1090: I would simply have `None` as default, not a string `"default"`. s.t. None means use whatever you have and approximation otherwise.

jvanhoefer: As far as I understand fides correctly, `None` has a special meaning (i.e. handling a `None` Hessian approximation is something different as what we intend here...). Therefore we need that...

PR #808, opened 2022-01-27
Waterfall zoom

plakrisenko:
numpy version was fixed, because of the problems with Jupyter  notebooks 
https://stackoverflow.com/questions/70839312/module-numpy-distutils-config-has-no-attribute-blas-opt-info


pypesto/visualize/waterfall.py
```diff
@@ -23,6 +93,7 @@ def waterfall(
     scale_y: Optional[str] = 'log10',
     offset_y: Optional[float] = None,
     start_indices: Optional[Union[Sequence[int], int]] = None,
+    n_starts_to_zoom: int = 10,
```
PaulJonasJost, line 96:
```suggestion
    n_starts_to_zoom: int = 0,
```

Perhaps the default should be not to zoom in?

test/visualize/test_visualize.py
```diff
@@ -109,10 +109,9 @@ def create_optimization_result():
             grad=np.array([2.5 + j + 0.1, 2 + j + 1]),
         )
         result.optimize_result.append(optimizer_result=optimizer_result)
-    for j in range(0, 4):
+    for j in range(0, n):
         optimizer_result = pypesto.OptimizerResult(
-            id=str(j + 3),
-            fval=10 + j * 0.01,
+            id=str(j + 3), fval=10 + j * 0.01,
```
dilpath, line 114: Might be interpreted as a complex number (i.e. might be better to use a different name for the index, instead of `j`).

setup.cfg
```diff
@@ -47,7 +47,7 @@ keywords =
 
 [options]
 install_requires =
-    numpy >= 1.19.1
+    numpy == 1.20.3
```
PaulJonasJost, line 50:
```suggestion
    numpy >= 1.19.1, <= 1.21.5
```

PaulJonasJost (approved):

jvanhoefer (approved):

Issue #710, opened 2021-07-22
Select unique ids also when a result object is handed over to optimization


jvanhoefer:
**Bug description**
Currently in the `minimize` function does not check for uniqueness in `ids`, if a `result` object is provided to the function. This can lead to problems downstream.

See the lines following
https://github.com/ICB-DCM/pyPESTO/blob/72488fbb3eaa91dd163f88bac71a1a165a0da70f/pypesto/optimize/optimize.py#L95

**Expected behavior**
Check, if `results is not None`, and only choose "new" ids.

@vanako had issues related to that problem.

PR #815, opened 2022-02-01
Append method for OptimizeResult

PaulJonasJost:
Currently if I append a result with e.g. 10 runs to one with 5, only the first 5 of the results to be appended get a new id. Is this in general a problem?

### DIFF
--- a/pypesto/optimize/util.py
+++ b/pypesto/optimize/util.py
@@ -158,19 +158,22 @@ def assign_ids(
     n_starts:
         Number of starts of the optimizer.
     """
-    used_ids = set({})
+    used_ids = set()
     if result is not None:
-        used_ids = result.optimize_result.id
+        used_ids = set(result.optimize_result.id)
+        N_used = len(used_ids)
     if ids is None:
         i = 0
         ids = [str(j) for j in range(i * n_starts, (i + 1) * n_starts)]
         while not used_ids.isdisjoint(ids):
             i += 1
-            ids = [str(j) for j in range(i * n_starts, (i + 1) * n_starts)]
+            ids = [
+                str(j) for j in range(N_used * i, N_used * (i + 1) + n_starts)
+            ]
     if len(ids) != n_starts:
         raise AssertionError("Number of starts and ids must coincide.")
     if not used_ids.isdisjoint(ids):
         raise AssertionError(
-            "Manually assigned ids must differ from " "existing ones."
+            "Manually assigned ids must differ from existing ones."
         )
     return ids
--- a/pypesto/optimize/util.py
+++ b/pypesto/optimize/util.py
@@ -3,13 +3,14 @@
 import logging
 import os
 from pathlib import Path
-from typing import List
+from typing import Iterable, List
 
 import h5py
 
 from ..C import PYPESTO_MAX_N_STARTS
 from ..engine import Engine, SingleCoreEngine
 from ..objective import HistoryOptions
+from ..result import Result
 from ..store.save_to_hdf5 import get_or_create_group
 from .optimizer import OptimizerResult
 
@@ -138,3 +139,38 @@ def bound_n_starts_from_env(n_starts: int):
     )
 
     return n_starts_new
+
+
+def assign_ids(
+    ids: Iterable[str] = None,
+    result: Result = None,
+    n_starts: int = 100,
+) -> Iterable[str]:
+    """
+    Assign ids to starts.
+
+    ids:
+        Ids assigned to the startpoints.
+    result:
+        A result object to append the optimization results to. For example,
+        one might append more runs to a previous optimization. Assign only
+        unique ids.
+    n_starts:
+        Number of starts of the optimizer.
+    """
+    used_ids = set({})
+    if result is not None:
+        used_ids = result.optimize_result.id
+    if ids is None:
+        i = 0
+        ids = [str(j) for j in range(i * n_starts, (i + 1) * n_starts)]
+        while not used_ids.isdisjoint(ids):
+            i += 1
+            ids = [str(j) for j in range(i * n_starts, (i + 1) * n_starts)]
+    if len(ids) != n_starts:
+        raise AssertionError("Number of starts and ids must coincide.")
+    if not used_ids.isdisjoint(ids):
+        raise AssertionError(
+            "Manually assigned ids must differ from " "existing ones."
+        )
+    return ids
--- a/pypesto/result/optimize.py
+++ b/pypesto/result/optimize.py
@@ -2,13 +2,15 @@
 
 import warnings
 from copy import deepcopy
-from typing import Sequence
+from typing import Sequence, Union
 
 import numpy as np
 import pandas as pd
 
 from ..objective import History
 from ..problem import Problem
+
+OptimizationResult = Union['OptimizerResult', 'OptimizeResult']
 
 
 class OptimizerResult(dict):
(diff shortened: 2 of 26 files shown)


pypesto/result/optimize.py
```diff
@@ -223,3 +232,18 @@ def get_for_key(self, key) -> list:
...
+
+    def _assign_unique_id(self, id: str, suffix: int = None):
+        """
+        Assign a unique id to an id if needed.
+
+        Returns id if not used already otherwise tries id_suffix with
+        increasing suffix until it is unique.
+        """
+        if suffix is None:
+            if id not in self.id:
```
dilpath, line 244: Where is `self.id` defined?

pypesto/result/optimize.py
```diff
@@ -162,20 +164,27 @@ def __getitem__(self, index):
...
+            The result of one or more (local) optimizer run.
+        to_sort:
+            Boolean used so we only sort once when appending an
+            optimize_result.
         """
-        self.list.append(optimizer_result)
-        self.sort()
+        if isinstance(opt_result, OptimizeResult):
+            for optimizier_result in opt_result.list:
+                self.append(optimizier_result, to_sort=False)
```
dilpath, line 181:
```suggestion
            for optimizer_result in opt_result.list:
                self.append(optimizer_result, to_sort=False)
```

pypesto/result/optimize.py
```diff
@@ -162,20 +164,27 @@ def __getitem__(self, index):
...
 
         Parameters
         ----------
-        optimizer_result:
-            The result of one (local) optimizer run.
+        opt_result:
+            The result of one or more (local) optimizer run.
+        to_sort:
+            Boolean used so we only sort once when appending an
+            optimize_result.
```
dilpath, line 177: Can rename `to_sort` to `sort`

pypesto/result/optimize.py
```diff
@@ -162,20 +164,27 @@ def __getitem__(self, index):
...
         """
-        self.list.append(optimizer_result)
-        self.sort()
+        if isinstance(opt_result, OptimizeResult):
+            for optimizier_result in opt_result.list:
+                self.append(optimizier_result, to_sort=False)
+        elif isinstance(opt_result, OptimizerResult):
+            opt_result.id = self._assign_unique_id(opt_result.id)
+            self.list.append(opt_result)
+            self.sort()
```
dilpath, line 185: Remove? Already handled by `to_sort`?

pypesto/result/optimize.py
```diff
@@ -162,20 +164,27 @@ def __getitem__(self, index):
     def __len__(self):
         return len(self.list)
 
-    def append(
-        self,
-        optimizer_result: OptimizerResult,
-    ):
+    def append(self, opt_result: OptimizationResult, to_sort: bool = True):
```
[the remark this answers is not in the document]
dilpath: And default `id_prefix` to e.g. some short, random hash

pypesto/result/optimize.py
PaulJonasJost: in recent changes it was changed so that now `self.id` is equal to `self.get_for_key(id)`

pypesto/result/optimize.py
PaulJonasJost: thought that would be misleading together with self.sort()

dilpath: Or, less misleading to use the same name :see_no_evil:

pypesto/result/optimize.py
dilpath: Could you link me to where it's defined? Can't find it in this class.

PaulJonasJost: https://github.com/ICB-DCM/pyPESTO/blob/4d737237ed85dfb9f051af6a1331e85666f62f24/pypesto/result/optimize.py#L145

dilpath: Thanks!

pypesto/result/optimize.py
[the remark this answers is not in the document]
dilpath: As I guess it would be nice for a user to specify the "set" of optimizations that an `OptimizerResult` belongs to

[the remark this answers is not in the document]
PaulJonasJost: _assign_unique_id is called recursively and will make sure the id is unique. Not sure how you would intend to do it 🤔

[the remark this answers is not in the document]
dilpath: Could let a user specify a prefix (default to empty string) (or suffix as done, but I guess prefix would be more conventional?), and raise an exception if the prefix+id already exists.

dilpath (approved):

pypesto/result/optimize.py
```diff
@@ -164,18 +166,46 @@ def __len__(self):
...
+        elif isinstance(optimize_result, OptimizerResult):
+            new_id = prefix + optimize_result.id
+            if new_id in current_ids:
+                raise ValueError(
+                    "The id you want to merge coincides with "
+                    "the existing id's. Please use an "
+                    "appropriate prefix such as 'run_2_'."
+                )
+            optimize_result.id = new_id
+            self.list.append(optimize_result)
```
dilpath, line 206:
```suggestion
        if isinstance(optimize_result, OptimizeResult):
            for optimizer_result in optimize_result.list:
                self.append(optimizer_result, sort=False, prefix=prefix)
        elif isinstance(optimize_result, OptimizerResult):
            new_id = prefix + optimize_result.id
            if new_id in self.id:
                raise ValueError(
                    "An ID you want to merge coincides with "
                    "the existing IDs. Please set the `prefix` argument"
                    "to e.g. 'run_2_'."
                )
            optimize_result.id = new_id
            self.list.append(optimize_result)
```

pypesto/result/optimize.py
```diff
@@ -164,18 +166,46 @@ def __len__(self):
...
 
         Parameters
         ----------
-        optimizer_result:
-            The result of one (local) optimizer run.
+        optimize_result:
+            The result of one or more (local) optimizer run.
+        sort:
+            Boolean used so we only sort once when appending an
+            optimize_result.
```
dilpath, line 182:
```suggestion
            optimize_result.
        prefix:
            The IDs for all appended results will be prefixed with this.
```

dilpath (approved):
:+1:

pypesto/result/optimize.py
PaulJonasJost: if I would do it like this what could happen is that some runs get appended cause they do have a unique id, but then one duplicate gets appended and throws an error... wouldn't it be faster/more convenient to check before?

dilpath: Ah right, since it's in-place. Perhaps appending to the same "run" should be in-place, and merging two different runs should return a new `OptimizeResult`, so that the constituent `OptimizeResult`s are left as-is? e.g. if this is used in pyPESTO behind the scenes, a user's objects should not be mutated I guess. Could still e.g. include an `inplace` argument for users that want the operation to be in-place (and agreed, if this case then a check can make sense)?

jvanhoefer:
Could be related to #710

pypesto/optimize/util.py
```diff
@@ -138,3 +139,38 @@ def bound_n_starts_from_env(n_starts: int):
...
+    ids:
+        Ids assigned to the startpoints.
+    result:
+        A result object to append the optimization results to. For example,
+        one might append more runs to a previous optimization. Assign only
+        unique ids.
+    n_starts:
+        Number of starts of the optimizer.
+    """
+    used_ids = set({})
```
dilpath, line 161:
```suggestion
    used_ids = set()
```

pypesto/optimize/util.py
```diff
@@ -138,3 +139,38 @@ def bound_n_starts_from_env(n_starts: int):
...
+    if result is not None:
+        used_ids = result.optimize_result.id
```
dilpath, line 163: `used_ids` changes to type `list` here (maybe causes the `isdisjoint` call to fail?)
```suggestion
        used_ids = set(result.optimize_result.id)
```
Or change `used_ids.isdisjoint(ids)` to `set(used_ids).isdisjoint(ids)`

pypesto/optimize/util.py
```diff
@@ -138,3 +139,38 @@ def bound_n_starts_from_env(n_starts: int):
...
+        i = 0
+        ids = [str(j) for j in range(i * n_starts, (i + 1) * n_starts)]
+        while not used_ids.isdisjoint(ids):
+            i += 1
+            ids = [str(j) for j in range(i * n_starts, (i + 1) * n_starts)]
+    if len(ids) != n_starts:
+        raise AssertionError("Number of starts and ids must coincide.")
+    if not used_ids.isdisjoint(ids):
+        raise AssertionError(
+            "Manually assigned ids must differ from " "existing ones."
```
dilpath, line 174:
```suggestion
            "Manually assigned ids must differ from existing ones."
```

pypesto/optimize/util.py
```diff
@@ -138,3 +139,38 @@ def bound_n_starts_from_env(n_starts: int):
...
+    """
+    used_ids = set({})
+    if result is not None:
+        used_ids = result.optimize_result.id
+    if ids is None:
+        i = 0
+        ids = [str(j) for j in range(i * n_starts, (i + 1) * n_starts)]
+        while not used_ids.isdisjoint(ids):
+            i += 1
+            ids = [str(j) for j in range(i * n_starts, (i + 1) * n_starts)]
```
dilpath, line 169: If `N_starts = len(used_ids)` is the number of previous starts, and `N_starts >> n_starts`, then this might loop many times. Perhaps can use multiples of `N_starts` as the starting point? e.g. `range(i * N_starts, i * N_starts + n_starts)`

PaulJonasJost: might be a good idea 👍

PR #818, opened 2022-02-05
Added Pymc v4 Sampler

PaulJonasJost:
This is to be able to use Apsara in combination with pymc. For now only a draft.

TODO:
- [x] Import for pymc via GitHub probably until official version is released
- [x] make pymc3 deprecated as it requires old versions of e.g. arviz which would need to be updated for pymc
- [x] make DensityDist work with tensor vector
- [x] rework notebooks when pymc sampler is working to fix bumpy problem
- [x] PyMC3 tests and examples have been temporarily disabled in https://github.com/ICB-DCM/pyPESTO/pull/831. They should be re-enabled here with their PyMC4 counterparts


dweindl:
- [ ] PyMC3 tests and examples have been temporarily disabled in https://github.com/ICB-DCM/pyPESTO/pull/831.They should be re-enabled here with their PyMC4 counterparts

dweindl (approved):
I don't have any experience with pymc/aesara, but I don't see any obvious issues.
Somebody more qualified should have another look.

pypesto/objective/aesara.py
```diff
@@ -152,6 +158,69 @@ def __deepcopy__(self, memodict=None):
...
+            ndim_supp=0,
+            ndims_params=[x_size],
+            dtype='floatX',
+        )
+
+
+@_logprob.register(AesaraObjectiveRV)
+def aesara_logprob(op, values, *inputs, **kwargs):
+    """
+    Aeppl registration of the logprob function of AesaraObjectiveRV class.
```
yannikschaelte, line 216:
```suggestion
    Apply/apple/aesarapla? registration of the logprob function of AesaraObjectiveRV class.
```

pypesto/sample/pymc.py
```diff
@@ -123,37 +140,30 @@ def sample(self, n_samples: int, beta: float = 1.0):
             ]
 
             # convert to tensor vector
-            theta = tt.as_tensor_variable(k)
+            theta = at.as_tensor_variable(k)
 
             # use a DensityDist for the log-posterior
```
yannikschaelte, line 145: update docstring

yannikschaelte (approved):
looks good to me, great!

pypesto/objective/aesara.py
FFroehlich: It's really called `aeppl` :D https://github.com/aesara-devs/aeppl

yannikschaelte: oh whoopsie :rofl:

PR #822, opened 2022-02-24
Autosave filename method

dilpath:
Autosaving now supports specifying the filename with a user-supplied method. Useful to organize autosave files when many such files are created. The previous `"Auto"` option is now implemented as such a method (should be backwards compatible).
e.g. following organizes autosave files neatly on a HPC with Slurm job arrays
```python
import os
from pathlib import Path


job_array_id = os.environ.get('SLURM_ARRAY_TASK_ID', 0)
autosave_path = Path('output') / 'autosave'


def filename(iteration=[0], **kwargs):
    iteration[0] += 1
    # `5` set arbitrarily here to provide sufficient padding.
    # Could use something like the following instead (untested).
    # padding = math.ceil(math.log10(os.environ.get['SLURM_ARRAY_TASK_MAX']))
    directory = autosave_path / kwargs['store_type'] / f'{job_array_id:0>5}'
    directory.mkdir(parents=True, exist_ok=True)
    name = f'{iteration[0]}.h5'
    return str(directory / name)


# Use, e.g. in optimization
pypesto.optimize.minimize(..., filename=filename)
```

The following should simply move the default autosave files to their own directory (untested)
```python
from pathlib import Path

from pypesto.store.auto import default_filename


def filename(**kwargs):
    directory = Path('autosave')
    name = default_filename(**kwargs)
    return str(directory / name)


# Use, e.g. in optimization
pypesto.optimize.minimize(..., filename=filename)
```

### DIFF
--- a/pypesto/store/auto.py
+++ b/pypesto/store/auto.py
@@ -3,13 +3,14 @@
 import binascii
 import datetime
 import os
+from typing import Callable, Union
 
 from ..result import Result
 from .save_to_hdf5 import write_result
 
 
 def autosave(
-    filename: str,
+    filename: Union[str, Callable, None],
     result: Result,
     store_type: str,
     overwrite: bool = False,
@@ -23,6 +24,9 @@ def autosave(
         Either the filename to save to or "Auto", in which case it will
         automatically generate a file named
         `year_month_day_{type}_result.hdf5`.
+        A method can also be provided. All input to the autosave method will
+        be passed to the filename method. The output should be the filename
+        (`str`).
     result:
         The result to be saved.
     store_type:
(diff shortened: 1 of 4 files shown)


pypesto/store/auto.py
```diff
@@ -23,6 +24,9 @@ def autosave(
         Either the filename to save to or "Auto", in which case it will
         automatically generate a file named
         `year_month_day_{type}_result.hdf5`.
+        A method can also be provided. All input to the autosave method will
+        be passed to the filename method. The output should be the filename
+        (`str`).
     result:
```
PaulJonasJost, line 30: This would then only allow for result, store_type and overwrite to be passed to the callable? perhaps also add *kwargs there?

PaulJonasJost (approved):

pypesto/store/auto.py
dilpath: Could you describe the use case? i.e. what else can be passed from `autosave` to `default_filename`?

PR #825, opened 2022-03-18
Fix #824

### LINKED ISSUE #824: Optimization with more than 100 starts defaults to 100
On the `develop` branch,
when using `optimize.minimize` with more than `n_starts=100` it defaults to 100 starts.

The reason is that `optimize.minimize` generates `ids` using the `utils.assign_ids` function without giving over `n_starts`
https://github.com/ICB-DCM/pyPESTO/blob/3fc99ad6dceafc5129cc2b1259d1b01daa1c57b5/pypesto/optimize/optimize.py#L106-L109

resulting in the use of `n_starts=100` by default.
https://github.com/ICB-DCM/pyPESTO/blob/3fc99ad6dceafc5129cc2b1259d1b01daa1c57b5/pypesto/optimize/util.py#L144-L148

The optimization `tasks` are defined by jointly iterating over `startpoints` and `ids` 
https://github.com/ICB-DCM/pyPESTO/blob/3fc99ad6dceafc5129cc2b1259d1b01daa1c57b5/pypesto/optimize/optimize.py#L127-L128
until either the `startpoints` run out (`n_starts` < 100) or `ids` run out (`n_starts` > 100)

m-philipps:
Specify the `n_starts` argument in `optimize.minimize` call to `optimize.util.assign_ids`

### DIFF
--- a/pypesto/optimize/optimize.py
+++ b/pypesto/optimize/optimize.py
@@ -106,7 +106,7 @@ def minimize(
     ids = assign_ids(
         n_starts=n_starts,
         ids=ids,
-        result=result
+        result=result,
     )
 
     # prepare result
--- a/pypesto/optimize/optimize.py
+++ b/pypesto/optimize/optimize.py
@@ -106,6 +106,7 @@ def minimize(
     ids = assign_ids(
         ids=ids,
         result=result,
+        n_starts=n_starts
     )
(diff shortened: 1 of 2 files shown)


PaulJonasJost (approved):
👍

PR #826, opened 2022-03-28
Temporarily fix jinja2 version

dilpath:
As similarly done for libpetab-python https://github.com/PEtab-dev/libpetab-python/pull/115


yannikschaelte (approved):
as similarly done for _each_ of our python packages ;)

PR #834, opened 2022-03-30
fix optimizer history fval offset

plakrisenko:
fix `fval` offset that is needed for log scale


dweindl:
Can't see where this fixes an offset :thinking:

plakrisenko:
> Can't see where this fixes an offset thinking

There is an if statement `if y_label == 'fval':`, but `y_label` is actually `objective value` instead of `fval`, so the offset is never applied.

plakrisenko:
Probably even better to introduce constants to the visualization module

dweindl:
> > Can't see where this fixes an offset thinking
> 
> There is an if statement `if y_label == 'fval':`, but `y_label` is actually `objective value` instead of `fval`, so the offset is never applied.

Ohh. This looks extremely fragile. Can't that be handled in a way other than comparing axis labels? At least, define some constant that is used in both places, to make that somewhat more robust?

PaulJonasJost:
> > > Can't see where this fixes an offset thinking
> > 
> > 
> > There is an if statement `if y_label == 'fval':`, but `y_label` is actually `objective value` instead of `fval`, so the offset is never applied.
> 
> Ohh. This looks extremely fragile. Can't that be handled in a way other than comparing axis labels? At least, define some constant that is used in both places, to make that somewhat more robust?

Would agree wich Daniel and your opinion Polina, some constants would be useful.

plakrisenko:
In `optimizer_history` trace_y can be equal to one of the tree values
https://github.com/ICB-DCM/pyPESTO/blob/866f2ab13a651da38f33010fce3271f77c4249e2/pypesto/visualize/optimizer_history.py#L51

In `get_trace` one can see that only 'fval' and 'gradnorm' are actually implemented
https://github.com/ICB-DCM/pyPESTO/blob/866f2ab13a651da38f33010fce3271f77c4249e2/pypesto/visualize/optimizer_history.py#L252-L269

But in `handle_options` 'fval' references are plotted 
https://github.com/ICB-DCM/pyPESTO/blob/866f2ab13a651da38f33010fce3271f77c4249e2/pypesto/visualize/optimizer_history.py#L413

Do references make sense in case trace_y == 'gradnorm'?

pypesto/visualize/optimizer_history.py
```diff
@@ -33,44 +45,44 @@ def optimizer_history(
...
-        Default: 'steps'
-    trace_y: str, optional
+        Possibilities: TRACE_X
+        Default: TRACE_X_STEPS
+    trace_y:
         What should be plotted on the y-axis?
-        Possibilities: 'fval', 'gradnorm', 'stepsize'
-        Default: 'fval'
-    scale_y: str, optional
+        Possibilities: TRACE_Y
```
PaulJonasJost, line 61: not sure whether this will be displayed correctly as the tuple. If so everything fine, otherwise would change to
`TRACE_Y_FVAL, TRACE_Y_GRADNORM`

PaulJonasJost (approved):

PaulJonasJost:
> In `optimizer_history` trace_y can be equal to one of the tree values
> 
> https://github.com/ICB-DCM/pyPESTO/blob/866f2ab13a651da38f33010fce3271f77c4249e2/pypesto/visualize/optimizer_history.py#L51
> 
> In `get_trace` one can see that only 'fval' and 'gradnorm' are actually implemented
> 
> https://github.com/ICB-DCM/pyPESTO/blob/866f2ab13a651da38f33010fce3271f77c4249e2/pypesto/visualize/optimizer_history.py#L252-L269
> 
> But in `handle_options` 'fval' references are plotted
> 
> https://github.com/ICB-DCM/pyPESTO/blob/866f2ab13a651da38f33010fce3271f77c4249e2/pypesto/visualize/optimizer_history.py#L413
> 
> Do references make sense in case trace_y == 'gradnorm'?
> 
> Should I keep 'stepsize' value for trace_y as todo or simply remove it?

Would just remove stepsize imo

PR #835, opened 2022-03-30
AmiciObjective: Set amici.Solver.setReturnDataReportingMode

dweindl:
... to only compute those quantities relevant for parameter estimation

### DIFF
--- a/pypesto/objective/amici.py
+++ b/pypesto/objective/amici.py
@@ -65,6 +65,7 @@ class AmiciObjective(ObjectiveBase):
         fim_for_hess: Optional[bool] = True,
         amici_object_builder: Optional[AmiciObjectBuilder] = None,
         calculator: Optional[AmiciCalculator] = None,
+        amici_reporting: Optional['amici.RDataReporting'] = None,
     ):
         """
         Initialize objective.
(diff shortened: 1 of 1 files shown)


pypesto/objective/amici.py
```diff
@@ -372,6 +379,20 @@ def call_unprocessed(
         """
         x_dct = self.par_arr_to_dct(x)
 
+        # only ask amici to compute required quantities
```
FFroehlich, line 382: Go to full reporting if return of rdatas is requested? Would also work for time trajectories without the changes below.

FFroehlich (approved):
Makes sense!

pypesto/objective/amici.py
dweindl: Oh, good point.

dweindl: Wait. You mean `return_dict==True`? That is only handled in `ObjectBase`. `AmiciObjective` doesn't ever get so see that. Would have to pass that into all descendants of `ObjectBase` then. Do we want that?

dweindl: Okay, handled it via __call__

FFroehlich: neat!

PR #845, opened 2022-04-05
Optimize pypesto/objective/amici_util.py::par_index_slices

dweindl:
Improve parameter mapping performance in `pypesto/objective/amici_util.py::par_index_slices`.

For large models, this used to perform extremely bad. When optimizing https://github.com/ICB-DCM/CS_Signalling_ERBB_RAS_AKT/tree/master/FroehlichKes2018/PEtab, almost half of the time is spent on this parameter mapping, as profiled by @plakrisenko.

These changes here speed things up by a factor of about 250x.

### DIFF
--- a/pypesto/objective/amici_util.py
+++ b/pypesto/objective/amici_util.py
@@ -172,7 +172,7 @@ def par_index_slices(
     zip_iterator = zip(
         *(
             (
-                # simulation parameter index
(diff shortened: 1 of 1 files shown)


FFroehlich (commented):
I think I need some inline comments to follow what is happening here

FFroehlich (approved):

PR #848, opened 2022-04-07
PetabImporter: Auto-regenerate AMICI models in case of version mismatch

dweindl:
For models that were generated with amici>=0.11.27, this will automatically re-import AMICI models in case of AMICI version mismatch


FFroehlich (approved):
👍

dilpath (approved):
Could print a message since users may be surprised to see the regeneration.

PR #855, opened 2022-04-14
Ensemble calc cutoff

PaulJonasJost:


pypesto/ensemble/ensemble.py
```diff
@@ -603,8 +604,9 @@ def from_sample(
     @staticmethod
     def from_optimization_endpoints(
         result: Result,
-        cutoff: float = np.inf,
+        cutoff: float = None,
         max_size: int = np.inf,
+        percentile: float = 0.95,
```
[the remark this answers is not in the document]
dilpath: This is how percentiles are handled in ensemble visualization currently.

https://github.com/ICB-DCM/pyPESTO/blob/4c30abfca56ba714c302141cd44a9dd366bff4bb/pypesto/visualize/sampling.py#L132-L133

pypesto/ensemble/ensemble.py
```diff
@@ -616,14 +618,26 @@ def from_optimization_endpoints(
...
         The ensemble.
         """
+        if cutoff is None:
+            if percentile is None:
+                raise ValueError(
+                    'Neither cutoff nor percentile were set. '
+                    'Determining a cutoff value is not '
+                    'possible. Please set one of the two.'
+                )
+            cutoff = calculate_cutoff(result=result, percentile=percentile)
```
dilpath, line 640: Warn if percentile is not default, and both percentile and cutoff are specified? Another reason why defaulting both to `None` can be better

pypesto/ensemble/ensemble.py
```diff
@@ -616,14 +618,26 @@ def from_optimization_endpoints(
...
         Returns
         -------
         The ensemble.
         """
+        if cutoff is None:
+            if percentile is None:
+                raise ValueError(
+                    'Neither cutoff nor percentile were set. '
+                    'Determining a cutoff value is not '
+                    'possible. Please set one of the two.'
```
dilpath, line 638:
```suggestion
                    'Please specify either a cutoff or percentile. '
                    'To use all vectors, specify `percentile=1.0`.'
```

pypesto/ensemble/ensemble.py
```diff
@@ -671,10 +685,11 @@ def from_optimization_endpoints(
     @staticmethod
     def from_optimization_history(
```
dilpath, line 686: Some comments in `from_optimization_endpoints` apply here too.

pypesto/ensemble/ensemble.py
```diff
@@ -1157,3 +1184,33 @@ def get_percentile_label(percentile: Union[float, int, str]) -> str:
...
     return f'{PERCENTILE} {percentile}'
+
+
+def calculate_cutoff(result: Result, percentile: float = 0.95):
+    """
+    Calculate the cutoff of the ensemble.
+
+    Based on the number of parameters of the problem. Based on the
+    assumpiton that the difference of the NLLs of the true and optimal
+    parameter is chi^2 distributed with n_theta degress of freedom.
```
dilpath, line 1195: - `nllh` elsewhere in this PR, `NLL` here
- `n_theta` could be changed e.g. "chi^2 distribution with degrees of freedom equal to the number of estimated parameters"
- typo "assumption"

pypesto/ensemble/ensemble.py
```diff
@@ -1157,3 +1184,33 @@ def get_percentile_label(percentile: Union[float, int, str]) -> str:
...
+
+    Parameters
+    ----------
+    result: The optimization result from which to create the ensemble.
+    percentile: The percentile of the chi^2 distribution. Between 0 and 1.
+    Higher values will result in a more lax cutoff. If the value is greater
+    than 1, the cutoff will be returned as np.inf.
```
dilpath, line 1202: Different docstring format?

```suggestion
    result:
        The optimization result from which to create the ensemble.
    percentile:
        The percentile of the chi^2 distribution. Between 0 and 1.
        Higher values will result in a more lax cutoff. If the value is greater
        than 1, the cutoff will be returned as np.inf.
```

pypesto/ensemble/ensemble.py
```diff
@@ -616,14 +618,26 @@ def from_optimization_endpoints(
             A pyPESTO result that contains an optimization result.
         cutoff:
             Exclude parameters from the optimization if the
-            nllh is higher than the `cutoff`.
+            nllh is higher than the `cutoff`. If cutoff is None, a cutoff
+            will be calculated based on the percentile passed.
```
dilpath, line 622: Describe that `cutoff` is actually an absolute threshold on the likelihood. e.g. to get all vectors s.t. `fval(vector) <= fval(optimal_vector) + 5`, specify `cutoff=fval(optimal_vector) + 5`.

dilpath: Would prefer if `cutoff` could be a relative difference in `fval`. e.g.
`cutoff=5`
instead of
`cutoff=fval(optimal_vector) + 5`

This perhaps makes more sense since a user may then calculate the cutoff independently and in advance of the optimization result, e.g. the following is independent of the optimization result.
```python
    df = result.problem.dim
    range = chi2.ppf(q=percentile, df=df)
    cutoff = range
```

dilpath (changes requested):

pypesto/ensemble/ensemble.py
```diff
@@ -1191,26 +1193,28 @@ def calculate_cutoff(result: Result, percentile: float = 0.95):
     Calculate the cutoff of the ensemble.
 
     Based on the number of parameters of the problem. Based on the
-    assumpiton that the difference of the NLLs of the true and optimal
+    assumpiton that the difference of the nllh's of the true and optimal
```
dilpath, line 1196:
```suggestion
    assumption that the difference of the nllh's of the true and optimal
```

pypesto/ensemble/ensemble.py
```diff
@@ -1191,26 +1193,28 @@ def calculate_cutoff(result: Result, percentile: float = 0.95):
...
-    result: The optimization result from which to create the ensemble.
-    percentile: The percentile of the chi^2 distribution. Between 0 and 1.
-    Higher values will result in a more lax cutoff. If the value is greater
-    than 1, the cutoff will be returned as np.inf.
+    result:
+        The optimization result from which to create the ensemble.
+    percentile:
+        The percentile of the chi^2 distribution. Between 0 and 100.
+        Higher values will result in a more lax cutoff. If the value is greater
+        than 100, the cutoff will be returned as np.inf.
```
dilpath, line 1206: In general I am in favor of failing when given unexpected input, but fine as is

pypesto/ensemble/ensemble.py
```diff
@@ -697,19 +713,31 @@ def from_optimization_history(
...
         Returns
         -------
         The ensemble.
         """
+        if rel_cutoff is None and percentile is None:
+            abs_cutoff = np.inf
+        elif rel_cutoff is not None:
+            abs_cutoff = result.optimize_result[0].fval + rel_cutoff
+        else:
+            abs_cutoff = calculate_cutoff(result=result, percentile=percentile)
```
dilpath, line 729: Repeated code, but fine as is

pypesto/ensemble/ensemble.py
```diff
@@ -1210,7 +1215,10 @@ def calculate_cutoff(result: Result, percentile: float = 0.95):
     The calculated cutoff value.
     """
     if percentile >= 100:
```
dilpath, line 1217:
```suggestion
    if percentile > 100:
```

PR #858, opened 2022-04-21
trace equilibration times and total simulation time

plakrisenko:


PaulJonasJost:
@plakrisenko what is the status on this? :)

PR #859, opened 2022-04-27
Add inf value to optimize summary

PaulJonasJost:


pypesto/result/optimize.py
```diff
@@ -217,13 +217,14 @@ def summary(self):
         summary = (
             "## Optimization Result \n\n"
             f"* number of starts: {len(self)} \n"
+            f"* best value: {self[0]['fval']}, "
+            f"worst value: {self[-1]['fval']} \n"
+            f"* number of infinite values: {np.isinf(self.fval).sum()}\n\n"
```
dweindl, line 222: Let's include NaNs?
```suggestion
            f"* number of non-finite values: {np.logical_not(np.isfinite(self.fval)).sum()}\n\n"
```

dilpath (approved):

Issue #828, opened 2022-03-29
Using Hdf5History with an existing file produces funny results


dweindl:
**Bug description**
When saving optimization history using `Hdf5History` to an existing files, values for individual iterations are overwritten as they occur. Other old entries remain, making it very unpleasant to deal with such histories.

**Expected behavior**
Either refuse to operate on files that already have a history for the respective id, or delete the respective group upon first write.

**To reproduce**
run `doc/example/store.ipynb` twice and check output of 
```
for r in result.optimize_result.list:
    print(np.diff(r.history.get_time_trace()))
```
this will occasionally show negative values, whereas values of `get_time_trace` should be strictly increasing.

PR #861, opened 2022-05-03
Fix hdf5 history overwrite

PaulJonasJost:
For now, not allowing to write into the history file on already existing id's.

### DIFF
--- a/pypesto/objective/history.py
+++ b/pypesto/objective/history.py
@@ -1229,10 +1229,13 @@ class Hdf5History(History):
             false if the history is a loaded one to prevent overwriting.
 
         """
-        with h5py.File(file, 'r') as f:
-            return file, (
-                'history' not in f.keys() or self.id not in f['history']
-            )
+        try:
+            with h5py.File(file, 'r') as f:
+                return file, (
+                    'history' not in f.keys() or self.id not in f['history']
+                )
+        except OSError:  # if the file is non-existent, return editable = True
+            return file, True
 
 
 class OptimizerHistory:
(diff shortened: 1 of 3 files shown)


pypesto/objective/history.py
```diff
@@ -1205,6 +1207,23 @@ def get_time_trace(
...
 
+    def _check_file_id(self, file: str):
+        """
+        Check, whether the id is already existent in the file.
+
+        Parameters
+        ----------
+        file:
+            HDF5 file name.
+        """
```
dweindl, line 1218: Please document return values. It's not obvious for me what "editable" means here.

pypesto/objective/history.py
```diff
@@ -1205,6 +1207,23 @@ def get_time_trace(
...
+        with h5py.File(file, 'a') as f:
```
dweindl, line 1219: Why `'a'`? To check that it's actually writable? If so, potential exceptions should be handled.

pypesto/objective/history.py
```diff
@@ -1205,6 +1207,23 @@ def get_time_trace(
...
+            editable = True
+            if 'history' not in f.keys():
+                return file, editable
+            if self.id in f['history']:
+                editable = False
+            return file, editable
```
dweindl, line 1225:
```suggestion
            return file, ('history' not in f.keys() or self.id not in f['history'])
```

pypesto/objective/history.py
```diff
@@ -936,6 +936,8 @@ def update(
         result: ResultDict,
     ) -> None:
         """See `History` docstring."""
+        if not self.editable:
+            raise ValueError('This id is already used in the history file.')
```
dweindl, line 940:
```suggestion
            raise ValueError(f'ID "{self.id}" is already used in history file "{self.file}".')
```

jvanhoefer (approved):

PaulJonasJost:
closes #828

PR #862, opened 2022-05-03
Read problem with results from file

plakrisenko:
Of course, pypesto problem can be added to the result afterwards. If you think it's unnecessary, pull-request can be deleted :)


PaulJonasJost (approved):
I do think it makes a lot of sense 👍

PR #864, opened 2022-05-11
Updated the profile, minimize, sample and added overwrite as argument.

PaulJonasJost:
I added `overwrite` to the three main functions as argument. Currently `autosave` has the potential argument overwrite but it is never used. This is so that people can overwrite their results without having to save it manually. Is set to false by default, so should be fine. 

If it is not wanted to have many arguments in minimise, could potentially add **kwargs to it?

### DIFF
--- a/pypesto/optimize/optimize.py
+++ b/pypesto/optimize/optimize.py
@@ -32,6 +32,7 @@ def minimize(
     options: OptimizeOptions = None,
     history_options: HistoryOptions = None,
     filename: Union[str, Callable, None] = "Auto",
+    overwrite: bool = False,
 ) -> Result:
     """
     Do multistart optimization.
@@ -68,6 +69,8 @@ def minimize(
         `year_month_day_optimization_result.hdf5`. Deactivate saving by
         setting filename to `None`.
         Optionally a method, see docs for `pypesto.store.auto.autosave`.
+    overwrite:
+        Boolean, whether to potentially overwrite the existing file `filename`.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


pypesto/optimize/optimize.py
```diff
@@ -68,6 +69,8 @@ def minimize(
         `year_month_day_optimization_result.hdf5`. Deactivate saving by
         setting filename to `None`.
         Optionally a method, see docs for `pypesto.store.auto.autosave`.
+    overwrite:
+        Boolean, whether to potentially overwrite the existing file `filename`.
```
dilpath, line 73:
```suggestion
        Whether to overwrite the autosave file if it already exists.
```

pypesto/profile/profile.py
```diff
@@ -66,6 +67,8 @@ def parameter_profile(
         `year_month_day_profiling_result.hdf5`. Deactivate saving by
         setting filename to `None`.
         Optionally a method, see docs for `pypesto.store.auto.autosave`.
+    overwrite:
+        Boolean, whether to potentially overwrite the existing file `filename`.
```
dilpath, line 71:
```suggestion
        Whether to overwrite the autosave file if it already exists.
```

pypesto/sample/sample.py
```diff
@@ -48,6 +49,8 @@ def sample(
         `year_month_day_sampling_result.hdf5`. Deactivate saving by
         setting filename to `None`.
         Optionally a method, see docs for `pypesto.store.auto.autosave`.
+    overwrite:
+        Boolean, whether to potentially overwrite the existing file `filename`.
```
dilpath, line 53:
```suggestion
        Whether to overwrite the autosave file if it already exists.
```

dilpath (approved):
Maybe a little confusing to users, since this overwrites only some of the pre-existing HDF5 file right? Whereas normal usage of "overwrite" would mean "delete the old file first, then start writing current results".

PR #873, opened 2022-05-23
Enabled saving messages and exit flags in hdf5 history in case of finished run

PaulJonasJost:


pypesto/optimize/load.py
```diff
@@ -82,6 +82,12 @@ def fill_result_from_history(
     if not optimize_options.history_beats_optimizer:
         return result
 
+    if isinstance(optimizer_history.history, Hdf5History):
+        if optimizer_history.history.message is not None:
+            result.message = optimizer_history.history.message
+        if optimizer_history.history.exitflag is not None:
+            result.exitflag = optimizer_history.history.exitflag
```
dweindl, line 89:
```suggestion
        if (message := optimizer_history.history.message) is not None:
            result.message = message
        if (exitflag := optimizer_history.history.exitflag) is not None:
            result.exitflag = exitflag
```
Avoids opening the file a second time.

PaulJonasJost: uh nice, didn't know that one :)

dweindl: More info: https://realpython.com/python-walrus-operator/ (python >=3.8)

plakrisenko (approved):

pypesto/objective/history.py
```diff
@@ -1315,9 +1350,20 @@ def update(
...
+        Parameters
+        ----------
+        message:
+            Optimizer message to be saved.
+        exitflag:
+            Optimizer exitflag to be saved.
+        """
         self.history.finalize()
+        if isinstance(self.history, Hdf5History):
+            self.history.finalize(message=message, exitflag=exitflag)
```
yannikschaelte, line 1366: this looks not clean -- the arguments should probably be passed to the history in any case without type checking.

yannikschaelte (approved):

pypesto/objective/history.py
PaulJonasJost: I was thinking about this as well, but that would need me to include those in the finalize of e.g. `MemoryHistory` which definitely does not need to save this in any way. Should I do it anyway? then probably in the `HistoryBase.finalize()`?

PR #874, opened 2022-05-26
Pass problem kwargs via petab importer

dilpath:


PaulJonasJost (approved):
What are possible kwargs?

dilpath:
> What are possible kwargs?

Just needed `copy_objective=False` for a custom PEtab-based objective.

PR #878, opened 2022-06-02
fides.Optimizer includes message according to exitflag

PaulJonasJost:


FFroehlich (approved):

pypesto/optimize/optimizer.py
```diff
@@ -1298,3 +1297,41 @@ def minimize(
...
+        opt:
+            The fides.Optimizer that has finished minimizing storing the
+            exitflag.
+
+        Returns
+        -------
+            An informative message on the cause of termination. Based on
+            fides documentation.
+        """
+        if opt.exitflag == 0:
```
FFroehlich, line 1316: Would recommend using values of the `fides.constants.ExitFlag` enum here. Might also be better to use a dict instead of `elifs`,=,

PaulJonasJost: Ok, thanks for the suggestions. :)

PR #880, opened 2022-06-03
Fix hdf5 result mismatch

### LINKED ISSUE #743: Flaky test_result_from_hdf5_history?

PaulJonasJost:
should fix #743. Ran the `test_result_from_hdf5_history` test 1000 times and it was successful every time.

Potentially we could remove the `_update` function of `OptimizerHistory`?

### DIFF
--- a/pypesto/objective/history.py
+++ b/pypesto/objective/history.py
@@ -1367,6 +1367,7 @@ class OptimizerHistory:
             Optimizer exitflag to be saved.
         """
         self.history.finalize(message=message, exitflag=exitflag)
+        self._compute_vals_from_trace()
 
     def _update_vals(self, x: np.ndarray, result: ResultDict):
         """Update initial and best function values."""
--- a/pypesto/objective/history.py
+++ b/pypesto/objective/history.py
@@ -1367,7 +1367,8 @@ class OptimizerHistory:
             Optimizer exitflag to be saved.
         """
         self.history.finalize(message=message, exitflag=exitflag)
-        self._compute_vals_from_trace()
+        if isinstance(self.history, (Hdf5History, MemoryHistory, CsvHistory)):
+            self._compute_vals_from_trace()
 
     def _update_vals(self, x: np.ndarray, result: ResultDict):
         """Update initial and best function values."""


yannikschaelte:
Sure this would remove any ambiguity by always taking results from the trace. But is it clear why this happens in the first place?

PaulJonasJost:
> Sure this would remove any ambiguity by always taking results from the trace. But is it clear why this happens in the first place?

Do not quote me on it, but I think it has to do with the fact that in line 1459, the algorithm goes in reversed order while in the update function he goes from the other side.

dweindl:
> > Sure this would remove any ambiguity by always taking results from the trace. But is it clear why this happens in the first place?
> 
> Do not quote me on it, but I think it has to do with the fact that in line 1459, the algorithm goes in reversed order while in the update function he goes from the other side.

Not sure I understand the original problem. Could you elaborate?

Also, the default tolerances implicitly used in `allclose` (`rtol=1e-05, atol=1e-08`) in the referenced line seem very loose to me.

dweindl:
Sorry for the delay, and thanks for the explanation.

> So in the [_update_vals()](https://github.com/ICB-DCM/pyPESTO/blob/46eb703c8fc21d1fcebc9dd1b2076bf72880355d/pypesto/objective/history.py#L1402) function of the Optimizer_History, we assign the grad_min, hess_min, res_min, sres_min in a way, that they get assigned if they are not yet assigned and the current updated point is close to the minimal point.

I don't quite understand why they are only assigned if they are `None` (@yannikschaelte ?). Could that be the underlying problem? 
https://github.com/ICB-DCM/pyPESTO/blob/46eb703c8fc21d1fcebc9dd1b2076bf72880355d/pypesto/objective/history.py#L1403-L1409

yannikschaelte:
closed in favor of #902

PR #883, opened 2022-06-24
Fix set of keys to aggregate results

vwiela:
Calling an `AmiciObjective` with `sensi-order=(1,)` returns a dictionary with keys `fval, radta, grad`, but a `pypesto.Objective` will just return a dictionary with keys `grad`. So an aggregated objective build by a combination of both will fail if called with `sensi-orders=(1,)`, if the `Amici.Objective` is the first partial objective. This also means the `.get_grad` function will fail and one cannot perfom optimization. 
This can be fixed by taking the intersection of the keys of all results instead of just taking the keys of the first partial objective.

### DIFF
--- a/pypesto/objective/aggregated.py
+++ b/pypesto/objective/aggregated.py
@@ -122,7 +122,7 @@ def aggregate_results(rvals: Sequence[ResultDict]) -> ResultDict:
     result = {
         key: sum(rval[key] for rval in rvals)
         for key in [FVAL, CHI2, SCHI2, GRAD, HESS, HESSP]
-        if all([key in rval for rval in rvals])
+        if all(key in rval for rval in rvals)
     }
 
     # extract rdatas and flatten
--- a/pypesto/objective/aggregated.py
+++ b/pypesto/objective/aggregated.py
@@ -118,9 +118,7 @@ def aggregate_results(rvals: Sequence[ResultDict]) -> ResultDict:
     rvals:
         results to aggregate
     """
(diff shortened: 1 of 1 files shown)


pypesto/objective/aggregated.py
```diff
@@ -118,15 +118,11 @@ def aggregate_results(rvals: Sequence[ResultDict]) -> ResultDict:
...
-    # whether each objective can be called with the respective
-    # sensi_orders/mode
-
-    # sum over fval/grad/hess
+    # sum over fval/grad/hess, if available in all rvals
     result = {
         key: sum(rval[key] for rval in rvals)
         for key in [FVAL, CHI2, SCHI2, GRAD, HESS, HESSP]
-        if rvals[0].get(key, None) is not None
+        if all([rval.get(key, None) is not None for rval in rvals])
```
FFroehlich, line 125:
```suggestion
        if all([key in rval for rval in rvals])
```

FFroehlich (approved):

pypesto/objective/aggregated.py
```diff
@@ -122,7 +122,7 @@ def aggregate_results(rvals: Sequence[ResultDict]) -> ResultDict:
     result = {
         key: sum(rval[key] for rval in rvals)
         for key in [FVAL, CHI2, SCHI2, GRAD, HESS, HESSP]
-        if all([rval.get(key, None) is not None for rval in rvals])
+        if all([key in rval for rval in rvals])
```
dweindl, line 125:
```suggestion
        if all(key in rval for rval in rvals)
```

PR #887, opened 2022-07-12
Fix CITATION.cff

dweindl:
* Remove `version` attribute which messes up Zenodo metadata if not updated on every release
* Add required `message`
* Auto-validate via GHA


FFroehlich (approved):
👍

PR #889, opened 2022-07-18
Parameter Indices Ordering

PaulJonasJost:
Changed visualize.profiles in a way that if you supply a different ordering in your parameter indices, this ordering is retained in the visualization.


pypesto/visualize/profiles.py
```diff
@@ -228,8 +242,12 @@ def profiles_lowlevel(
             tmp_legend = None
 
         # create or choose an axes object
+        if profile_order is None:
```
yannikschaelte, line 245: document what's happening here

yannikschaelte (commented):
What about just passing profile_indices into profile_lowlevel always? no need to have the check if any order is unexpected.

PaulJonasJost:
closing this for a lack of usage currently

PR #893, opened 2022-07-26
Fix doc build with latest sphinx etc

yannikschaelte:
* recommonmark -> myst as deprecated
* sphinx version can be unfixed again without error
* explicit language definition "en" required


dweindl (approved):
Looks good

Issue #708, opened 2021-07-21
Use Literal consistently to annotate type hint, where string arguments come from a predefined input range.


jvanhoefer:
**Feature description**
Since python 3.8 one can add a type hint of the form `argument: Literal['a', 'b', 'c']`, if a string argument comes from a predefined range of inputs. We could use this at many locations, where options are handed over as strings.

**To be discussed before**
* If we want to do this, we should do it consistently throughout pypesto,  and not only at some random spots...
* We would drop python 3.6/7 support by doing that (python would not be importable in that python versions any more). We have to discuss python version support first.

yannikschaelte:
not before #737

PR #899, opened 2022-07-26
Literal typehints for `mode`

dweindl:
See #708


dweindl:
> improves readability a bit I guess, 

Not sure about that :sweat_smile:, but it may help to spot typos before running anything

> but it appears a pity that Literal does not allow constant arguments, such that one needs to retype values.

Yes.

PR #902, opened 2022-07-30
Maybe fix history - result reconstruction mismatch

yannikschaelte:
Should fix #743.

* Essentially @PaulJonasJost was right in #880: At the end in finalize, one should consult the history, which may have more entries than saved in the ..._min attributes of OptimizerHistory.
* However, only overwrite if history has a better value (hardly possible) or has additional values, as ..._min may contain values not saved in history, e.g. when trace_record_grad is False.

Further
* Set some checks to np.array_equal instead of np.allclose, as in-memory changes to array entries should not be possible.
* Remaining np.allclose checks are I think only required because CsvHistory has an information loss, which MemoryHistory and Hdf5History don't have. If possible, one could further adjust the tolerances atol, rtol to the accuracy of the storage (e.g. 16 digits instead of numpy's default 1e-5, 1e-8?)
* And did some tidying up.

PR #903, opened 2022-07-31
Move history to own module

yannikschaelte:
Reason: One file with >>1000 loc is nasty to navigate.

No code changes, just refactoring. Waiting for #902.

If external code uses `pypesto.objective.history.{...}`, one can instead move the history folder to a subfolder of observable.


dweindl:
> Reason: One file with >>1000 loc is nasty to navigate.

Talking about nasty to navigate: +2,079 −1,754 :grimacing: 

But makes sense to move, yes.

yannikschaelte:
> > Reason: One file with >>1000 loc is nasty to navigate.
> 
> Talking about nasty to navigate: +2,079 −1,754 grimacing
> 
> But makes sense to move, yes.

now reduced to +1,916 −1,796 ^^

dweindl (approved):

PR #907, opened 2022-08-02
Fix read_result with history

PaulJonasJost:
Fixed a typo in a function when reading a result with history. Also added some clarification to doctoring of read_result.

### DIFF
--- a/pypesto/store/read_from_hdf5.py
+++ b/pypesto/store/read_from_hdf5.py
@@ -68,7 +68,7 @@ def read_hdf5_optimization(
         if optimization_key == 'history':
             if optimization_key in f:
                 result['history'] = Hdf5History(id=opt_id, file=file_name)
-                result['history']._recover_options(file_name)
+                result['history'].recover_options(file_name)
                 continue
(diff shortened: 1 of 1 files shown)


yannikschaelte (approved):

pypesto/store/read_from_hdf5.py
```diff
@@ -269,6 +269,9 @@ def read_result(
 ) -> Result:
     """Save the whole pypesto.Result object in an HDF5 file.
 
+    By default, loads everything. If any of `optimize, profile, sample` is
+    explicitly set to true, load <u>__only__</u> this one.
```
yannikschaelte, line 273: should work without html, e.g. *only* `*only*`

yannikschaelte:
> Fixed a typo in a function when reading a result with history. Also added some clarification to doctoring of read_result.

lol doctoring

pypesto/store/read_from_hdf5.py
yannikschaelte, line 273:
```suggestion
    explicitly set to true, load *only* this one.
```

PR #908, opened 2022-08-02
Clean up history hierarchy

yannikschaelte:
* Properly mark abstract base class methods as abstract instead of NotImplementedError, to not get surprises about missing methods at runtime.
* Refactor history hierarchy, as previously e.g. hdf5history was re-implementing and not using existing fields.


pypesto/history/base.py
```diff
@@ -246,51 +251,117 @@ def get_trimmed_indices(self) -> np.ndarray:
...
-    Parameters
-    ----------
-    options:
-        History options.
+    def update(  # noqa: D102
+        self,
+        x: np.ndarray,
+        sensi_orders: Tuple[int, ...],
+        mode: ModeType,
+        result: ResultDict,
```
dweindl, line 276:
```suggestion
        *args, 
        **kwargs,
```
Also for all other not-planned-to-ever-be-implemented methods?

pypesto/history/base.py
```diff
@@ -246,51 +251,117 @@ def get_trimmed_indices(self) -> np.ndarray:
...
+    Can be used whenever a history object is needed, but no history is desired.
+    Can be created, but not queried.
     """
-    Tracks number of function evaluations only, no trace.
 
-    Parameters
-    ----------
-    options:
-        History options.
+    def update(  # noqa: D102
```
dweindl, line 271: Those `# noqa: D102` are so ugly. Is there no way around that to not enforce docstrings in overriding methods?

dweindl (approved):

pypesto/history/base.py
yannikschaelte: I think it makes sense to define the function header in as much detail as possible, right?

pypesto/history/base.py
yannikschaelte: Yes, I agree. I played around with https://pypi.org/project/overrides/ and `flake8 --ignore-decorators=overrides` for a moment, but got recursion errors and decided it's not worth it ...

yannikschaelte: or now that most functions are documented, we could ignore D102 again, and just pay attention with new prs ... would have no problem with that per se ...

dweindl: Maybe safer with D102... Okay to leave.

pypesto/history/base.py
dweindl: Fine to keep, but doesn't feel helpful here. args/kwargs is less verbose and won't require changes if the base signature changes.

PR #909, opened 2022-08-02
Improve hdf5 history file lock

yannikschaelte:
* Avoid claustrophobia from "with" contexts
* Add field for start_time in hdf5 attributes
* Remove broad OSError catches from initial check functions, as these may hide issues.

### DIFF
--- a/pypesto/history/hdf5.py
+++ b/pypesto/history/hdf5.py
@@ -67,6 +67,20 @@ def with_h5_file(mode: str):
     return decorator
 
 
+def check_editable(fun):
+    """Check if the history is editable."""
+
+    def wrapper(self, *args, **kwargs):
+        if not self.editable:
+            raise ValueError(
+                f'ID "{self.id}" is already used in history file '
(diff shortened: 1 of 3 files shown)


pypesto/history/hdf5.py
```diff
@@ -362,7 +419,7 @@ def get_time_trace(
         """See `HistoryBase` docstring."""
         return self._get_hdf5_entries(TIME, ix)
 
-    def _is_editable(self, file: str) -> bool:
+    def _editable(self) -> bool:
```
dweindl, line 422: Why no `with_h5_file` here?

dweindl (approved):
:+1:

PaulJonasJost (approved):

pypesto/history/hdf5.py
yannikschaelte: because this function is evaluated at the beginning, and thus has an additional catch to just flag as not editable, but not raise an error.

PR #910, opened 2022-08-03
Added the offset value to waterfall plot for better intuitive understanding.

PaulJonasJost:


pypesto/visualize/waterfall.py
```diff
@@ -148,7 +148,10 @@ def waterfall(
     if offset_y == 0.0:
         ax.set_ylabel('Function value')
     else:
-        ax.set_ylabel('Offsetted function value (relative to best start)')
+        ax.set_ylabel(
```
yannikschaelte, line 151: Not sure whether this would be too much info in the plot. Alternatively one could just print the offset as info when plotting. Opinions / allow to pass ylabel as input and keep this here as default?

yannikschaelte (approved):

pypesto/visualize/waterfall.py
dweindl: I don't think it's too much info. It's rather a requirement for interpreting offset plots.

No strong opinion on whether it should go into the axis label or somewhere else.
For my purposes, I added it to the plotting region with
```python
        best_fval = min(x.fval for x in result.optimize_result.list)
        from matplotlib.offsetbox import AnchoredText

        text_box = AnchoredText(
            f"best fval: {best_fval}",
            frameon=True, loc='lower right', pad=0.5
        )
        plt.setp(text_box.patch,facecolor='white', alpha=0.5)
        ax.add_artist(text_box)
```
but that's even more prominent.

pypesto/visualize/waterfall.py
```diff
@@ -248,7 +251,10 @@ def waterfall_lowlevel(
...
+            'Offsetted function value (relative to best start, '
+            f'offset = {offset_y:0.3f})'
```
yannikschaelte, line 256:
```suggestion
            f'offset = {offset_y:.3e})'
```

pypesto/visualize/waterfall.py
```diff
@@ -148,7 +148,10 @@ def waterfall(
     if offset_y == 0.0:
         ax.set_ylabel('Function value')
     else:
-        ax.set_ylabel('Offsetted function value (relative to best start)')
+        ax.set_ylabel(
+            'Offsetted function value (relative to best start, '
+            f'offset = {offset_y:0.2f})'
```
yannikschaelte, line 153:
```suggestion
            f'offset = {offset_y:.3e})'
```

yannikschaelte (approved):

PR #913, opened 2022-08-03
[WIP] Try pytest-xdist

yannikschaelte:


yannikschaelte:
could reduce time; but too cumbersome to pull out the sequential setup code now.

PR #915, opened 2022-08-05
Use benchmark models petab

yannikschaelte:


setup.cfg
```diff
@@ -145,6 +145,7 @@ test =
     autograd >= 1.3
 test_petab =
     petabtests >= 0.0.0a6
+    benchmark_models_petab >= 0.0.0a1
```
PaulJonasJost, line 148: Shouldn't it be 
```suggestion
    Benchmark-Models-PEtab >= 0.0.0a1
```
according to the GitHub name?

PaulJonasJost (approved):

setup.cfg
yannikschaelte: that does not matter, as pip is case and _/- insensitive, according to https://peps.python.org/pep-0426/#name

PR #917, opened 2022-08-08
Fix flaky test CRFunModeHistoryTest.test_trace_all

dweindl:
The previous check did not handle large values too well. Failed here: https://github.com/ICB-DCM/pyPESTO/runs/7724332758?check_suite_focus=true


PaulJonasJost (approved):

yannikschaelte (commented):
looks good

PR #918, opened 2022-08-10
Select: cleanup use of `minimize_options` in model problem

dilpath:


pypesto/select/model_problem.py
```diff
@@ -122,12 +122,13 @@ def __init__(
                     )
                 # TODO rename `minimize_options` to `minimize_kwargs`.
                 # TODO or allow users to provide custom `minimize` methods?
-                elif minimize_options:
+                else:
```
PaulJonasJost, line 125: What happens now if minimise_options is None?

PaulJonasJost (approved):

pypesto/select/model_problem.py
dilpath: A default value of `{}` was already being set.

https://github.com/ICB-DCM/pyPESTO/blob/45e5ac09fe8c1f45cccfdd392c9133a58a9ef8c0/pypesto/select/model_problem.py#L86-L88

yannikschaelte (commented):
looks good

PR #922, opened 2022-08-12
Fix OptimizerResult.append

dweindl:
We want disjoint IDs, not overlapping IDs in the results to merge.

### DIFF
--- a/pypesto/result/optimize.py
+++ b/pypesto/result/optimize.py
@@ -269,12 +269,12 @@ class OptimizeResult:
         """
         current_ids = set(self.id)
         if isinstance(optimize_result, OptimizeResult):
-            new_ids = [
+            new_ids = {
                 prefix + identifier
                 for identifier in optimize_result.id
                 if identifier is not None
-            ]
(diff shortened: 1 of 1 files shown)


pypesto/result/optimize.py
```diff
@@ -269,12 +269,12 @@ def append(
...
         if isinstance(optimize_result, OptimizeResult):
-            new_ids = [
+            new_ids = {
                 prefix + identifier
                 for identifier in optimize_result.id
                 if identifier is not None
-            ]
-            if current_ids.isdisjoint(new_ids) and new_ids:
+            }
+            if not current_ids.isdisjoint(new_ids) and new_ids:
```
dilpath, line 277:
```suggestion
            if current_ids.intersection(new_ids):
```

dweindl: The former is faster. We don't need the full intersection, it's enough to know that the intersection is non-empty.

PaulJonasJost (approved):
Slip up on my side, thank for correcting :)

PR #930, opened 2022-09-08
Turn off default auto saving

PaulJonasJost:


yannikschaelte (approved):
make sense I think, as discussed in the developer meeting :+1: 
(note: at some point, we should then remove some `filename=None` in notebooks and tests too.

PR #933, opened 2022-09-08
Simplify Julia CI test

yannikschaelte:
As we do not care about the type of model, remove DifferentialEquation.jl for the moment, to speed up CI.

### DIFF
--- a/doc/example/model_julia/LR.jl
+++ b/doc/example/model_julia/LR.jl
@@ -2,6 +2,9 @@
 
 module LR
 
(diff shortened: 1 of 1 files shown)


dweindl (approved):
Can't say much on the Julia part, but looks good to me.

yannikschaelte:
> Can't say much on the Julia part, but looks good to me.

Julia should be ok ^^ Main point is 16 minutes --> 2 minutes runtime.

PR #936, opened 2022-09-08
Allow problem=None in read_result_from_file

dweindl:
Allows loading some result without having the corresponding `Problem` around.

### DIFF
--- a/pypesto/optimize/load.py
+++ b/pypesto/optimize/load.py
@@ -3,6 +3,7 @@
 import logging
 import os
 from pathlib import Path
+from typing import Union
 
 import h5py
 import numpy as np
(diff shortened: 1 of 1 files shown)


yannikschaelte (approved):
I'm not up to date with loading. Definitely makes sense to import without a problem at hand. Maybe raise an info/debug that in this case some visualization routines may work differently, as they lack info e.g. about boundaries?

dweindl:
> Maybe raise an info/debug that in this case some visualization routines may work differently, as they lack info e.g. about boundaries?

As far as I can tell, it's not stored, but only used to filter the History for feasible points and then thrown away.

pypesto/optimize/load.py
```diff
@@ -114,7 +115,7 @@ def fill_result_from_history(
 
 
 def read_result_from_file(
-    problem: Problem,
+    problem: Union[Problem, None],
```
FFroehlich, line 118:
```suggestion
    problem: Optional[Problem],
```

PaulJonasJost (approved):

PR #938, opened 2022-09-09
Fix flaky test_mpipoolengine

dweindl:
See e.g. https://github.com/ICB-DCM/pyPESTO/runs/8266214721?check_suite_focus=true


yannikschaelte (approved):
This serves to use n cores even if not enough cores are available on the runner?

dweindl:
> This serves to use n cores even if not enough cores are available on the runner?

It will use N processes even if fewer than N cores are available.

PR #946, opened 2022-10-05
Speed-up import via lazy imports

### LINKED ISSUE #934: pypesto import is slow
`python -c "import pypesto"` takes 7s on my computer. 
`python -c "import scipy"` takes 0.5s on my computer. 

I think we should think more carefully about what is imported in `pypesto/__init__.py`

yannikschaelte:
fixes #934


yannikschaelte:
Substantially faster now again (19s -> 3.5s, or 8s -> 1.3s). E.g. amici+petab not lazified yet, could be done but leaving for the moment as pypesto's use of those tools is quite strong.

dweindl (approved):
Much better, thanks. On-demand import of amici would be nice, but fine to merge as is. Definitely an improvement.

yannikschaelte:
@dweindl now I also lazified amici+pypesto, in version (2) with local imports. does this seem fine to keep?

dweindl (approved):
:+1:

PR #954, opened 2022-10-24
Cooperative enhanced scatter search

### LINKED ISSUE #284: Interfacing CeSS
I am working with a model for which multistart optimization does not find the ground truth optimum after 100 runs and 2.3 days of computation (with most found local minima being different from each other). However, CeSS global + adjoint sensitivity based local search finds the ground truth optimum after 1.3 days. I was therefore wondering if there are any plans to implement CeSS in pyPESTO (or parPE; see corresponding issue there).

dweindl:
Cooperative enhanced scatter search (CESS) and some fixes to ESS (e.g. bounds handling and option to pass custom refset).

Continuation of #941. Closes #284.


pypesto/optimize/ess/cess.py
```diff
@@ -0,0 +1,340 @@
...
+      [2.5, 3.5], with a recommended default value of 2.5.
+
+    .. [VillaverdeEge2012] 'A cooperative strategy for parameter estimation in
+       large scale systems biology models', Villaverde, A.F., Egea,
+       J.A. & Banga, J.R. BMC Syst Biol 2012, 6, 75.
+       https://doi.org/10.1186/1752-0509-6-75
+
+    Attributes
+    ----------
+    ess_init_args:
```
yannikschaelte, line 47: duplicate of init docstring?

pypesto/optimize/ess/cess.py
```diff
@@ -0,0 +1,340 @@
...
+        problem: Problem,
+        startpoint_method: StartpointMethod,
+        refset: RefSet,
+    ) -> ESSOptimizer:
+        """
+        Run ESS.
+
+        Helper for `starmap`.
+        """
+        # different random seeds per prcess?
```
yannikschaelte, line 282:
```suggestion
        # different random seeds per process?
```

pypesto/optimize/ess/cess.py
```diff
@@ -0,0 +1,340 @@
...
+        np.random.seed((os.getpid() * int(time.time() * 1000)) % 2**32)
```
yannikschaelte, line 283: setting the random seed (which is otherwise inherited) new per process would be good. is it possible to instead use an own random number generator, as np.random.seed is deprecated?

pypesto/optimize/ess/cess.py
```diff
@@ -0,0 +1,340 @@
...
+        ess = ESSOptimizer(**ess_kwargs)
+        ess.minimize(problem, startpoint_method, refset=refset)
+        return ess
+
+    def _keep_going(self, i_eval) -> bool:
+        """Check exit criteria.
+
+        Returns
+        -------
+        ``True`` if not of the exit criteria is met, ``False`` otherwise.
```
yannikschaelte, line 294:
```suggestion
        ``True`` if none of the exit criteria is met, ``False`` otherwise.
```

pypesto/optimize/ess/ess.py
```diff
@@ -297,7 +305,8 @@ def _combine_solutions(self) -> Tuple[np.array, np.array]:
         """Combine solutions and evaluate.
 
         Creates the next generation from the RefSet by pair-wise combinations
-        of all RefSet members.
+        of all RefSet members. Creates ``RefSet.dim ** 2 - RefSet.dim`` new
+        parameter vectors, test them, and keeps the best child of each parent.
```
yannikschaelte, line 309:
```suggestion
        parameter vectors, tests them, and keeps the best child of each parent.
```

yannikschaelte (approved):
just typoss

pypesto/optimize/ess/cess.py
dweindl: Class attributes vs. __init__ arguments. But, yes, significant overlap.

pypesto/optimize/ess/cess.py
dweindl: I choose this because it seems to be the one used for starting point sampling in pypesto which is the most relevant RNG application here.

```
grep -r np.random pypesto/startpoint/
pypesto/startpoint/latin_hypercube.py:        xs[:, i_dim] = np.random.permutation(np.arange(n_starts))
pypesto/startpoint/latin_hypercube.py:        xs += np.random.random((n_starts, dim))
pypesto/startpoint/uniform.py:    xs = np.random.random((n_starts, dim))
```

Seeding any other RND would have no effect, unless this is also changed elsewhere. Maybe replace all of them separately later?

dweindl: Ah, you mean being able to supply any other random regenerator? I guess that would only be relevant in combination with a custom starting point method that uses a different RNG than the rest of pypesto. Could be done. But I'd generalize that then to an arbitrary init function that can be passed, independent of RNGs.

yannikschaelte: ah if it's the pypesto starting point generation, then yes we would need to change it there too. good to leave for the moment, probably at some point we should go over all np.random samplers and supply proper random number generators.

PR #956, opened 2022-10-26
Fix AMICI deprecation warning

dweindl:
during PEtab import.

Waiting for https://github.com/AMICI-dev/AMICI/pull/1877


FFroehlich (approved):
👍

PR #958, opened 2022-10-28
GHA: Update actions

### LINKED ISSUE #957: Plenty of GitHub Actions warnings
Seems like we should use newer versions of some actions:

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/setup-python, actions/cache, actions/cache, actions/checkout

dweindl:
Fixes #957

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,15 +20,15 @@ jobs:
 
     steps:
     - name: Check out repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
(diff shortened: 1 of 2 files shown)


dweindl:
ping @yannikschaelte @PaulJonasJost

PaulJonasJost (approved):
thanks

yannikschaelte (approved):
can they just have a no version that defaults to latest

PR #961, opened 2022-11-04
emcee sampler: initialize walkers near optimum

dilpath:
As recommended:
- https://msp.org/camcos/2010/5-1/p04.xhtml
- https://emcee.readthedocs.io/en/stable/user/faq/

Old code is use if multiple vectors are provided to `Sampler.initialize`


yannikschaelte (approved):
:+1:

PR #963, opened 2022-11-04
dynesty Sampler

dilpath:
https://dynesty.readthedocs.io/en/stable/

Nested sampling inputs/results differ from other MCMC methods in pyPESTO, so there might be some issues with using `dynesty` samplers with pyPESTO code. e.g. removing burn-in doesn't make sense.

Some additional things would be nice but it remains to be seen whether they are easy to implement and how useful they are. For example, somehow providing gradients to `dynesty` would be useful if Hamiltonian slice sampling is selected -- currently, it appears `dynesty` will use finite differences. `dynesty` supports a user-provided gradient function, but this is called in addition to the objective (`loglikelihood`) function call.


pypesto/sample/dynesty.py
```diff
@@ -0,0 +1,170 @@
+# TODO set `nlive` to > a few * n_params**2
```
yannikschaelte, line 1: file docstring

pypesto/sample/dynesty.py
```diff
@@ -0,0 +1,170 @@
...
+
+        self.problem = problem
+
+        def loglikelihood(x):
+            """Log-probability density function."""
+            # check if parameter lies within bounds
+            if any(x < self.problem.lb) or any(x > self.problem.ub):
+                return -np.inf
+            # invert sign
+            # FIXME this is possibly the posterior if priors are defined
```
yannikschaelte, line 119: I think we are only using TODO as annotation to track. thus maybe todo fixme?

yannikschaelte (approved):
:+1:

PR #967, opened 2022-11-09
Ensure model selection models are compatible with available AMICI model

dilpath:
TODO check for compatibility with AMICI model in `ModelProblem`


dilpath:
fixed in #956

PR #969, opened 2022-11-10
Model fit visualization: use `result.problem.objective` to simulate, instead of AMICI directly

dilpath:
This makes it possible to visualize e.g. results from hierarchical optimization, since the inner parameters are computed and supplied to AMICI on the fly.

Breaks compatibility with previous use of this method.


yannikschaelte (approved):
fine by me. only point: it was never really planned to have the objective as part of the result. thus preferably pass separately?

dilpath:
> fine by me. only point: it was never really planned to have the objective as part of the result. thus preferably pass separately?

:+1:

Issue #167, opened 2019-04-11
Implement Hierarchical Optimization


yannikschaelte:
As started in some (little bit stale) branch already, implement hierarchical optimization covering a few standard settings (scalings, offsets, sigmas), using forward and adjoint sensis.

The idea is to define an HierarchicalAmiciCalculator, which hops inside the amici_objective call function, and which performs the actual calculation of llh, sllh, etc., and which in the default case, without static, i.e. hierarchical. parameters, just calls amici, but which can inside the amici_objective be replaced by a Calculator which does hierarchical optimization in different ways.

This way should be minimum invasive and transparent too. Will be easy to define the hierarchical problem using petab. Should also be possible to generalize to apply an ordinal hierarchical problem.

dweindl:
There are currently 4 rather stale branches related to that :(

https://github.com/ICB-DCM/pyPESTO/tree/feature_ordinal
https://github.com/ICB-DCM/pyPESTO/tree/feature_qual_gradients
https://github.com/ICB-DCM/pyPESTO/tree/feature_hierarchical
https://github.com/ICB-DCM/pyPESTO/tree/hierarchical_with_simulation_chunks_PJ

dweindl:
All done, right?

PR #971, opened 2022-11-15
Parameter estimation from qualitative data

dweindl:
Contains (parts of?) @LeonardSchmiester's work on parameter estimation with qualitative data. Needs to reconciled with the current hierarchical optimization interface. See also #167 and https://github.com/ICB-DCM/pyPESTO/tree/feature_qual_gradients.


Doresic:
Completely implemented original and gradient-based optimal scaling method for integrating ordinal data into parameter estimation including example notebook and tests, producing results as in Schmiester et al. 2021. 
Most of the implementation is in `/pypesto/hieararchical/optimal_scaling_approach`. Except that, there are changes in the `/pypesto/petab/importer.py` and the constants file `C.py`. 

PEtab formulation change with respect to the former implementation:
- Formerly the optimal scaling inner parameters needed to be specified explicitly in the `parameters.tsv` file. This was a tedious and long process in most cases.
- Now the PEtab specification differs from the standard one only in the measurement file, where ordinal measurements have to be specified as such with their respective groups and categories.

Now the optimal scaling inner parameters are being saved into the optimizer_result history. This is currently implemented only for csv histories. This involves very little change in implementation and can be implemented for other hierarchical approaches as well.

A couple of additional notes:
- The` /doc/example/example_ordinal` PEtab model used by `test/hierarchical/test_ordinal.py` and  /doc/example/example_ordinal.ipynb` is not in the form it should be ideally: the measurement column is not filled with NaNs as it should be for ordinal measurements because of a petab.lint check which fails (https://github.com/PEtab-dev/libpetab-python/blob/main/petab/lint.py#:~:text=def%20assert_measurements_not_null(,null%20(missing).%27))
- There are no validate_petab_ordinal (analog to validate_petab_hierarchical) checks implemented yet.

pypesto/C.py
```diff
@@ -94,12 +95,14 @@ class InnerParameterType(str, Enum):
     OFFSET = 'offset'
     SCALING = 'scaling'
     SIGMA = 'sigma'
+    OPTIMALSCALING = 'optimal_scaling'
```
yannikschaelte, line 98:
```suggestion
    OPTIMAL_SCALING = 'optimal_scaling'
```

pypesto/C.py
```diff
@@ -94,12 +95,14 @@ class InnerParameterType(str, Enum):
...
 
 
 DUMMY_INNER_VALUE = {
     InnerParameterType.OFFSET: 0.0,
     InnerParameterType.SCALING: 1.0,
     InnerParameterType.SIGMA: 1.0,
+    InnerParameterType.OPTIMALSCALING: 0.0,
```
yannikschaelte, line 105:
```suggestion
    InnerParameterType.OPTIMAL_SCALING: 0.0,
```

pypesto/C.py
```diff
@@ -115,8 +118,26 @@ class InnerParameterType(str, Enum):
         LOWER_BOUND: 0,
         UPPER_BOUND: float('inf'),
     },
+    InnerParameterType.OPTIMALSCALING: {
```
yannikschaelte, line 121:
```suggestion
    InnerParameterType.OPTIMAL_SCALING: {
```

pypesto/hierarchical/optimal_scaling_approach/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""
+Optimal scaling
+============
```
yannikschaelte, line 3:
```suggestion
Optimal scaling
===============
```

pypesto/hierarchical/optimal_scaling_approach/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+
+Optimal scaling optimization sub-package.
```
yannikschaelte, line 5: mention and link where appropriate also the papers where details on the implementation can be found

pypesto/hierarchical/optimal_scaling_approach/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""
```
yannikschaelte, line 1: would a more descriptive (and shorter) name of the package be `ordinal` instead of `optimal_scaling_approach`?

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_calculator.py
```diff
@@ -0,0 +1,170 @@
...
+    import amici
+    from amici.parameter_mapping import ParameterMapping
+except ImportError:
+    pass
+
+
+class OptimalScalingAmiciCalculator:
+    """A calculator is passed as `calculator` to the pypesto.AmiciObjective.
+
+    While this class cannot be used directly, it has two subclasses
```
yannikschaelte, line 25: This does not seem to be the case: (1) this class has no abstract methods, (2) there seem to be no forward/adjoint subclasses. Legacy comment?

yannikschaelte, line 25: This does not seem to be the case: (1) this class has no abstract methods, (2) there seem to be no forward/adjoint subclasses. Legacy comment?

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_calculator.py
```diff
@@ -0,0 +1,170 @@
...
+        x_inner_opt = self.inner_solver.solve(self.inner_problem, sim)
+        inner_result[FVAL] = self.inner_solver.calculate_obj_function(
+            x_inner_opt
+        )
+        inner_result[
+            X_INNER_OPT
+        ] = self.inner_problem.get_inner_parameter_dictionary()
+
+        # calculate analytical gradients if requested
+        if sensi_order > 0:
```
yannikschaelte, line 157: check if sensi_order == 2 (Hessian) and fail? As FIM cannot be calculated. (Fail at beginning of function)

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_parameter.py
```diff
@@ -0,0 +1,52 @@
+import logging
```
yannikschaelte, line 1: Ideally add a 1 line description at the beginning of each file what it is about.

sth like ``"""Definition of an optimal scaling parameter class."""``

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_parameter.py
```diff
@@ -0,0 +1,52 @@
...
+        **kwargs,
+    ):
+        """Construct.
+
+        Parameters
+        ----------
+        See class attributes.
+        """
+        super().__init__(*args, **kwargs)
+        if self.inner_parameter_type != InnerParameterType.OPTIMALSCALING:
```
yannikschaelte, line 39:
```suggestion
        if self.inner_parameter_type != InnerParameterType.OPTIMAL_SCALING:
```

yannikschaelte, line 39:
```suggestion
        if self.inner_parameter_type != InnerParameterType.OPTIMAL_SCALING:
```

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_parameter.py
```diff
@@ -0,0 +1,52 @@
...
+            raise ValueError(
+                "For the OptimalScalingParameter class, the parameter type has to be optimal_scaling."
```
yannikschaelte, line 41:
```suggestion
                "For the OptimalScalingParameter class, the parameter type has to be "
                f"{InnerParameterType.OPTIMAL_SCALING}."
```

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_parameter.py
```diff
@@ -0,0 +1,52 @@
...
+            )
+
+        if group is None:
```
yannikschaelte, line 44: does the constructor need None default arguments? If not, please remove the default args there (can still keep this check to be sure).

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_parameter.py
```diff
@@ -0,0 +1,52 @@
...
+        Group index.
+    value:
+        Current value of the inner parameter.
+    estimate:
+        Whether to estimate inner parameter in inner subproblem.
+    """
+
+    def __init__(
+        self,
+        *args,
```
yannikschaelte, line 26: are the *args, **kwargs arguments needed here? If the original parameters need to be passed (such as id, lb, ub, dumy_value), I think it is better design to keep them here explicitly, as *args, **kwargs may hide stuff from the developer.

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+            data:
+                The data of the problem.
+            method:
+                A string representing the method of the Optimal Scaling approach, either 'reduced' or 'standard'.
+        """
+        super().__init__(xs, data)
+        self.groups = {}
+        self.method = method
+
+        for group in self.get_groups_for_xs(InnerParameterType.OPTIMALSCALING):
```
yannikschaelte, line 57:
```suggestion
        for group in self.get_groups_for_xs(InnerParameterType.OPTIMAL_SCALING):
```

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+        self, group: int
+    ) -> List[OptimalScalingParameter]:
+        """Get ``OptimalScalingParameter``s that are category lower boundaries and belong to the given group."""
+        return [
+            x
+            for x in self.xs.values()
+            if x.group == group and x.inner_parameter_id[:6] == 'cat_lb'
+        ]
+
+    def initialize_c(self, group):
```
yannikschaelte, line 169: indicate return type

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+        self.groups = {}
+        self.method = method
+
+        for group in self.get_groups_for_xs(InnerParameterType.OPTIMALSCALING):
+            self.groups[group] = {}
+            xs = self.get_xs_for_group(group)
+            self.groups[group]['num_categories'] = len(
+                {x.category for x in xs}
+            )
+            self.groups[group]['num_datapoints'] = np.sum(
```
yannikschaelte, line 63: for list inputs, python `sum()` should be more efficient than first converting to `np.ndarray` for `np.sum()`

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+            self.groups[group]['W'] = self.initialize_w(group)
+
+            self.groups[group]['Wdot'] = self.initialize_w(group)
+
+    @staticmethod
+    def from_petab_amici(
+        petab_problem: petab.Problem,
+        amici_model: 'amici.Model',
+        edatas: List['amici.ExpData'],
+        method: str = None,
```
yannikschaelte, line 111:
```suggestion
        method: str = REDUCED,
```

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+    ):
+        """Construct the inner problem from the `petab_problem`."""
+        if method is None:
+            method = REDUCED
```
yannikschaelte, line 115:
```suggestion
```

yannikschaelte, line 115:
```suggestion
```

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+            )
+
+            self.groups[group]['C'] = self.initialize_c(group)
+
+            self.groups[group]['W'] = self.initialize_w(group)
+
+            self.groups[group]['Wdot'] = self.initialize_w(group)
+
+    @staticmethod
+    def from_petab_amici(
```
yannikschaelte, line 107: return type

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+    ) -> List[OptimalScalingParameter]:
+        """Get ``OptimalScalingParameter``s that are category lower boundaries and belong to the given group."""
+        return [
+            x
+            for x in self.xs.values()
+            if x.group == group and x.inner_parameter_id[:6] == 'cat_lb'
+        ]
+
+    def initialize_c(self, group):
+        """Initialize the constraints matrix for the group."""
```
yannikschaelte, line 170: a (colloquial/detailed in words) description of what this function does / how it initializes the matrix?

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+        constr = np.zeros(
+            [
+                self.groups[group]['num_constr_full'],
+                self.groups[group]['num_inner_params'],
+            ]
+        )
+        data_idx = 0
+        for cat_idx, x in enumerate(
```
yannikschaelte, line 178: here some more code annotation would be good imo ...

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+                        * np.sum(sy_all)
+                        / ((np.sum(np.abs(y_sim_all)) + 1e-8) ** 2)
+                    ),
+                    np.zeros(2 * self.groups[gr]['num_categories']),
+                ]
+            )
+        )
+        return weights
+
+    def get_w_squared(self, gr, y_sim_all):
```
yannikschaelte, line 279: what does the function calculate (informal description)?

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_solver.py
```diff
@@ -0,0 +1,831 @@
...
+            Ids of outer simulation parameters, includes fixed parameters.
+        snllh:
+            Empty dictionary with optimization parameters as keys.
+
+        Returns
+        -------
+        Filled in snllh dictionary with objective function gradients.
+        """
+        condition_map_sim_var = parameter_mapping[0].map_sim_var
+        # TODO: Doesn't work with condition specific parameters
```
yannikschaelte, line 149: todo? if left for further work, make issue and reference it here.

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_solver.py
```diff
@@ -0,0 +1,831 @@
...
+        for par_sim, par_opt in condition_map_sim_var.items():
+            if not isinstance(par_opt, str):
+                continue
+            par_sim_idx = par_sim_ids.index(par_sim)
+            par_opt_idx = par_opt_ids.index(par_opt)
+            grad = 0.0
+
+            # Iterate over inner parameter groups.
+            for idx, group in enumerate(
+                problem.get_groups_for_xs(InnerParameterType.OPTIMALSCALING)
```
yannikschaelte, line 161:
```suggestion
                problem.get_groups_for_xs(InnerParameterType.OPTIMAL_SCALING)
```

pypesto/hierarchical/parameter.py
```diff
@@ -70,6 +70,7 @@ def __init__(
         self.scale = scale
 
         if inner_parameter_type not in (
+            InnerParameterType.OPTIMALSCALING,
```
yannikschaelte, line 73:
```suggestion
            InnerParameterType.OPTIMAL_SCALING,
```

pypesto/hierarchical/petab.py
```diff
@@ -66,6 +66,7 @@ def validate_hierarchical_petab_problem(petab_problem: petab.Problem) -> None:
                 InnerParameterType.OFFSET,
                 InnerParameterType.SIGMA,
                 InnerParameterType.SCALING,
+                InnerParameterType.OPTIMALSCALING,
```
yannikschaelte, line 69:
```suggestion
                InnerParameterType.OPTIMAL_SCALING,
```

pypesto/hierarchical/optimal_scaling_approach/__init__.py
yannikschaelte: and add some text on the problem this sub-module solves (ordinal data, scale, hierarchical, ...)

pypesto/hierarchical/problem.py
```diff
@@ -6,6 +6,8 @@
 import pandas as pd
 
 from ..C import PARAMETER_TYPE, InnerParameterType
+
+# from ..C import PARAMETER_CATEGORY, PARAMETER_GROUP
```
yannikschaelte, line 10:
```suggestion

```

pypesto/petab/importer.py
```diff
@@ -64,6 +69,7 @@ def __init__(
         validate_petab: bool = True,
         validate_petab_hierarchical: bool = True,
         hierarchical: bool = False,
+        ordinal: bool = False,
```
yannikschaelte, line 72: add to docstring

pypesto/petab/importer.py
```diff
@@ -417,6 +424,31 @@ def create_objective(
             # FIXME: currently not supported with hierarchical
             kwargs['guess_steadystate'] = False
 
+        if self._ordinal:
+            # TODO add constants to C
```
yannikschaelte, line 428: todo

pypesto/petab/importer.py
```diff
@@ -417,6 +424,31 @@ def create_objective(
...
+            if 'inner_problem_method' in kwargs:
+                inner_problem_method = kwargs.pop('inner_problem_method')
```
yannikschaelte, line 430: is popping necessary? usually unnecessary

pypesto/petab/importer.py
```diff
@@ -417,6 +424,31 @@ def create_objective(
...
+            else:
+                inner_problem_method = None
```
yannikschaelte, line 432:
```suggestion
            inner_problem_method = kwargs.get('inner_problem_method', None)
```

pypesto/petab/importer.py
```diff
@@ -417,6 +424,31 @@ def create_objective(
...
+            if 'inner_solver_options' in kwargs:
+                inner_solver_options = kwargs.pop('inner_solver_options')
+            else:
+                inner_solver_options = None
```
yannikschaelte, line 436:
```suggestion
            inner_solver_options = kwargs.get('inner_solver_options', None)
```

yannikschaelte, line 436:
```suggestion
            inner_solver_options = kwargs.get('inner_solver_options', None)
```

pypesto/petab/importer.py
```diff
@@ -417,6 +424,31 @@ def create_objective(
...
+            inner_solver = OptimalScalingInnerSolver(
+                options=inner_solver_options
+            )
+            calculator = OptimalScalingAmiciCalculator(
+                inner_problem, inner_solver
+            )
+            amici_reporting = amici.RDataReporting.full
+
+            # FIXME: currently not supported with hierarchical
+            kwargs['guess_steadystate'] = False
```
yannikschaelte, line 450: also add warning? might cause unexpected behavior

pypesto/petab/importer.py
```diff
@@ -624,13 +656,20 @@ def create_problem(
...
         ub = self.petab_problem.ub_scaled
 
-        # In case of hierarchical optimization, parameters estimated in the
-        # inner subproblem are removed from the outer problem
+        # Raise error if the correct calculator is not used.
         if self._hierarchical:
             if not isinstance(
                 objective.calculator, HierarchicalAmiciCalculator
             ):
                 raise AssertionError()
```
yannikschaelte, line 664: please add informative error message here, also below, thanks :)

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,103 @@
...
+import numpy as np
+import petab
+import pytest
+
+import pypesto
+import pypesto.logging
+import pypesto.optimize
+import pypesto.petab
+
+inner_solver_options = [
```
yannikschaelte, line 13: this can surely be reduced (+ readability improved) with e.g. a few list comprehensions ...

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,103 @@
...
+
+def test_optimization(inner_solver_options: List[Dict]):
+    """Check that optimizations finishes without error."""
+    petab_problem = petab.Problem.from_yaml(example_ordinal_yaml)
+
+    optimizer = pypesto.optimize.ScipyOptimizer(
+        method='Nelder-Mead', options={'maxiter': 10}
+    )
+    for option in inner_solver_options:
+        problem = create_problem(petab_problem, option)
```
yannikschaelte, line 86:
```suggestion
        problem = _create_problem(petab_problem, option)
```

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,103 @@
...
+        pypesto.optimize.minimize(
+            problem=problem, n_starts=1, optimizer=optimizer
+        )
+
+
+def create_problem(
```
yannikschaelte, line 92:
```suggestion
def _create_problem(
```

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,103 @@
...
+    petab_problem: petab.Problem, option: Dict
+) -> pypesto.Problem:
+    importer = pypesto.petab.PetabImporter(petab_problem, ordinal=True)
```
yannikschaelte, line 95: docstring

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
yannikschaelte, line 111:
```suggestion
        method: str = REDUCED,
```

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+                [
+                    np.ones(self.groups[gr]['num_datapoints'])
+                    / (np.sum(np.square(y_sim_all)) + 1e-8),
+                    np.zeros(2 * self.groups[gr]['num_categories']),
+                ]
+            )
+        )
+        return weights
+
+    def get_wdot_squared(self, gr, y_sim_all, sy_all):
```
yannikschaelte, line 292: explain difference to previous function

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
```diff
@@ -0,0 +1,530 @@
...
+        ] = (interval_gap + eps)
+
+        d[
+            2 * self.groups[gr]['num_datapoints']
+            + self.groups[gr]['num_categories'] :
+        ] = interval_range
+        return d
+
+    def get_dd_dtheta(self, gr, xs, y_sim_all, sy_all):
+        """Return the derivative of vector of minimal gaps and ranges with respect to a outer parameter."""
```
yannikschaelte, line 332:
```suggestion
        """Return the derivative of vector of minimal gaps and ranges with respect to an outer parameter."""
```
(several places)

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,103 @@
...
+    importer.create_model()
+
+    objective = importer.create_objective(
+        inner_problem_method=option['method'],
+        inner_solver_options=option,
+    )
+    problem = importer.create_problem(objective)
+    return problem
```
yannikschaelte, line 103: I think that given the complexity of the method, a substantially increased set of actual unit tests (i.e. test single components - functions, classes - instead of the full workflow) would be much more suited to ensuring correctness of the approach - e.g. regarding the solution of the inner problem, the grouping, ...

I am rn not sure what would constitute good unit tests. are there worked examples where you manually derived the true solutions? then one could use those here. otherwise we could maybe define a set of useful unit tests.

yannikschaelte (changes requested):

yannikschaelte:
very nicely written! I did not yet check the methods implementation in detail, I think these would benefit from some more unit tests that ensure correctness of the implementation (see detailed comments).

doc/example/example_ordinal.ipynb
```diff
@@ -0,0 +1,1047 @@
...
+   "source": [
+    "# Parameter estimation using ordinal data"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This Notebook eplains the use of ordinal data for parameter estimation, as described in [Schmiester et al. (2019)](https://www.biorxiv.org/content/10.1101/848648v1.abstract). An example model is provided in `example_ordinal`"
```
dilpath, line 16:
```suggestion
    "This Notebook eplains the use of ordinal data for parameter estimation, as described in [Schmiester et al. (2019)](https://www.biorxiv.org/content/10.1101/848648v1.abstract). An example model is provided in `pypesto/doc/example/example_ordinal`"
```

Link if possible, not important

doc/example/example_ordinal.ipynb
```diff
@@ -0,0 +1,1047 @@
...
+   "source": [
+    "np.random.seed(n_starts)\n",
+    "problem.objective.calculator.inner_solver = OptimalScalingInnerSolver(\n",
+    "    options={\n",
+    "        'method': 'reduced',\n",
+    "        'reparameterized': True,\n",
+    "        'intervalConstraints': 'max',\n",
+    "        'minGap': 1e-10,\n",
+    "    }\n",
+    ")\n",
```
dilpath, line 304: Maybe better to provide these `options` to `PetabImporter(..., ordinal=True, inner_solver_options=...)` so that the user doesn't need to import `OptimalScalingInnerSolver` and do these additional steps

doc/example/example_ordinal.ipynb
```diff
@@ -0,0 +1,1047 @@
...
+     ]
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Optimization terminated successfully.\n",
+      "         Current function value: 0.000574\n",
+      "         Iterations: 146\n",
+      "         Function evaluations: 271\n"
```
dilpath, line 131: Fine to output this data in one cell if it's somehow especially useful to a user using your work, but otherwise would be good to disable by default and/or mute in other cells of the notebook

doc/example/example_ordinal.ipynb
dilpath: Ah, I see `PetabImporter.create_objective` takes this as input, I guess it would be good to allow users to also specify in `__init__` as above, for simpler code.

dilpath: Also, for the notebook, maybe better to run with default options in the first case, then import `OptimalScalingInnerSolver` later to show how it can be customized in the next cases

dilpath: Fine as is, if you don't want to showcase default behavior

doc/example/example_ordinal.ipynb
```diff
@@ -0,0 +1,1047 @@
...
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "from pypesto.visualize import waterfall\n",
+    "\n",
+    "waterfall([res_standard, res_reduced, res_reduced_reparameterized])"
```
dilpath, line 804: Assuming you use the same starts for each result: adding `waterfall(..., order_by_id=True)` changes the waterfall plot to ensure the same starts are plotted at the same `x-positions`. Won't change your result here since there appears to be no difference between settings. The waterfall plot says they're all the same but it's difficult to tell. Perhaps a table of fvals makes this clearer?

doc/example/example_ordinal.ipynb
```diff
@@ -0,0 +1,1047 @@
...
+    "    options={\"disp\": None, \"ftol\": 2.220446049250313e-09, \"gtol\": 1e-5},\n",
+    ")"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Running optimization with a gradient-based optimizer"
```
dilpath, line 833: Why is the notebook duplicated for two different optimizers? Can just show the gradient-based example, then specify in words that other optimizers, including gradient-free, also work fine?

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_solver.py
```diff
@@ -0,0 +1,831 @@
...
+        elif x_category > 1:
+            x_lower = optimal_scaling_bounds[x_category - 2] + interval_gap
+        else:
+            raise ValueError('Category value needs to be larger than 0.')
+    elif options['method'] == STANDARD:
+        x_lower = optimal_scaling_bounds[2 * x_category - 2]
+        x_upper = optimal_scaling_bounds[2 * x_category - 1]
+    else:
+        raise NotImplementedError(
+            f"Unkown optimal scaling 'method' {options['method']}. "
```
dilpath, line 768:
```suggestion
            f"Unknown optimal scaling 'method' {options['method']}. "
```

pypesto/petab/importer.py
```diff
@@ -652,10 +691,11 @@ def create_problem(
...
 
         if prior is not None:
-            if self._hierarchical:
+            if self._hierarchical or self._ordinal:
                 raise NotImplementedError(
-                    "Hierarchical optimization in combination "
-                    "with priors is not yet supported."
+                    "Hierarchical optimization or Optimal scaling"
+                    "optimization in combination with priors"
+                    "is not yet supported."
```
dilpath, line 698:
```suggestion
                    "Hierarchical or optimal scaling "
                    "optimization in combination with priors "
                    "is not yet supported."
```

dilpath (approved):
Nice, thanks a lot! More tests would be great

pypesto/hierarchical/optimal_scaling_approach/__init__.py
Doresic: Good description?

Contains the implementation of the optimal scaling approach, applied for integration
of ordinal data in ODE modeling. Each ordinal datapoint is defined as being part of 
a category, where the mutual ordering of the categories of the same group is known.
The category interval bounds are numerically optimized and quantitative surrogate
measurements are calculated to represent the ordinal measurements. This constitutes
the inner subproblem of the hierarchical optimization problem.

Details on the optimal scaling approach can be found in Shepard, 1962, https://doi.org/10.1007/BF02289621
Details on the application of the gradient-based optimal scaling approach to mechanistic modeling
with ordinal data can be found in Schmiester et al. 2020 (https://doi.org/10.1007/s00285-020-01522-w)
and Schmiester et al. 2021 (https://doi.org/10.1093/bioinformatics/btab512).

pypesto/hierarchical/optimal_scaling_approach/__init__.py
Doresic: I was thinking about this for a while and later discussed it with Daniel. 
On one hand, the method or approach that's implemented is the optimal scaling one and it just deals with ordinal data. So it would be more correct to name it like it is in this sense.

On the other hand, it would make it much simpler if it was all just called ordinal, since it's the specific application to ordinal data, and the optimal scaling currently cannot be used for some other data.

But then when I'll implement the optimal scaling approach for censored data. I'll probably just extend the optimal scaling approach to detect the censored data and include it (in the same module).
Or I can separate it and call that one censored (might be easier to read and understand what's handling what).

So not sure what's best. Open to suggestions/other namings.

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_calculator.py
Doresic: Yes, was left over from hierarchical by accident. Will change.
On that note, is this currently accurate for the hierarchical calculator (HierarchicalAmiciCalculator)?

pypesto/hierarchical/optimal_scaling_approach/__init__.py
yannikschaelte: So none of the names is good :sweat_smile: My recommendations:

1. Keep it short. Thus, definitely remove the `_approach`.
2. Whether to call it `ordinal` or `optimal_scaling` (or `rescaling` or `scaling` or `scale`) I would let depend on whether you intend to implement the censored case in the same submodule (if it uses much of the same base code) (--> optimal_scaling) or implement it as a separate submodule (-->ordinal).

pypesto/hierarchical/optimal_scaling_approach/__init__.py
yannikschaelte: Perfect :+1: If you want, can even add more details in the respective code segments / functions, but this level of description will be good here.

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_calculator.py
yannikschaelte: That does indeed not seem to be the case, the problem has been solved by just calling super() at the end. Could you also remove the comment there?

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
Doresic: This was a laugh, interesting that this came first to mind to do

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
Doresic: Don't seem to be able to do it, since it returns the objects it's a function of (OptimalScalingProblem). It's done in the same way in hierarchical/problem.py. 
Is there a nicer/better definition of this?

yannikschaelte: I meant indicate the type of the return object, in `): -> {TYPE}`

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
Doresic: I've tested some alternative weights of the objective function, and they were not performing as well, so I didn't include them in the implementation of the solver, but I left the weight calculating functions.

I'm going to remove them as they're not being used currently and are easy to implement later if needed.

yannikschaelte: Ah, if these are just alternative weighting functions, also fine to keep them included, and clearly mark which one is used, and which ones are just alternatives.

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_solver.py
Doresic: Yes, it's a TODO still from Leonard. Will open an issue.

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
Doresic: Yes, the return type is `OptimalScalingProblem`, which is the class this function is a part of. So I don't seem to be able to indicate it.

pypesto/hierarchical/optimal_scaling_approach/__init__.py
Doresic: Yeah, true,`_approach` is definitely not important.

Ok, I'll leave further renaming for when I decide how's best to implement censored data.

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
yannikschaelte: Ah then just return it as string `'OptimalScalingProblem'`

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_calculator.py
Doresic: Sure, will do.

pypesto/hierarchical/optimal_scaling_approach/__init__.py
Doresic: Further, on the point of naming: I saw a pattern in pypesto that even though a python file is specific to a module it still has a general name, for instance:
petab checks for hierarchical are in the file hierarchical/petab.py and not  hierarchical/hierarchical_petab.py
or the parameter is hierarchical/parameter.py and not something like hierarchical/inner_parameter.py

So should all things in the optimal_scaling folder be just calculator, solver, parameter and problem, or be left as optimal_scaling_calculator, ... etc.

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_problem.py
Doresic: Will do the same with hierarchical then.

pypesto/hierarchical/optimal_scaling_approach/__init__.py
yannikschaelte: As the fully qualified path would be `pypesto.hierarchical.{optimal_scaling etc}.solver`, the additional `optimal_scaling` can be omitted indeed. Note: Actually, as all important exports are given in `pypesto.hierarchical.optimal_scaling.__init__`, nobody should ever directly refer to the file name, but only use what is in `__init__`. This could be enforced more strictly by setting files to private, `pypesto.hierarchical.optimal_scaling._solver`, but this we have done nowhere in pypesto either, so one could do it here but also fine without.

pypesto/petab/importer.py
Doresic: As the kwargs are later passed to `AmiciObjective` and it doesn't take any kwargs, it fails there if they're not popped.

pypesto/petab/importer.py
Doresic: This is also in part a question: should these strings be in the constants file, as they're not something used throughout, but just a name of variable passed for constructors? (like 'guess_steadystate' is)

pypesto/petab/importer.py
Doresic: Think get will not work, as described up.
But I put this: 
            inner_problem_method = kwargs.pop(INNER_PROBLEM_METHOD, None)

pypesto/petab/importer.py
Doresic: Think get will not work, as described up.
But I put this: 
            inner_problem_method = kwargs.pop(INNER_SOLVER_OPTIONS, None)

pypesto/hierarchical/optimal_scaling_approach/__init__.py
Doresic: Ok, will rename to only `calculator`, `solver`, `parameter` and `problem`, looks nicer. 
I won't add _ as it wasn't added to anything else, for consistency of the whole package.

test/hierarchical/test_ordinal.py
Doresic: Reduced :D 

```
inner_solver_options = [
    [
        {
            key: value
            for key, value in zip(
                ['method', 'reparameterized', 'intervalConstraints'],
                [method, reparameterized, intervalConstraints],
            )
        }
        for method, reparameterized in zip(
            ['standard', 'reduced', 'reduced'], [False, False, True]
        )
    ]
    for intervalConstraints in ['max', 'max-min']
]
```

test/hierarchical/test_ordinal.py
Doresic: I agree. 

With a particular seed and settings, the parameter inference is consistent with a pretty good optimal plateau for simple models (RAF, Boehm), even though they're not the true parameters (I believe non-importantly for this). I can then write test using optimal outer parameters of these optimizations, for instance:
- check problem is constructed as should be, with correct group dictionaries (inner_problem.groups[group] dictionaries)
- inner optimization results are the ones they should be: surrogate data, lower, upper bounds.
- check fval is the same.
- check that gradient calculated is close to 0 to check solver gradient calculation.

I believe that would be enough/test all features.

pypesto/petab/importer.py
yannikschaelte: so this is about e.g. `inner_problem_method`? if it's just an argument that is used in only exactly this function, fine to keep it as a string literal. only as soon as you use the string literal >1 times, make it a variable.

yannikschaelte: + document this argument in the method docstring

pypesto/petab/importer.py
yannikschaelte: :+1:

doc/example/example_ordinal.ipynb
Doresic: Yes, I was thinking for some time about where it makes the most sense to give the options, so I decided at objective, as it's used there directly. But it can be still changed post-objective creation, as shown in the notebook, by creating a new OptimalScalingInnerSolver basically.

In which __init__ would it be nice to be able to give them as well? 

I agree on the notebook, I'll include the different ways of construction.

doc/example/example_ordinal.ipynb
Doresic: Will mute it, it's too much output for a notebook.

doc/example/example_ordinal.ipynb
Doresic: Didn't know about this setting, neat. Will use it :D 
I added a listing of fvals as well for exactness.

doc/example/example_ordinal.ipynb
Doresic: Yes, can do that, it was only gradient-free, so wanted to show the other one as well. But makes sense to show only one.

pypesto/petab/importer.py
Doresic: Ok, won't add to C then and will add to kwargs part of docstring.

doc/example/example_ordinal.ipynb
dilpath: I meant in `PetabImporter.__init__`, since that's where users can also specify e.g. `hierarchical=True, ordinal=True` etc.

It can make sense to also accept this at other points, e.g. `PetabImporter.create_objective` as you have done, so that users can override/fine-tune/perform complicated experiments easier. But otherwise default to a `self.inner_solver_options` that is created in `PetabImporter.__init__`.

Anyway, up to you, this is a minor point, you've already designed it so users can specify things easily, nice job :+1:

README.md
```diff
@@ -25,7 +25,7 @@ pyPESTO features include:
 * Parameter estimation pipeline for systems biology problems specified in
   [SBML](http://sbml.org/) and [PEtab](https://github.com/PEtab-dev/PEtab)
   ([example](https://github.com/ICB-DCM/pyPESTO/blob/master/doc/example/petab_import.ipynb))
-* Parameter estimation with qualitative data as described in
+* Parameter estimation with ordinal data as described in
   [Schmiester et al. (2020)](https://doi.org/10.1007/s00285-020-01522-w).
   This is currently implemented in the `feature_ordinal` branch.
```
dweindl, line 30:
```suggestion
```

Branch no longer exists.

README.md
```diff
@@ -25,7 +25,7 @@ pyPESTO features include:
 * Parameter estimation pipeline for systems biology problems specified in
   [SBML](http://sbml.org/) and [PEtab](https://github.com/PEtab-dev/PEtab)
   ([example](https://github.com/ICB-DCM/pyPESTO/blob/master/doc/example/petab_import.ipynb))
-* Parameter estimation with qualitative data as described in
+* Parameter estimation with ordinal data as described in
```
dweindl, line 28: I think consensus was to use "qualitative".

pypesto/hierarchical/optimal_scaling/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+===============
+
+Contains the implementation of the optimal scaling approach, applied for integration
+of ordinal data in ODE modeling. Each ordinal datapoint is defined as being part of
+a category, where the mutual ordering of the categories of the same group is known.
+The category interval bounds are numerically optimized and quantitative surrogate
+measurements are calculated to represent the ordinal measurements. This constitutes
+the inner subproblem of the hierarchical optimization problem.
+
+Details on the optimal scaling approach can be found in Shepard, 1962, https://doi.org/10.1007/BF02289621
```
dweindl, line 12:
```suggestion
Details on the optimal scaling approach can be found in Shepard, 1962, https://doi.org/10.1007/BF02289621.
```

pypesto/hierarchical/optimal_scaling/calculator.py
```diff
@@ -0,0 +1,186 @@
...
+    import amici
+    from amici.parameter_mapping import ParameterMapping
+except ImportError:
+    pass
+
+
+class OptimalScalingAmiciCalculator:
+    """A calculator is passed as `calculator` to the pypesto.AmiciObjective.
+
+    The object is called by the pypesto.AmiciObjective in ``call_unprocessed``
```
dweindl, line 26:
```suggestion
    The object is called by :func:`pypesto.AmiciObjective.call_unprocessed`
```

yields proper links on readthedocs.

pypesto/hierarchical/optimal_scaling/solver.py
```diff
@@ -0,0 +1,832 @@
...
+            Ids of outer simulation parameters, includes fixed parameters.
+        snllh:
+            Empty dictionary with optimization parameters as keys.
+
+        Returns
+        -------
+        Filled in snllh dictionary with objective function gradients.
+        """
+        condition_map_sim_var = parameter_mapping[0].map_sim_var
+        # TODO: Doesn't work with condition specific parameters
```
dweindl, line 150: That should `raise` then.

README.md
Doresic: I should add this too, right? 
([example](https://github.com/ICB-DCM/pyPESTO/blob/master/doc/example/example_ordinal.ipynb))
Possibly commented out until it's in the main (master) branch?

README.md
Doresic: I don't remember anymore. 
In any case, I would still vote for ordinal. I think it's a more exact and descriptive name. Additionally, that's what it's called on Wikipedia :D https://en.wikipedia.org/wiki/Ordinal_data

README.md
dweindl: Yeah, adding the example is good. Fine with me to have a dead link on `develop` until the next release.

README.md
yannikschaelte: Ordinal and qualitative are two different things - ordinal requires a relative ordering (x, y) -> x<y. What Leonard assumed was at least such an ordering, so I'd say ordinal is more accurate?

pypesto/hierarchical/optimal_scaling/solver.py
Doresic: condition-specific parameters should be supported with the new commit. Will close the issue.

doc/example/example_ordinal.ipynb
Doresic: Ohh that's what you mean. Yes, good idea. Will add.

I added it now as `self._inner_solver_options` to the importer init. 
I also removed inner_problem_method, because it was given in the solver options, so now, only solver options need to be passed.

pypesto/hierarchical/optimal_scaling_approach/optimal_scaling_solver.py
Doresic: Fixed it in the new commit. Will close issue

README.md
Doresic: Ok, will add it too.

pypesto/petab/importer.py
```diff
@@ -415,6 +437,36 @@ def create_objective(
...
+                )
+            kwargs['guess_steadystate'] = False
+
+        if self._ordinal:
+            inner_solver_options = kwargs.pop(
+                'inner_solver_options', self._inner_solver_options
+            )
+            inner_problem_method = inner_solver_options.get(
+                'method', self._inner_solver_options.get('method')
+            )
```
dilpath, line 452: Not sure if this is ideal, since a user might find it unintuitive that some of their options provided via `PetabImporter.__init__(..., inner_solver_options)` get used as default here, but not others. I would suggest either using all as default, or none, with a preference for none -- better if the user explicitly specifies every time they customize via `create_objective` -- otherwise, there are two sources of defaults: `PetabImporter._inner_solver_options` and `OptimalScalingInnerSolver.get_default_options`.

Doresic: Yes, that's true. But then again, I don't see another clean way of being able to specify the options through `PetabImporter.__init__(..., inner_solver_options)`.
Possibly it's just best to revert to without that possibility.

test/hierarchical/test_ordinal.py
Doresic: Added some new tests in the last commit.
Some tests are failing, not sure why, will see tomorrow.

pypesto/petab/importer.py
dilpath: Hm, just to elaborate, what I meant is, if you opt for no defaults, then, inside `create_objective`:
```python
inner_solver_options = (
    inner_solver_options
    if inner_solver_options is not None
    else self._inner_solver_options
)
```

---

If you opt for defaults, then, inside `create_objective`:
```python
inner_solver_options = {
    **self._inner_solver_options,
    **inner_solver_options,
}
```

---

At the moment, you have the equivalent of, inside `create_objective`:
```
inner_solver_options = {
    **{
        k:v
        for k, v in self._inner_solver_options.items()
        if k == "method"
    },
    **inner_solver_options,
}
```

Doresic: Thanks for the clarification. Yes, I had a weird thing in `create_objective` that if you provided `inner_solver_options=None` then it would take `inner_solver_options = None`, but the method would be taken from `self._inner_solver_options`.

What I was trying to do is the first option you wrote, not having in mind the possibility that the user can provide`inner_solver_options=None` in `create_objective`. 

But now that I see your second suggestion, that seems like the best option to use, to use what options are provided in `create_objective`, and add options provided in `PetabImporter.__init__(..., inner_solver_options)`. And then again do the same with the default inner_solver_options.

Do you think this is unnecessary? And confusing (having a 2-level-default options, user provided and internal from inner_solver)

dilpath: Hm, my preference is for the first case, but I'm fine with either. You could emit a warning/exception if `self._inner_solver_options` is not empty when the user supplies `inner_solver_options` to `create_objective`. In this way, these can be treated as separate, incompatible use cases.

As for a user-supplied `None` -- it should be made clear in the docs that a user supplies `{}` to indicate an empty `inner_solver_options`, not `None`. Then, you can treat `None` as the lack of any user input for this argument.

Doresic: Yes, on second thought, you're right. 
As a user, I would hate the interaction of pulling the additional settings from `PetabImporter.__init__(..., inner_solver_options)` even if I give something in objective. 
Cool, thank you for the discussion. 

Will make it nice :)

README.md
Doresic: Daniel asked me to consider Jan as well, he confirmed ordinal would be more accurate.

Doresic:
Added a general visualization method of the optimal categories from either pypesto_result or even inner_result.
A use example can be seen at the end of `example_ordinal.ipynb`.

The initial idea was to implement it as an extension of `pypesto.visualize.model_fit.visualize_optimized_model_fit()`, but it turned out much simpler to implement it separately for now and place it in `pypesto.hierarchical.optimal_scaling.visualize`. 
So, the method is not in the visualization module. Should it be moved there?

LeonardSchmiester (changes requested):
Nice work. Thanks a lot @Doresic

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,310 @@
...
+                4.59542030e00,
+                6.23751538e00,
+                6.82578436e00,
+                8.40233862e00,
+                8.99060060e00,
+                1.10767572e01,
+            ]
+        ),
+        REDUCED: np.asarray(
+            [1.99999999, 4.00689557, 6.23751538, 8.40233862, 11.07675724]
```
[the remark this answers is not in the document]
Doresic: Yeah true. They should be the same. The true value is 11.07675723985916

What happened is that when I printed one and the other, due to the different "notation" of printing, for one case it printed one decimal more. 
In any case, doesn't matter as much since I selected `rtol=1e-3`. But will change for correctness.

pypesto/hierarchical/optimal_scaling/problem.py
```diff
@@ -0,0 +1,538 @@
...
+    estimate_ub = True
+    estimate_lb = False
+
+    if method == STANDARD:
+        estimate_lb = True
+
+    return estimate_lb, estimate_ub
+
+
+def qualitatiave_ixs_for_measurement_specific_parameters(
```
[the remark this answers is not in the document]
Doresic: Agree. Though, the other ones I renamed to `optimal_scaling`, as it's related to the `OptimalScalingProblem`, and not so directly to the ordinal measurements.

pypesto/hierarchical/optimal_scaling/solver.py
```diff
@@ -0,0 +1,879 @@
...
+
+
+class OptimalScalingInnerSolver(InnerSolver):
+    """Solve the inner subproblem of the optimal scaling approach for ordinal data.
+
+    Options
+    -------
+    method:
+        The method to use for the inner optimization problem.
+        Can be 'standard' or 'reduced'.
```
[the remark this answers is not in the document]
Doresic: True. This is stressed in the example notebook already, but it should pop as a warning or be at least documented.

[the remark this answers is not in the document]
Doresic: Will add this:
`
        elif self.options['method'] == STANDARD:
            warnings.warn(
                'Standard approach is not recommended, as it is less efficient.'
                'Please consider using the reduced approach instead.'
            )
`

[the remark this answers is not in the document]
Doresic: And this:
method:
        The method to use for the inner optimization problem.
        Can be 'standard' or 'reduced'. The latter is more efficient.

pypesto/hierarchical/optimal_scaling/solver.py
```diff
@@ -0,0 +1,873 @@
...
+    )
+
+    dxi_dtheta = linalg.spsolve(A_sp, b)
+    return dxi_dtheta[: problem.groups[group]['num_inner_params']]
+
+
+def get_dy_dtheta(gr: int, problem: OptimalScalingProblem, sy_all: np.ndarray):
+    """Restructure sensitivities into a numpy matrix of right dimension."""
+    return np.block(
+        [sy_all, np.zeros(2 * problem.groups[gr]['num_categories'])]
```
LeonardSchmiester, line 348:
```suggestion
def get_dy_dtheta(group: int, problem: OptimalScalingProblem, sy_all: np.ndarray):
    """Restructure sensitivities into a numpy matrix of right dimension."""
    return np.block(
        [sy_all, np.zeros(2 * problem.groups[group]['num_categories'])]
```

LeonardSchmiester (approved):

doc/example/example_ordinal.ipynb
```diff
@@ -0,0 +1,421 @@
+{
```
yannikschaelte, line 1: Could you run the notebook on your computer once, and keep the outputs? So they will show in the documentation.

doc/example/example_ordinal.ipynb
```diff
@@ -0,0 +1,421 @@
...
+ "cells": [
```
yannikschaelte, line 2: and please add your notebook to the list here https://github.com/ICB-DCM/pyPESTO/blob/main/test/run_notebook.sh#L17 (does it run in less than say a minute?) so that it is run regularly with the tests.

yannikschaelte: (if it has special requirements, make sure those are installed)

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+@pytest.fixture(params=inner_solver_options)
+def inner_solver_options(request):
+    return request.param
+
+
+def test_evaluate_objective(inner_solver_options: List[Dict]):
+    """Check that standard / reduced / reparameterized formulations yield the
+    same result."""
```
yannikschaelte, line 59: :+1:

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,310 @@
...
+    vals = []
+    for idx, option in enumerate(inner_solver_options):
+        problem = _create_problem(petab_problem, option)
+        val = problem.objective(np.array([0, 0]))
+        vals.append(val)
+        assert np.isclose(vals[idx], vals[idx - 1])
+
+
+def test_optimization(inner_solver_options: List[Dict]):
+    """Check that optimizations finishes without error."""
```
yannikschaelte, line 70: you could add a check whether 
* the obtained values (fval, x, grad, ...) are not nan or inf (`np.isfinite, np.allfinite`)
* whether the final fval is < fval0

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,310 @@
...
+        method: calculate(problems[method], x_dct=x_dct) for method in methods
+    }
+
+    # The results of the objective gradient and function value
+    # should not depend on the method given.
+    assert np.isclose(
+        calculator_results[STANDARD]['fval'],
+        calculator_results[REDUCED]['fval'],
+    )
+    assert np.isclose(
```
yannikschaelte, line 143: could you also compare gradients to finite differences? --> pypesto.objective.FD

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,310 @@
...
+    # Construct inner problem
+    inner_problem = OptimalScalingProblem(
+        xs=inner_parameters, data=[data], method=STANDARD
+    )
+
+    return inner_problem, expected_inner_parameter_values, simulation
+
+
+def test_optimal_scaling_solver():
+    """Test the Optimal scaling solver."""
```
yannikschaelte, line 222: what exactly does it test?

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,310 @@
...
+
+    assert np.all(
+        np.isclose(reduced_result['x'], expected_values[REDUCED], rtol=rtol)
+    )
+    assert np.all(np.isclose(reduced_result['fun'], 0, rtol=rtol))
+    assert np.all(np.isclose(reduced_result['jac'], 0, rtol=rtol))
+
+
+def test_surrogate_data_analytical_calculation():
+    """Test analytical calculation of surrogate data.
```
yannikschaelte, line 263: :+1:

test/hierarchical/test_ordinal.py
```diff
@@ -0,0 +1,310 @@
...
+        interval_gap,
+        options,
+    )
+
+    assert np.isclose(
+        interval_range, expected_values['interval_range'], rtol=rtol
+    )
+    assert np.isclose(interval_gap, expected_values['interval_gap'], rtol=rtol)
+    assert np.all(
+        np.isclose(
```
yannikschaelte, line 307: np.allclose

yannikschaelte (approved):
Some final comments, but looks very nice!

test/hierarchical/test_ordinal.py
yannikschaelte: would be even better with formulas that reproduce how the analytical values were calculated, but ok

Doresic: `get_surrogate_all` and `compute_interval_constraints` are the functions of analytical formulas for everything that's being computed analytically in the inner_solver. So those are the ones I'm testing. 
Do you mean to get the `expected_values` with new, rewritten formulas in `test_ordinal.py`?

yannikschaelte: yes, explicit formulas for expected values instead of "magic" numbers would improve credibility, but only if easily possible. otherwise we trust that you calculated these correctly :D

Doresic: Yes, it's easy to do. Fixed in next commit.

test/hierarchical/test_ordinal.py
Doresic: It tests the inner solver directly using a very simple synthetic example from `_inner_problem_exp()`. 
Since the problem is so simple, it should solve it with fval=0 and grad=0 and get the `expected_inner_parameter_values`.

doc/example/example_ordinal.ipynb
```diff
@@ -0,0 +1,504 @@
...
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "from pypesto.visualize import waterfall\n",
+    "\n",
+    "waterfall(\n",
+    "    [res_standard, res_reduced, res_reduced_reparameterized], order_by_id=True\n",
+    ")"
```
FFroehlich, line 382:
```suggestion
    ");"
```

doc/example/example_ordinal.ipynb
```diff
@@ -0,0 +1,504 @@
...
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this plot we can see that all surrogate data is inside their respective categories. So the ordering given in the measurements is fully satisfied. This constitutes a perfect fit of the data in the ordinal sense.  "
+   ]
+  },
+  {
```
FFroehlich, line 470: remove empty cell

Doresic (approved):

pypesto/petab/importer.py
```diff
@@ -415,6 +436,37 @@ def create_objective(
             par_ids = [x for x in par_ids if x not in inner_parameter_ids]
 
             # FIXME: currently not supported with hierarchical
+            if 'guess_steadystate' in kwargs and kwargs['guess_steadystate']:
+                warnings.warn(
+                    "`guess_steadystate` not supported with hierarchical optimization. Disabling `guess_steadystate`."
+                )
+            kwargs['guess_steadystate'] = False
+
+        if self._ordinal:
```
FFroehlich, line 445: shouldn't be there be separate handling of `self._ordinal={True}`x`self._hierarchical={True,False}` conditions?

pypesto/petab/importer.py
```diff
@@ -415,6 +436,37 @@ def create_objective(
...
+            )
+            inner_solver = OptimalScalingInnerSolver(
+                options=inner_solver_options
+            )
+            calculator = OptimalScalingAmiciCalculator(
+                inner_problem, inner_solver
+            )
+            amici_reporting = amici.RDataReporting.full
+
+            # FIXME: currently not supported with hierarchical
```
FFroehlich, line 465: this can be reached even if hierarchical is false.

pypesto/petab/importer.py
```diff
@@ -415,6 +436,37 @@ def create_objective(
...
+                inner_solver_options
+                if inner_solver_options is not None
+                else self._inner_solver_options
+            )
+            inner_problem_method = inner_solver_options.get('method', None)
+
+            inner_problem = OptimalScalingProblem.from_petab_amici(
+                self.petab_problem, model, edatas, inner_problem_method
+            )
+            inner_solver = OptimalScalingInnerSolver(
```
FFroehlich, line 457: I don't understand why construction of inner problem/solver is handled differently from the standard hierarchical setting. Wouldn't it make more sense to implement inner solver option handling more generically?

FFroehlich (changes requested):
Had a few minor comments, but only looked at the notebook and `pypesto/petab/importer.py`.

pypesto/petab/importer.py
Doresic: As it is currently implemented, ordinal measurements can be dealt with only hierarchically, with optimal scaling parameters (category parameters) being optimized in the inner loop. 

It would be possible to implement the possibility of optimizing them non-hierarchically, but it would take time, and as it would probably be much slower because sometimes there can be as many optimal scaling parameters as many measurements there are, I would say it's probably not worth it.

But I agree that the checks here are ambiguous (or naming). Because it would then make sense to have `self._hierarchical=True` obligatory for `self._ordinal=True`. But since the optimal scaling method of dealing with ordinal data is so different from the hierarchical manner of dealing with scaling, offset, and noise parameters, it made the most sense to separate it completely and with that to have different checks. 

I guess what would make the most sense would be to have:
`if self._hierarchical and not self._ordinal` -> Go into the hierarchical approach of optimizing scaling, offset, and noise parameters
`if self._hierarchical and self._ordinal` -> Go into Optimal Scaling for dealing with ordinal data.

But then again, don't know if this is just overcomplicating over a simple issue, and will be even worse with the later addition of `self._nonlinear_monotone`.

Doresic: Or it's better even:
`
if self._hierarchical and not self._ordinal
   ... optimal scaling & ordinal
elif self._hierarchical
` ... scaling offset & noise

PaulJonasJost (approved):

pypesto/petab/importer.py
```diff
@@ -354,6 +373,8 @@ def create_objective(
             Whether to force-compile the model if not passed.
         **kwargs:
             Additional arguments passed on to the objective.
+            In case of ordinal measurements, inner_solver_options can optionally be passed here,
+            otherwise, those given to the importer constructor (or inner solver defaults) will be chosen.
```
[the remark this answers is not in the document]
Doresic: It refers to "no inner_solver_options provided". I agree, it's ambiguous.
Will remove "In case of ordinal measurements,", it's not needed.

pypesto/petab/importer.py
```diff
@@ -415,6 +436,37 @@ def create_objective(
...
+            inner_problem = OptimalScalingProblem.from_petab_amici(
+                self.petab_problem, model, edatas, inner_problem_method
+            )
+            inner_solver = OptimalScalingInnerSolver(
+                options=inner_solver_options
+            )
+            calculator = OptimalScalingAmiciCalculator(
+                inner_problem, inner_solver
+            )
+            amici_reporting = amici.RDataReporting.full
```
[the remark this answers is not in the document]
Doresic: Good point.
So we do need from amici ReturnData the simulations and sensitivities, so I need to set the reporting to something more than just likelihood, but asking Daniel now, it turns out I don't need `...full` and that I can just use the `...residuals` one as it has all of that too and is faster.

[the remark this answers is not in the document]
Doresic: It seems like we can do this with the hierarchical one as well as it needs only simulations and sigma's from ReturnData. @yannikschaelte should I change it from `amici.RDataReporting.full` to `amici.RDataReporting.residuals` there too?

pypesto/petab/importer.py
Doresic: `OptimalScalingProblem` and `OptimalScalingParameter` are implemented as child classes of the generic `InnerProblem` and `InnerParameter` that are used for relative data (scaling, offset, noise). 

On the other hand, `OptimalScalingSolver` and `OptimalScalingCalculator` are not since they're completely different from the calculator and solvers used for relative data. 

So because some of the classes are completely separate, I implemented the construction of everything completely separate too. I agree it's not nice and would like it to be nicer. Especially it won't be nice when I PR the spline method for nonlinear monotone data too. 

So I'm open to suggestions now, and then, on how to make it nicer.

pypesto/petab/importer.py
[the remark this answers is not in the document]
dweindl: `RDataReporting.residuals` will not give us the gradient. Might be a good idea to think about some extra mode as suggested by @FFroehlich, but that's out of scope here.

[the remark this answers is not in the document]
FFroehlich: gradient is accessible as `res.dot(sres)` though.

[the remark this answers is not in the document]
Doresic: Usually yes, for ordinal measurements it's not `res.dot(sres)`. There's surrogate data instead of measurements, which depends on the outer parameters. So optimal scaling still needs `sy`.

pypesto/petab/importer.py
FFroehlich: okay should then be fine to refactor later on.

pypesto/petab/importer.py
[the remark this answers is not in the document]
FFroehlich: I think "In case of ordinal measurements" is still important as those solver options do not apply in the case of standard hierarchical estimation.

pypesto/petab/importer.py
FFroehlich: I think if there is no intention to implement a non-hierarchical ordinal approach it's fine to set `self._hierarchical=True` in case `ordinal==True`, but this should be noted in the docstrings for both `ordinal` and `hierarchical`.

pypesto/petab/importer.py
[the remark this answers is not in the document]
Doresic: Yeah true. I'll make it clear that's the case:

In case of ordinal measurements, Inner_solver_options can optionally be passed here.
If none are given, inner_solver_options given to the importer constructor
(or inner solver defaults) will be chosen.

FFroehlich (approved):

plakrisenko (approved):

stephanmg (approved):
Looks good to me.

PR #972, opened 2022-11-15
ESS: Fix local_n1, local_n2 parameters

dweindl:
The *ESS publications are ambiguous on what this parameter controls. The text refers to ESS iterations, the pseudocode refers to function evaluations. The former seems to be more convenient. With these changes, local_n{1,2} will be applied to iterations instead of function evaluations.

### DIFF
--- a/pypesto/optimize/ess/ess.py
+++ b/pypesto/optimize/ess/ess.py
@@ -88,9 +88,9 @@ class ESSOptimizer:
         max_iter:
             Maximum number of ESS iterations.
         local_n1:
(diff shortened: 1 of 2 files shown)


dilpath (approved):
Code quality looks fine so fine to merge.

Is it clear from an implementation what these parameters are?

dweindl:
> Is it clear from an implementation what these parameters are?

With these changes, it should be that same as e.g. in https://bitbucket.org/jrbanga_/meigo64/src/master/MEIGO/eSS/ess_kernel.m

PR #976, opened 2022-11-16
Plot measurements in `sampling_prediction_trajectories`

dilpath:
Example at the end of the notebook

https://github.com/ICB-DCM/pyPESTO/blob/0cd9c8de8414ac72a4624178d26a5f682c9c7291/doc/example/sampling_diagnostics.ipynb


elbaraim:
Thanks @dilpath !
I would suggest some minor changes:
- Add in the legend the label for the dot plot, e.g. measurement data / data / experimental data
- Default color for the dot plot should be changed when used with default color options for the predictions, i.e. cannot see the black dots on top of the dark purple. Switch dot plot to a lighter color.

dilpath:
Thanks for the review!

> * Add in the legend the label for the dot plot, e.g. measurement data / data / experimental data

:+1:

> * Default color for the dot plot should be changed when used with default color options for the predictions, i.e. cannot see the black dots on top of the dark purple. Switch dot plot to a lighter color.

The dots are not black, they are filled with the color of the observable. I decided to keep the color the same, to ensure consistency with colors in the legend. Instead, the data points now have a white outline.

elbaraim:
> The dots are not black, they are filled with the color of the observable. I decided to keep the color the same, to ensure consistency with colors in the legend. Instead, the data points now have a white outline.

I did not noticed as the color is quite dark. Good idea! It actually looks much better now. However for lighter colors (e.g. green color used in the example above) I am not sure if the white outline will be helpful there. Maybe you can try it out and see how it looks.

dilpath:
> However for lighter colors (e.g. green color used in the example above) I am not sure if the white outline will be helpful there. Maybe you can try it out and see how it looks.

Agreed, now switches between black or white outlines depending on "lightness" of color.
image

elbaraim:
Looks great!

elbaraim (approved):
👍🏽

PR #979, opened 2022-12-06
Allow scipy optimizer to use fun with integrated grad 

### LINKED ISSUE #871: SciPy optimizer does redundant function evaluations e.g. for AMICI objectives
## Problem/Bug description

The scipy optimizer calls `fun` and `grad` via `object` via `objective.get_fun` and `objective.get_grad`, see

https://github.com/ICB-DCM/pyPESTO/blob/4c30abfca56ba714c302141cd44a9dd366bff4bb/pypesto/optimize/optimizer.py#L420

and for the function/gradient definition see (and line below)
https://github.com/ICB-DCM/pyPESTO/blob/4c30abfca56ba714c302141cd44a9dd366bff4bb/pypesto/optimize/optimizer.py#L399

The `get_grad` function is defined here (not. overwritten in the `AmiciObjective`...)
https://github.com/ICB-DCM/pyPESTO/blob/4c30abfca56ba714c302141cd44a9dd366bff4bb/pypesto/objective/base.py#L351

and therefore the ODE is evaluated twice, once for `fun`, once implicitly for `grad`.

## Solution

If one defines `fun` to return a tuple `(f_val, grad_val)`, and sets `jac=True`, scipy is able to deal with objectives, that return both simultaneously, see

https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html

MerktSimon:
Scipy optimizer redundantly evaluates objective function twice if you pass over function and gradients to objective in one function x -> (fun, grad)(x).
(fixes #871)


MerktSimon:
not sure what to make of 
` E: Package 'python-dev' has no installation candidate`
in optimize test

dweindl:
> not sure what to make of ` E: Package 'python-dev' has no installation candidate` in optimize test

Fixed in https://github.com/ICB-DCM/pyPESTO/pull/979/commits/103b357d60cb7d7e073a3854c4f3c77033d28c0e

test/optimize/test_optimize.py
```diff
@@ -522,3 +522,28 @@ def test_ess(problem, local_optimizer, ess_type, request):
...
+        problem=problem,
+        optimizer=optimizer,
+        n_starts=1,
+        startpoint_method=pypesto.startpoint.uniform,
+        options=optimize_options,
+        history_options=history_options,
+        progress_bar=False,
+    )
+    assert (
+        len(result.optimize_result.history[0]._trace['fval'])
```
dilpath, line 547: This `_trace` looks odd to me. Should possibly be replaced with e.g. `result.optimize_result.....get_fval_trace()`. But fine to keep this as is.

https://github.com/ICB-DCM/pyPESTO/blob/86f4feaa8024c14bd0037cfc1522d9d5bb3fa3b6/pypesto/history/memory.py#L76-L83

dilpath (approved):
:+1: Thanks for fixing this!

MerktSimon:
seems like some tests fail to reasons unrelated to this merge request, e.g. the select ones. How is this handled in general?

MerktSimon:
Also https://github.com/ICB-DCM/pyPESTO/pull/976#event-7988419424 merged without any failed tests, but merging new develop into this branch and rerunning them does not help.


In particular 
image
image
seems to be fixed in  https://github.com/ICB-DCM/pyPESTO/blob/e4c2f31e96efbafc85896c2c149d69b2a7906712/tox.ini#L53

dweindl:
> Also [#976 (comment)](https://github.com/ICB-DCM/pyPESTO/pull/976#event-7988419424) merged without any failed tests, but merging new develop into this branch and rerunning them does not help.

Resolved in #981

pypesto/optimize/optimizer.py
```diff
@@ -382,12 +383,37 @@ def minimize(
...
+            #  in objective and pass to scipy seperately
+            if objective.hess is True:
+                logger.warning(
+                    "scipy.optimize.minimize does not support "
+                    "passing fun and hess as one function. Hence "
+                    "for each evaluation of hess, fun will be "
+                    "evaluated again. This can lead to increased "
+                    "computation times. If possible, separate fun "
+                    "and hess."
+                )
```
dweindl, line 403: @MerktSimon This warning is printed several thousand times during an optimization (see e.g. https://github.com/ICB-DCM/pyPESTO/actions/runs/3741447433/jobs/6351066417). I think it would be much preferable to use `warnings.warn`, where this will only be printed once, by default.

MerktSimon: I agree. Since this PR is already merged now, should I just add this to #988 ?

dweindl: Better create a separate PR.

PR #996, opened 2022-12-21
Warn only once if using ineffiecient objective settings

MerktSimon:
Substitute `logging.warning` by `wanrings.warn` s.t. warnings are only raised once (c.f. discussion in #979)


dweindl (approved):
:+1:

PR #1002, opened 2023-01-16
Dynesty sampler: get MCMC-like samples

dilpath:
Adds methods to save/restore the internal `dynesty` sampler itself, and use the internal `dynesty` sampler to generate MCMC-like samples or get the original samples.

By default, the sampler will return MCMC-like samples.


pypesto/sample/dynesty.py
```diff
@@ -40,6 +40,13 @@
 class DynestySampler(Sampler):
     """Use dynesty for sampling.
 
+    NB: `get_samples` returns MCMC-like samples, by resampling original
```
yannikschaelte, line 43: also point out what the difference between both result types is?

yannikschaelte (approved):
looks good! maybe a test for both cases?

PR #1006, opened 2023-01-20
Hierarchical: decouple PEtab validators

dilpath:


PaulJonasJost (approved):
Does the validation for a petab problem work on a hierarchical problem? If not (which I assume due to hier. having its own validation) how is the distinction handled? Cannot see this here.

FFroehlich (approved):
fine with me, would prefer separation of semantic vs support checks in the long run though

dilpath:
> Does the validation for a petab problem work on a hierarchical problem? If not (which I assume due to hier. having its own validation) how is the distinction handled? Cannot see this here.

Yes, hierarchical problems should also pass the standard PEtab validator. The additional hierarchical validation methods are to ensure that the hierarchical parts are specified correctly and are compatible with pyPESTO.

Issue #723, opened 2021-09-01
Consolidate different gradient checks


jvanhoefer:
**Feature description**
Currently we have several different (rather redundant) gradient checking routines, which only have slightly different functionalities, which in my opinion do not all justify being there in parallel. (In the end: We want to have a minimal interface to the outside world to be user-friendly):

In `objective/base`
https://github.com/ICB-DCM/pyPESTO/blob/0aaa34179497c429434b4b6d2aeb35309b113961/pypesto/objective/base.py#L370

https://github.com/ICB-DCM/pyPESTO/blob/0aaa34179497c429434b4b6d2aeb35309b113961/pypesto/objective/base.py#L422

https://github.com/ICB-DCM/pyPESTO/blob/0aaa34179497c429434b4b6d2aeb35309b113961/pypesto/objective/base.py#L589

In the PEtab importer
https://github.com/ICB-DCM/pyPESTO/blob/fb2be819b41411dc1686d8429ce5efa6c535b70b/pypesto/petab/importer.py#L83

**Motivation/Application**
I think, that having _one_ FD check in `objective/base` and maybe another one for checking multiple `eps` values is great. The one in the `PEtab` importer is clearly redundant, since you would rather create a `problem` object, and then call a gradient check on that (which is also, what the function does...). The third function in the objective base also seems close to the other ones and not "general purpose". 

Therefore I would like to 

a) propose to get rid of the gradient check in the PEtab importer and would like to 

b) open the discussion, if the third gradient check in `objective/base` adds enough "general purpose" value, that it justifies to have and maintain an other gradient checking method there. 

Opinions? @yannikschaelte @dweindl @stephanmg

stephanmg:
Dear @jvanhoefer,

as we have discussed, the method `check_gradients_match_finite_differences` will be used only internally, so on my fork I made changes accordingly by prefixing with `_`. I think this doesn't have to be part of pyPESTO's public API.

Instead, I've added a visualization method which uses `_check_gradients_match_finite_differences` to visualize gradients and might be more useful to the general public. I will request a PR / review once I am fully satisfied with the implementation of my gradient consistency visualization plotting.

yannikschaelte:
[...]

PR #1012, opened 2023-02-01
Consolidate grad checks

stephanmg:
Finally addressing issue #723


pypesto/objective/amici/amici.py
```diff
@@ -598,7 +598,7 @@ def set_custom_timepoints(
         amici_objective.apply_custom_timepoints()
         return amici_objective
 
-    def check_gradients_match_finite_differences(
+    def __check_gradients_match_finite_differences(
```
yannikschaelte, line 601:
```suggestion
    def _check_gradients_match_finite_differences(
```
Why not just a single underscore, marking this as a private function, instead of name mangling?

dweindl: Maybe I am missing something, but shouldn't gradient check be public?

pypesto/objective/amici/amici.py
```diff
@@ -615,6 +615,8 @@ def check_gradients_match_finite_differences(
         if x is None and 'petab_problem' in dir(self.amici_object_builder):
             x = self.amici_object_builder.petab_problem.x_nominal_scaled
             x_free = self.amici_object_builder.petab_problem.x_free_indices
-        return super().check_gradients_match_finite_differences(
-            x=x, x_free=x_free, *args, **kwargs
+        return (
```
FFroehlich, line 618: where is  this `_ObjectiveBase` here coming from?

dweindl: [Name mangling (`__name` attributes)](https://docs.python.org/3/tutorial/classes.html#private-variables)

pypesto/petab/importer.py
```diff
@@ -127,76 +127,6 @@ def from_yaml(
             model_name=model_name,
         )
 
-    def check_gradients(
```
FFroehlich, line 130: why is this function completely removed instead of replacing most of it's implementation with a call to `check_gradients_match_finite_differences`

stephanmg: See the discussion here: https://github.com/ICB-DCM/pyPESTO/issues/723

Back then I naively introduce the check_gradient(...) method, but in fact this method is superfluous, see the issue #723.

pypesto/objective/amici/amici.py
stephanmg: We have check_grad and check_grad_multi_eps already as per #723.

The method `check_garadients_match_finite_differences` I am using in a visualization routine in pypesto/visualize. (Will be added in a subsequent pull request, that's why method marked private as only used internally)

stephanmg:
Somehow the PR doesn't get updated...

stephanmg:
Could someone have a look again? Would like to merge this finally.

FFroehlich:
> Somehow the PR doesn't get updated with my latest commits...

is this still an issue? Also looks like there are some merge conflicts that still need to be resolved.

stephanmg:
Will have a look into this to be able to merge this PR soon.

stephanmg:
Closing this issue in favor of a from scratch PR.

PR #1014, opened 2023-02-08
added option to profile the whole parameter bounds.

PaulJonasJost:
Talked about it in subgroup meeting. Do you think we need any additional test for this?

### DIFF
--- a/pypesto/profile/walk_along_profile.py
+++ b/pypesto/profile/walk_along_profile.py
@@ -63,14 +63,20 @@ def walk_along_profile(
 
         # check if the next profile point needs to be computed
         if par_direction == -1:
-            stop_profile = (x_now[i_par] <= problem.lb_full[[i_par]]) or (
-                current_profile.ratio_path[-1] < options.ratio_min
-            )
+            if options.whole_path:
+                stop_profile = x_now[i_par] <= problem.lb_full[[i_par]]
+            else:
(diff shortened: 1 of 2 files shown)


pypesto/profile/walk_along_profile.py
```diff
@@ -63,14 +63,20 @@ def walk_along_profile(
...
         if par_direction == 1:
-            stop_profile = (x_now[i_par] >= problem.ub_full[[i_par]]) or (
-                current_profile.ratio_path[-1] < options.ratio_min
-            )
+            if options.whole_path:
+                stop_profile = x_now[i_par] >= problem.ub_full[[i_par]]
+            else:
+                stop_profile = (x_now[i_par] >= problem.ub_full[[i_par]]) or (
+                    current_profile.ratio_path[-1] < options.ratio_min
+                )
```
dweindl, line 79: Could combine those two `if par_direction` by multiplying on side of the `>=` with `par_direction`

dweindl (approved):
I am not familiar with that code, but could the same also be achieved `ratio_min=-inf` or is that also used in other places?

Otherwise good.

PaulJonasJost:
> I am not familiar with that code, but could the same also be achieved `ratio_min=-inf` or is that also used in other places?
> 
> Otherwise good.

might be possible, I shall see.

PR #1016, opened 2023-02-09
Doc: Install amici on RTD

dweindl:
To run example notebooks (-> https://readthedocs.org/projects/pypesto/builds/19417900/)


yannikschaelte (approved):
Thanks! Plus: This removes a deprecation (version) I had wanted to remove anyway.

PR #1018, opened 2023-02-10
PetabImporter.create_startpoint_method - kwargs, not-None

dweindl:
* Add kwarg forwarding to PetabImporter.create_startpoint_method (to use checked startpoints more conveniently)
* Always return a startpoint method


pypesto/petab/importer.py
```diff
@@ -568,24 +568,27 @@ def create_prior(self) -> Union[NegLogParameterPriors, None]:
...
         """Create a startpoint method.
 
-        If the PEtab problem specifies an initializationPrior. Returns None,
-        if no initializationPrior is specified.
+        If the PEtab problem specifies an ``initializationPrior``. Returns
+        ``None``, if no ``initializationPrior`` is specified.
+
+        Parameters
+        ----------
+        *+kwargs: Additional keyword arguments passed on to
```
dweindl, line 581:
```suggestion
        **kwargs: Additional keyword arguments passed on to
```

pypesto/petab/importer.py
```diff
@@ -568,24 +568,24 @@ def create_prior(self) -> Union[NegLogParameterPriors, None]:
         else:
             return None
 
-    def create_startpoint_method(self) -> Union[StartpointMethod, None]:
+    def create_startpoint_method(
+        self, **kwargs
+    ) -> Union[StartpointMethod, None]:
```
yannikschaelte, line 573:
```suggestion
    ) -> StartpointMethod:
```

yannikschaelte (approved):
PEtab then internally handles mapping from initialization or parameter priors. Looks reasonable to me.

dweindl:
An independent issue is, that changing the bounds on an PetabImporter-generated Problem, will have no effect on the startpoint sampling, as the bounds of the original PEtab problem will be used. Not sure if bug or feature.

yannikschaelte:
Certainly a feabug. Initialization prior precedes bounds when it comes to start point sampling. If this is clearly stated, then it should be fine?

FFroehlich:
> An independent issue is, that changing the bounds on an PetabImporter-generated Problem, will have no effect on the startpoint sampling, as the bounds of the original PEtab problem will be used. Not sure if bug or feature.

sounds like we need a PetabProblem class with some frozen member attributes? Sounds like a nice opportunity to shoot yourself in the foot otherwise.

FFroehlich (approved):

FFroehlich:
Also, why am I a codeowner here?

PR #1023, opened 2023-02-17
add getting started notebook

jvanhoefer:


doc/example/boehm_JProteomeRes2014/Boehm_JProteomeRes2014.yaml
```diff
@@ -0,0 +1,13 @@
+format_version: 1
+parameter_file: parameters_Boehm_JProteomeRes2014.tsv
+problems:
```
PaulJonasJost, line 3: in `petab_import.ipynb` it is shown how to use the benchmark model without having to add this as folder. would be nice here as well.

doc/example.rst
```diff
@@ -23,6 +23,7 @@ Getting started
 .. toctree::
    :maxdepth: 2
 
+   example/getting_started.ipynb
```
PaulJonasJost, line 26: Cannot make a comment within getting_started, therefore here:

I think Software Development Standard and Further topics are within the notebook twice

PaulJonasJost (approved):

doc/example.rst
jvanhoefer: At least I only see it in there one ^^

doc/example/boehm_JProteomeRes2014/Boehm_JProteomeRes2014.yaml
jvanhoefer: I am aware, but I would argue that:

a) this is not the default way of importing a PEtab problem (usually, we import via file dir and not via providing a `petab.Problem` instance). I.e., the resulting code does not show standard usage and is not copy-paste-able
b) This generates code that has nothing to do with core pypesto usage and hence is distracting from the central goal of the notebook.

jvanhoefer: On a different note: It did not suppress the Optimizer output, which makes stuff look horrible (basically hides sampling underneath a block that no one will scroll through...).

If you know why `%%capture` did not work and how to make it work, I am happy for fixes/suggestions.

PR #1027, opened 2023-03-01
Fix optimizers

PaulJonasJost:


test/optimize/test_optimize.py
```diff
@@ -566,3 +566,26 @@ def test_scipy_integrated_grad():
...
+    # run optimization
+    result = optimize.minimize(
+        problem=problem,
+        optimizer=opt,
+        n_starts=1,
+        progress_bar=False,
+        history_options=pypesto.HistoryOptions(trace_record=True),
+    )
+    # check that the startpoint was used
+    assert result.optimize_result.list[0].x0 == pytest.approx(
```
yannikschaelte, line 589: Good idea, but this test does not tell much yet ... rather define x0 explicitly in this function, and make sure that result.optimize_result used that value

PaulJonasJost: oh yes, I did in the test via the x_guesses :) will change to x_guesses

test/optimize/test_optimize.py
```diff
@@ -566,3 +566,31 @@ def test_scipy_integrated_grad():
...
     )
+
+
+def test_correct_startpoint_usage(optimizer):
+    """
+    Test that the startpoint is correctly used in all optimizers.
+    """
+    # initial start values is not supported for pyswarm, dlib and pyswarms
+    if optimizer in [('dlib', ''), ('pyswarm', ''), ('pyswarms', '')]:
+        assert True
```
yannikschaelte, line 577:
```suggestion
```
line is ineffective

test/optimize/test_optimize.py
```diff
@@ -566,3 +566,31 @@ def test_scipy_integrated_grad():
...
+        problem=problem,
+        optimizer=opt,
+        n_starts=1,
+        progress_bar=False,
+        history_options=pypesto.HistoryOptions(trace_record=True),
+    )
+    # check that the startpoint was used
+    assert problem.x_guesses[0] == pytest.approx(
+        result.optimize_result[0].history.get_x_trace(0)
+    )
```
yannikschaelte, line 596: test looks good to me

pypesto/optimize/optimizer.py
```diff
@@ -607,6 +603,13 @@ def get_default_options(self):
         """Create default options specific for the optimizer."""
         return {'maxiter': 10000}
 
+    def check_x0_support(self, problem: Problem):
+        """Check whether optimizer supports x0."""
+        if problem.x_guesses is not None:
+            logger.warn("The Dlib optimizer does not support x0.")
+        else:
+            logger.debug("The Dlib optimizer does not support x0.")
```
yannikschaelte, line 611:
```suggestion
```
no need imho

pypesto/optimize/optimizer.py
```diff
@@ -659,6 +662,13 @@ def is_least_squares(self):
         """Check whether optimizer is a least squares optimizer."""
         return False
 
+    def check_x0_support(self, problem: Problem):
+        """Check whether optimizer supports x0."""
+        if problem.x_guesses is not None:
+            logger.warn("The pyswarm optimizer does not support x0.")
+        else:
+            logger.debug("The pyswarm optimizer does not support x0.")
```
yannikschaelte, line 670:
```suggestion
```

pypesto/optimize/optimizer.py
```diff
@@ -795,7 +805,7 @@ def minimize(
         bounds = list(zip(problem.lb, problem.ub))
 
         result = scipy.optimize.differential_evolution(
-            problem.objective.get_fval, bounds, **self.options
+            problem.objective.get_fval, bounds, x0=x0, **self.options
```
yannikschaelte, line 808: :+1:

pypesto/optimize/optimizer.py
```diff
@@ -926,6 +937,13 @@ def is_least_squares(self):
         """Check whether optimizer is a least squares optimizer."""
         return False
 
+    def check_x0_support(self, problem: Problem):
+        """Check whether optimizer supports x0."""
+        if problem.x_guesses is not None:
+            logger.warn("The pyswarms optimizer does not support x0.")
+        else:
+            logger.debug("The pyswarms optimizer does not support x0.")
```
yannikschaelte, line 945:
```suggestion
```

pypesto/optimize/task.py
```diff
@@ -53,6 +53,8 @@ def __init__(
     def execute(self) -> OptimizerResult:
         """Execute the task."""
         logger.debug(f"Executing task {self.id}.")
+        # check for supplied x_guess support
+        self.optimizer.check_x0_support(self.problem)
```
yannikschaelte, line 57: could also directly pass self.problem.x_guesses; would be cleaner?

pypesto/visualize/sampling.py
```diff
@@ -1308,9 +1308,6 @@ def sampling_1d_marginals(
     # fig, ax = plt.subplots(nr_params, figsize=size)[1]
     for idx, par_id in enumerate(param_names):
         if plot_type == 'kde':
-            # bw_method argument is deprecated... FutureWarning:
```
yannikschaelte, line 1311: what happened here?

yannikschaelte (approved):

pypesto/visualize/sampling.py
PaulJonasJost: this is some old code, so I removed it, it is hinting towards the `bw` argument, which was used previously, and now is replaced by bw_method`

pypesto/optimize/task.py
PaulJonasJost: but x_guesses is not none if we initialised the problem but `np.zeros((0, self.dim_full))` which is an empty array. Could probably check for that 🤔

PR #1029, opened 2023-03-06
Bound handling

PaulJonasJost:
Could go into more detail for the x_guesses but not really needed imho

### DIFF
--- a/pypesto/problem/base.py
+++ b/pypesto/problem/base.py
@@ -1,4 +1,5 @@
 import copy
+import logging
 from typing import Iterable, List, Optional, SupportsFloat, SupportsInt, Union
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


pypesto/problem/base.py
```diff
@@ -231,6 +235,20 @@ def normalize(self) -> None:
...
+        """Check whether the supplied x_guesses adhere to the bounds."""
+        if self.x_guesses == np.zeros((0, self.dim_full)):
+            return
+        adheres_ub = self.x_guesses < self.ub
+        adheres_lb = self.x_guesses < self.lb
+        adheres_bounds = adheres_ub & adheres_lb
+        # if any bounds are violated, log a warning
+        if not adheres_bounds.all():
+            logger.warning(
+                "WARNING: Some initial guesses supplied violate the bounds "
```
yannikschaelte, line 248: No need to print WARNING. the logger takes care of that.

PR #1032, opened 2023-03-06
Allow custom multiprocessing context

### LINKED ISSUE #1031: Multiprocessing with PyTorch Autograd
**Bug description**
Hello guys! I am currently using the PyPesto framework for my PhD project and I ran into this issue while trying to parallelize the `optimize.minimize` routine with the option `engine = pypesto.engine.MultiProcessEngine()`.
After running the routine, if one prints one of the elements of the `result.optimize_result.as_list()`, it shows 

```python
{
'id': '0',
'x': None, 
#[...]
'exitflag': <ExitFlag.DID_NOT_RUN: 0>,
'time': 0.01579761505126953,
'message': "Unable to handle autograd's threading in combination with fork-based multiprocessing. See https://github.com/pytorch/pytorch/wiki/Autograd-and-Fork",
'optimizer': '<FidesOptimizer  hessian_update=BFGS verbose=False options={}>'
}
```
The problem seems to arise from a conflict between the `torch.autograd` package that I am using to compute the gradient of my objective function and the process instantiating method of `multiprocess.Pool`.
I tried implementing the fix suggested in the message in a forked version of pyPESTO, by editing the `execute` method of the `MultiProcessEngine` class in `pypesto.engine.multi_process.py` as 

```python
import multiprocessing as mp
# [...]

class MultiProcessEngine(Engine):
   # [...]

   def execute(self, tasks: List[Task], progress_bar: bool = True):
        # [...]

        # Experimental fix taken from https://github.com/pytorch/pytorch/wiki/Autograd-and-Fork
        # to allow multiprocessing with Pytorch Autograd
        ctx = mp.get_context('spawn')

        with ctx.Pool(processes=n_procs) as pool:
            results = pool.map(
                work, tqdm(pickled_tasks, disable=not progress_bar)
            )

        return results

   # [...]
```

I tried rerunning it, and apparently, it works fine! It seems that this issue occurs only in Linux machines as the `fork` process instantiation method is the default one on those (MacOS and Windows rely on `spawn`, and forking is not even possible on Windows as it is nicely explained [here](https://britishgeologicalsurvey.github.io/science/python-forking-vs-spawn/#:~:text=Forking%20and%20spawning%20are%20two,they%20were%20in%20the%20parent)).

I wanted to ask you if it would be convenient for you to add this "fix" in your source or if you were relying on the `fork` method for other reasons. 
By the way, props for the clean code; it is indeed very legible and easy to understand.

**Environment**
- Operating system: EndeavourOS
- `pypesto` version: 0.2.14
- Python version: 3.10.9

yannikschaelte:
Fixes #1031

### DIFF
--- a/test/base/test_engine.py
+++ b/test/base/test_engine.py
@@ -20,6 +20,8 @@ def test_basic():
         pypesto.engine.SingleCoreEngine(),
         pypesto.engine.MultiProcessEngine(n_procs=2),
         pypesto.engine.MultiProcessEngine(n_procs=2, method="spawn"),
+        pypesto.engine.MultiProcessEngine(n_procs=2, method="fork"),
+        pypesto.engine.MultiProcessEngine(n_procs=2, method="forkserver"),
         pypesto.engine.MultiThreadEngine(n_threads=4),
     ]:
         _test_basic(engine)
@@ -47,6 +49,8 @@ def test_petab():
         pypesto.engine.SingleCoreEngine(),
         pypesto.engine.MultiProcessEngine(n_procs=2),
         pypesto.engine.MultiProcessEngine(n_procs=2, method="spawn"),
+        pypesto.engine.MultiProcessEngine(n_procs=2, method="fork"),
+        pypesto.engine.MultiProcessEngine(n_procs=2, method="forkserver"),
         pypesto.engine.MultiThreadEngine(n_threads=4),
     ]:
         _test_petab(engine)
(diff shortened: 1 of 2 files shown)


test/base/test_engine.py
```diff
@@ -19,7 +19,8 @@ def test_basic():
     for engine in [
         pypesto.engine.SingleCoreEngine(),
         pypesto.engine.MultiProcessEngine(n_procs=2),
-        pypesto.engine.MultiThreadEngine(n_threads=8),
+        pypesto.engine.MultiProcessEngine(n_procs=2, method="spawn"),
+        pypesto.engine.MultiThreadEngine(n_threads=4),
```
PaulJonasJost, line 23: Why is only "spawn" additionally to default tested? shouldn't it be at least one more?

PaulJonasJost (approved):

Issue #1033, opened 2023-03-07
Way to parallelize dynesty


shoepfl:
Hi pyPESTO team,

i currently try to parallelize dynesty with pyPESTO on a cluster.
However, OpenMPI, which is used to parallelize dynesty via schwimmbad is conflicting heavily with amici.

Do you know a way to create a pool that works with amici?

yannikschaelte:
@dweindl @FFroehlich this question appears rather AMICI related?

FFroehlich:
No experience with schwimmbad, but serialization will in general be tricky with AMICI. You can look at the different engines in https://github.com/ICB-DCM/pyPESTO/tree/main/pypesto/engine to get an idea how we set up parallelization in pyPESTO, but I am not sure whether all of those work with AMICI. 

Not familiar with dynesty, but I don't see a good reason why simulations in amici should be parallelized via OpenMPI since there shouldn't be any inter-process communication.

Could you provide some details about the error messages that you encounter?

yannikschaelte:
Serialization of the AmiciObjective should be possible if it was e.g. generated via PEtab, or generally has a proper AmiciObjectBuilder, see https://github.com/ICB-DCM/pyPESTO/blob/main/pypesto/objective/amici/amici.py#L252

shoepfl:
My basic idea was to only parallelize dynesty by parsing a pool to its sampler_args when it is called in pyPESTO. 
But maybe that was a naive approach anyways. 

The dynesty docs suggest the pool that is created with swimmbad, which uses mpi4py but I did not get a proper error in executing as already conda shows a huge load of conflicting packages with amici like libcblas. So I think this is not a good solution.

I also tried to create the pool with multiprocessing and parsing it to the sampler_args. Here is a minimal example:

`import pypesto
import pypesto.petab
import pypesto.optimize as optimize
import pypesto.sample as sample
import multiprocessing

M = 'synthetic_data'
Path = Path(__file__).parent / M

# Name of the yaml file of the PEtab Model
yaml_file = os.path.join(Path, M + '.yaml')

importer = pypesto.petab.PetabImporter.from_yaml(yaml_file)
problem = importer.create_problem(force_compile=True)

result = optimize.minimize(problem, n_starts=1)

dynesty_args = {'nlive': 100, 'sample': 'unif',
                'pool': multiprocessing.Pool(), 'queue_size':multiprocessing.cpu_count()}
[...]

PR #1039, opened 2023-03-15
dynesty sampler: doc parallelized sampling

dilpath:
see #1033


yannikschaelte (approved):
(Could also add snippet to docstring)

PR #1046, opened 2023-04-11
Fixed bug on existing file and no overwrite

arrjon:
Bug: If optimization, sampling results etc. are saved and a filename is specified but `overwrite` it set to `False`, then an error is thrown only when trying to save it after the whole sampling/optimization is done. This can be really annoying when sampling is costly. 

Fix: instead of an error, only a warning is raised that the file could not be saved.


dweindl:
Hi, thanks for the contribution. I am not sure if it's good to change this to a warning, though. If I call `write`, I would like to either have things written to file successfully, or have raised an exception.

arrjon:
I understand the issue. It was the fastest way to solve the problem. I will think of another solution.

yannikschaelte (approved):
looks good to me, but @PaulJonasJost you're deeper in the logic here?

pypesto/optimize/util.py
```diff
@@ -153,17 +153,18 @@ def assign_ids(
...
+        n_used = len(used_ids)
     if ids is None:
         i = 0
-        ids = [str(j) for j in range(i * n_starts, (i + 1) * n_starts)]
+        ids = [str(j) for j in range(0, n_starts)]
         while not used_ids.isdisjoint(ids):
-            i += 1
             ids = [
-                str(j) for j in range(N_used * i, N_used * (i + 1) + n_starts)
+                str(j) for j in range(n_used + i, n_used + i + 1 + n_starts)
```
PaulJonasJost, line 165:
```suggestion
                str(j) for j in range(n_used * i, n_used * i + n_starts)
```
Otherwise we create a list of ids of length n_starts+1. Should probably write a test for this.

PaulJonasJost (approved):
Thanks for the addition. Feel free to merge or add the test if you want, otherwise just send me a quick message and I will add the test.

PR #1051, opened 2023-04-27
Reworked Rosenbrock

PaulJonasJost:
Renamed it to emphasise that this is a notebook for custom objective functions.

Mainly changed from comparing different optimisers, to showing why manually supplying the gradient is really useful.

Should I add sampling as well?


yannikschaelte (approved):
I'd keep the 2d visualization of the function, as images always help understanding. Further, remark then when running profiles or sampling, the cost function will be interpreted as a log-likelihood (/ log-posterior with the boundaries).

PaulJonasJost:
> I'd keep the 2d visualization of the function, as images always help understanding. Further, remark then when running profiles or sampling, the cost function will be interpreted as a log-likelihood (/ log-posterior with the boundaries).

Yes, will add the 2d visualization again 👍🏼 

I will do so, would you add sampling additionally? Was not sure, as it would be just an addition without any gain, as it is already in the getting started notebook and the amici one

yannikschaelte:
Sampling may be superfluous here (unless we also want to nicely visualize it on the 2d problem), as there would be no additional value, as yo ustated.

PR #1052, opened 2023-04-27
Refactor docs

yannikschaelte:
* Add [+] expand buttons
* Use `autosummary` for a visual overview, and avoiding full expansion of all sub-modules
* Temporarily fixate `sphinx-rtd-theme == 1.1.0` as expansion stops working in 1.2.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,7 @@ Thumbs.db
 
 #documentation
 doc/_build
+doc/api
 
 # data
 data/raw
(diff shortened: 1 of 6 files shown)


PaulJonasJost (approved):
Yes especially API looks much better. Would probably reorganise the examples a bit but not needed now.

yannikschaelte:
agreed, let's address that in a separate pr

PR #1055, opened 2023-04-28
Removed python 3.8 support according to NEP29

PaulJonasJost:


yannikschaelte (changes requested):
nice try ^^ all tests need to be updated to 9,11.

yannikschaelte (approved):

PR #1059, opened 2023-05-02
Select: Use predecessor specified in YAML

dilpath:
If a predecessor model is specified in the PEtab Select files, it will now be used as the initial model. e.g. [test case 0009](https://github.com/PEtab-dev/petab_select/tree/main/test_cases/0009).


FFroehlich:
why was this merged without passing tests?

dilpath:
> why was this merged without passing tests?

One failure is due to a package dependency, which I will try to fix in #1070 and merge soon

PR #1061, opened 2023-05-04
Parameter plot w/ hier. pars, noise estimation for splines

Doresic:
Extended the parameter plots to include hierarchical parameters. Not all, but the interpretable ones: scaling, offset, and noise. So excluding the very numerous spline/optimal scaling inner parameters. 
- The amici objective has an attribute `self.inner_parameters` in which it will save the best last inner parameters of the current optimization run (start point).
- Then at the end of the optimization run, it will be saved into the `optimization_result` dictionary.
- The parameter plotting routine will check for inner parameters in `parameters.py/handle_inputs` and concatenate them to the outputs accordingly.

Implemented the noise estimation for observables with non-linear monotone data. 
- The noise is calculated analytically, basically in the same way as with relative data. 
- To specify that the noise parameters should be optimized hierarchically, as with relative data, the `parameterType` column entry of the noise parameter should be `sigma`. Then the `SplineInnerProblem` will extract the noise parameters.


Doresic:
Example of what the plot looks like for the Boehm model with hierarchically estimated noise:

image

Since the scale of the hierarchical parameters is `lin`, it will be harder to see the kinetic parameters nicely. So I've added a flag `plot_inner_parameters` to the parameters plotting routine. The default is True.

pypesto/objective/amici/amici.py
```diff
@@ -214,6 +214,9 @@ def __init__(
         # `set_custom_timepoints` method for more information.
         self.custom_timepoints = None
 
+        # Initialize the dictionary for saving of inner parameters.
+        self.inner_parameters = {}
```
FFroehlich, line 218:
```suggestion
        self.inner_parameters: Dict[str,float] = dict()
```

to get type annotations right, also not intended to initalize as empty set, right?

FFroehlich (approved):
Graph looks nice! Only looked at changes is `pypesto/objective/amici/amici.py` and `pypesto/petab/importer.py`

pypesto/objective/amici/amici.py
Doresic: Yeah typo 😅

pypesto/petab/importer.py
```diff
@@ -700,10 +706,17 @@ def create_problem(
                 )
         # In case of hierarchical optimization, parameters estimated in the
         # inner subproblem are removed from the outer problem
-        if not self._non_quantitative_data_types and self._hierarchical:
-            inner_parameter_ids = (
-                objective.calculator.inner_problem.get_x_ids()
-            )
+        if self._hierarchical:
+            if self._non_quantitative_data_types:
```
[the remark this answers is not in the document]
Doresic: Yeah, true. Will be much cleaner like that.

dweindl (approved):
Thanks. Looks good as far as I can tell.

pypesto/objective/amici/amici.py
Doresic: Oh actually flake8 doesn't like dict()
image

pypesto/hierarchical/spline_approximation/calculator.py
```diff
@@ -140,10 +147,15 @@ def __call__(
...
         # If AMICI ReturnData is not provided, we need to simulate the model
+
         if rdatas is None:
             amici_solver.setSensitivityOrder(sensi_order)
 
             x_dct = copy.deepcopy(x_dct)
+            x_dct.update(
+                self.inner_problem.get_noise_dummy_values(scaled=True)
+            )
+            breakpoint()
```
dilpath, line 158:
```suggestion
```

pypesto/hierarchical/spline_approximation/problem.py
```diff
@@ -249,6 +314,58 @@ def spline_inner_parameters_from_measurement_df(
...
+                inner_parameter_type=InnerParameterType.SIGMA,
+                scale=LIN,
+                lb=row[LOWER_BOUND],
+                ub=row[UPPER_BOUND],
+                observable_id=observable_id,
+                group=group,
+                index=None,
+                estimate=estimate,
+            )
+        )
```
dilpath, line 364: Can simply set `estimate=True`

```suggestion
    # Create noise inner parameters.
    for _, row in parameter_df.iterrows():
        if row[PARAMETER_ID] not in noise_parameter_to_observable:
            continue
        if petab.is_empty(row.get(PARAMETER_TYPE)) or not row[ESTIMATE]:
            continue
        observable_id = noise_parameter_to_observable[row[PARAMETER_ID]]
        group = observable_ids.index(observable_id) + 1

        noise_parameters.append(
            SplineInnerParameter(
                inner_parameter_id=row[PARAMETER_ID],
                inner_parameter_type=InnerParameterType.SIGMA,
                scale=LIN,
                lb=row[LOWER_BOUND],
                ub=row[UPPER_BOUND],
                observable_id=observable_id,
                group=group,
                index=None,
                estimate=True,
            )
        )
```

pypesto/hierarchical/spline_approximation/solver.py
```diff
@@ -243,13 +277,26 @@ def calculate_gradients(
                         min_diff_factor=self.options[MIN_DIFF_FACTOR],
                     )
 
-                    # Calculate df_ds term only if mu is not all 0
+                    # If the spline parameter is at its boundary, the
+                    # corresponding Langrangian multiplier mu is set to 0.
```
dilpath, line 281:
```suggestion
                    # corresponding Lagrangian multiplier mu is set to 0.
```

test/hierarchical/test_spline.py
```diff
@@ -210,7 +215,7 @@ def _inner_problem_exp():
     n_spline_pars = int(np.ceil(spline_ratio * len(timepoints)))
 
     expected_values = {
-        'fun': 0.0,
+        'fun': np.log(2 * np.pi) * 11 / 2,
```
dilpath, line 218: Lots of magic numbers in this file, .e.g `11` here. Are they explained somewhere?

dilpath (approved):

test/hierarchical/test_spline.py
Doresic: True :D yes, it's the number of timepoints/datapoints in this example problem. I added a variable.

Doresic:
Base test is failing very often. Specifically `test_objective.py/test_finite_difference_checks()`:

`
# Test the single step size `check_grad` method.
        eps = 1e-5
        result_single_eps = objective.check_grad(np.array([theta]), eps=eps)
>       assert result_single_eps['rel_err'].squeeze() == rel_err(eps)
E       assert 1.1987971168495947e-10 == 1.1987977630586244e-10
E        +  where 1.1987971168495947e-10 = <bound method NDFrame.squeeze of x_0    1.198797e-10\nName: rel_err, dtype: float64>()
E        +    where <bound method NDFrame.squeeze of x_0    1.198797e-10\nName: rel_err, dtype: float64> = x_0    1.198797e-10\nName: rel_err, dtype:
`

It seems that the assert is checking if very small numbers are equal.
Is this failing in other PRs?

dweindl:
> Is this failing in other PRs?

Dunno, but it definitely makes sense to change that to something like [numpy.testing.assert_array_almost_equal_nulp](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal_nulp.html#numpy.testing.assert_array_almost_equal_nulp)

PaulJonasJost (approved):
Looks good to me. Do we have a test for the inner parameter visualisation? Could be useful just to ensure its running later on (as other visualization tests)

PaulJonasJost:
> > Is this failing in other PRs?
> 
> Dunno, but it definitely makes sense to change that to something like [numpy.testing.assert_array_almost_equal_nulp](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal_nulp.html#numpy.testing.assert_array_almost_equal_nulp)

should be fixed in #1064

yannikschaelte (approved):

stephanmg (approved):
Looks fine. I agree on Paul's comment, a test for inner parameter visualization would be great.

PR #1063, opened 2023-05-08
Fix Ensemble Bug

PaulJonasJost:
Speedup in history choosing. Now also covers the case in which the max size is smaller then the number of starts, as before this resulted in no ensemble


pypesto/ensemble/ensemble.py
```diff
@@ -1111,29 +1111,25 @@ def entries_per_start(
     ens_ind = [np.flatnonzero(fval <= cutoff) for fval in fval_traces]
 
     # count the number of candidates per start
-    n_per_start = np.array([len(start) for start in ens_ind])
+    n_theo = np.array([len(start) for start in ens_ind])
```
dilpath, line 1114: This and the previous line look like they could be combined into

```suggestion
    n_theo = np.array([(fval <= cutoff).sum() for fval in fval_traces)])
```

dilpath (approved):
:+1:

m-philipps:
I'm not sure if this is already the desired behaviour in `get_vector_indices` from line 1137, especially when not using a `cutoff`. 

When `distribute = True` but `n_vectors` is 1 or 2, the optimization startpoint or start- & endpoint are used, respectively.

When `distribute = False` the first `n_vector` points will be used, i.e. the startpoint and first steps.

PR #1065, opened 2023-05-11
AdaptiveMetropolis failure fix for bounded priors

Theorell:
If both the likelihood and the prior are -inf, llh_new on line 112 in metropolis.py becomes nan. This nan propagates and causes the sampler to fail. This happens frequently for bounded priors, that return -inf outside the feasible region.
Added an if clause catching log_p_acc=nan, occurring when prior is nan. Probably, a more elegant solution is possible.


dilpath (approved):
Thanks! Looks good to me. Failing tests look irrelevant, except for "quality". Could you run `black` on your branch? This can be done e.g. with the pre-commit hooks, as described here: https://pypesto.readthedocs.io/en/latest/contribute.html

PR #1068, opened 2023-05-25
AggregatedObjective: objective-specific kwargs for `call_unprocessed`

Theorell:
It should be possible to pass kwargs to AmiciObjective. At the moment, if you make an AggregatedObjective with a prior (which takes kwargs), and an AmiciObjective, trying to use the kwargs will lead to an error.


dilpath:
Looks reasonable but might be problematic to assume that keyword arguments between objectives are comparable.

Instead, we could change `AggregatedObjective.call_unprocessed` to accept a keyword argument `kwargs_list`, which is one dictionary of keyword arguments per objective. `**kwargs` can still be used for all objectives as before.

I made this suggestion in a commit, feel free to revert if undesired.

pypesto/objective/aggregated.py
```diff
@@ -81,18 +81,38 @@ def call_unprocessed(
...
         See `ObjectiveBase` for more documentation.
 
         Main method to overwrite from the base class. It handles and
         delegates the actual objective evaluation.
+
+        Parameters
+        ----------
+        kwargs_list:
+            Objective-specific keyword arguments, where the dictionaries are
+            ordered by the objectives.
```
dweindl, line 97: Makes sense to have it objective-specific :+1:

pypesto/objective/aggregated.py
```diff
@@ -81,18 +81,38 @@ def call_unprocessed(
...
-                objective.call_unprocessed(x, sensi_orders, mode, **kwargs)
-                for objective in self._objectives
+                objective.call_unprocessed(
+                    x,
+                    sensi_orders,
+                    mode,
+                    **kwargs,
+                    **kwargs_list[i],
+                )
+                for i, objective in enumerate(self._objectives)
```
dweindl, line 115:
```suggestion
                    **cur_kwargs,
                )
                for objective, cur_kwargs in zip(self._objectives, kwargs_list or [{}] * len(self._objectives))
```

pypesto/objective/amici/amici.py
```diff
@@ -402,6 +402,7 @@ def call_unprocessed(
         edatas: Sequence['amici.ExpData'] = None,
         parameter_mapping: 'ParameterMapping' = None,
         amici_reporting: Optional['amici.RDataReporting'] = None,
+        **kwargs,
```
dweindl, line 405: This wouldn't be required anymore if we allow for objective-specific kwargs, right?

dilpath: Yes, I would opt to remove/deprecate it. Only kept it for now to not break anyone's code.

dweindl: The one in `AggregatedObjective`, yes. But the one in `AmiciObjective` is only introduced here.

dilpath: Oops. Removed now.

FFroehlich (approved):
👍

PR #1073, opened 2023-06-02
Fix for petab 0.2.x

FFroehlich:


FFroehlich:
looking for volunteers to wrap this up

dweindl:
> looking for volunteers to wrap this up

Done

dweindl (approved):

.github/workflows/ci.yml
```diff
@@ -73,7 +73,7 @@ jobs:
       run: .github/workflows/install_deps.sh amici pysb
 
     - name: Run tests
-      timeout-minutes: 25
+      timeout-minutes: 35
```
yannikschaelte, line 76: if it takes that long out of a sudden, something should be wrong. was 15 not long ago

yannikschaelte (approved):

.github/workflows/ci.yml
dweindl: The number of test cases increased:
* additional test cases 
* running for both SBML and PySB models

PR #1081, opened 2023-06-19
Fix error in bound checking.

### LINKED ISSUE #1080: Parameter Bound Check
## Issues
I generate Latin Hypercube samples like this.

```python
# problem: pypesto.Problem 
method = pypesto.startpoint.LatinHypercubeStartpoints(
            use_guesses=False, check_fval=True, check_grad=True
        )
        init_samples = method.sample(
            n_starts=num_starts,
            lb=problem.lb_full,
            ub=problem.ub_full,
        )
problem.set_x_guesses(init_samples)
```

Yet during parameter check I get the warning:

```bash
19 [2023-06-16 21:54:58,562][pypesto.problem.base][WARNING] - Some initial guesses supplied violate the bounds set for this problem.
20 Some initial guesses supplied violate the bounds set for this problem.
```

## Docs Example 

Here I see this warning shows [Objective + Problem Definition Example](https://pypesto.readthedocs.io/en/latest/example/custom_objective_function.html#1.-Objective-+-Problem-Definition)

## Proposed Solution 
I think the issue is here in [src](https://github.com/ICB-DCM/pyPESTO/blob/27a48d28f38e4d13a0f5c24f129f38bb16659ec1/pypesto/problem/base.py#LL238C1-L251C1), I've added `# Issue Line` below.
```
    def _check_x_guesses(self):
        """Check whether the supplied x_guesses adhere to the bounds."""
        if self.x_guesses == np.zeros((0, self.dim_full)):
            return
        adheres_ub = self.x_guesses < self.ub
        adheres_lb = self.x_guesses < self.lb # Issue Line
        adheres_bounds = adheres_ub & adheres_lb
        # if any bounds are violated, log a warning
        if not adheres_bounds.all():
            logger.warning(
                "Some initial guesses supplied violate the bounds "
                "set for this problem."
            )
```

`adheres_ub` and adheres_lb both have the same inequality. I ran some checks locally, it seems the `# Issues Line` should be         `adheres_lb = self.x_guesses < self.lb`. If necessary I could submit a PR. Thanks.

PaulJonasJost:
Closes #1080.


pypesto/problem/base.py
```diff
@@ -239,8 +239,8 @@ def _check_x_guesses(self):
         """Check whether the supplied x_guesses adhere to the bounds."""
         if self.x_guesses.size == 0:
             return
-        adheres_ub = self.x_guesses < self.ub
-        adheres_lb = self.x_guesses < self.lb
+        adheres_ub = self.x_guesses <= self.ub
+        adheres_lb = self.x_guesses <= self.lb
```
yannikschaelte, line 243:
```suggestion
        adheres_lb = self.x_guesses >= self.lb
```

yannikschaelte (approved):

PR #1086, opened 2023-06-23
Speed up petab testsuite

PaulJonasJost:
Should speed up the tests potentially by removing multiple amici compilations.

Supposed to do three things:
* Fix the error of the notebook2 test run by changing from MultiThread to MultiProcess and reducing the size of the ensembles
* Speed up petab tests by compiling the required amici models once for all tests that use them. This does not reduce the speed by leaps and bounds, as most of the time is used on amici compilation of the petab_testsuite
* to also tackle this, petab tests should now only run if any relevant files are changed in PRs. This would have the downside that petab/amici might get new versions and thus without any changes the test would now fail which would only be tested **when** someone changes a file (this is potentially taken care of since the workflow still triggers on pushes to main?). It would have the upside (especially if done for all tests) that in a PR only the tests that might be influenced by changes will be run.

Happy for discussions on this :)


test/petab/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@pytest.fixture(scope="session")
+def loaded_models():
+    """Fixture to load all models used in Tests to be only loaded once."""
+    model_names = [
+        "Zheng_PNAS2012",
+        "Brannmark_JBC2010",
+        "Boehm_JProteomeRes2014",
+    ]
```
dweindl, line 17: Hm... So whenever a test uses any of these models, all of them have to be imported? I am not convinced that this is an improvement. Running only a single test is not that uncommon.

When called twice on the same model (from within the same directory), the PetabImporter should already reuse any already imported model. Is this not happening?

Moving  the PEtab import of models that are used in several tests to (individual) session-scoped fixtures makes sense though. (Ideally having both, the petab.Problem and pypesto.Problem in that fixture).

dweindl (changes requested):

.github/workflows/ci.yml
```diff
@@ -73,9 +91,14 @@ jobs:
       run: .github/workflows/install_deps.sh amici pysb
 
     - name: Run tests
+      if: steps.filter.outputs.petab_filter == 'true'
```
dweindl, line 94: I like the conditional tests for pull requests. For commits on `develop` and `main`, I'd prefer running them unconditionally, though.

PaulJonasJost: that could be a solution for this. Not entirely sure how to set this as an additional condition but will check 👍🏼

test/petab/conftest.py
PaulJonasJost: right, did not think about only individual tests. could just write a session wide fixture for models that are used and use that fixture then.

PR #1089, opened 2023-06-28
PetabJL integration

PaulJonasJost:
Broaden the usability of the Julia objective with a PEtab_jl objective, that is meant to be used for integration of PEtab.jl

The objective has the option to precompile its module which speeds the multiprocessing optimization up (as otherwise the module which is the time consuming part will always have to be reloaded). Currently if I run the code manually it works fine (even exact same code as the tests) but when running the tests it fails to precompile. So the precompilation is turned off in the tests.


sebapersson (changes requested):
Thank you so much for working on this PEtab.jl importer :)

I have added a few comments to allow more efficient usage of the Julia PEtab importer.

pypesto/objective/julia/petabJl.py
```diff
@@ -0,0 +1,214 @@
...
+        petab_jl_problem = self.get(petab_problem_name)
+        self.petab_jl_problem = petab_jl_problem
+
+        # get functions
+        fun = self.petab_jl_problem.computeCost
+        grad = self.petab_jl_problem.computeGradient
+        hess = self.petab_jl_problem.computeHessian
+        x_names = np.asarray(self.petab_jl_problem.θ_estNames)
+
+        # call the super super constructor
```
yannikschaelte, line 69: super super super

yannikschaelte (approved):
would be good to be able to test precompilation too

pypesto/objective/julia/petabJl.py
```diff
@@ -0,0 +1,215 @@
...
+class PEtabJlObjective(JuliaObjective):
+    """
+    Wrapper around an objective defined in PEtab.jl.
+
+    Parameters
+    ----------
+    module:
+        Name of the julia module containing the objective.
+    source_file:
+        Julia source file. Defaults to "{module_name}.jl".
```
dweindl, line 23:
```suggestion
        Julia source file. Defaults to "{module}.jl".
```
?

pypesto/objective/julia/petab_jl_importer.py
```diff
@@ -0,0 +1,318 @@
...
+    Importer for PEtab models in Julia, using PEtab.jl.
+
+    Create an `objective.JuliaObjective` or a `pypesto.Problem` from PEtab
+    files or from a julia module.
+    """
+
+    def __init__(
+        self,
+        module: str = None,
+        source_file: str = None,
```
dweindl, line 29: document?

test/julia/test_pyjulia.py
```diff
@@ -52,3 +57,80 @@ def test_pyjulia_pipeline():
     # check with analytical value
     p_opt = obj.get("p_opt")
     assert np.allclose(result.optimize_result[0].x, p_opt)  # noqa: S101
+
+
+def test_petabJL_interface():
+    """Test the interface to PEtab.jl with provided solutions from julia."""
+    model_name = "boehm_JProteomeRes2014"
+    yaml_file = f"doc/example/{model_name}/{model_name}.yaml"
```
dweindl, line 65: Please use an absolute path based on `__file__` so pytest can be invoked from any directory. Also below.

test/julia/test_pyjulia.py
```diff
@@ -52,3 +57,80 @@ def test_pyjulia_pipeline():
...
+def test_petabJL_from_yaml():
+    """Test that PEtab.jl from yaml file is running smoothly."""
+    yaml_file = "doc/example/conversion_reaction/conversion_reaction.yaml"
+
+    importer = PetabJlImporter.from_yaml(yaml_file)
+
+    problem = importer.create_problem(precompile=False)
+
+    # optimize with single core
+    optimize.minimize(problem, engine=SingleCoreEngine(), n_starts=10)
```
dweindl, line 136: Why 10 starts if not even a single results is checked?

PR #1090, opened 2023-06-28
Restrict fval magnitude in waterfall with order_by_id

m-philipps:
When using waterfall with `order_by_id=True` add a check for `fvals < 1e100` as is in place for `order_by_id=False`.


stephanmg (commented):
1e100 is a reasonable threshold?

m-philipps:
It's the same threshold that is used in the `order_by_id=False` case
https://github.com/ICB-DCM/pyPESTO/blob/462481eecb1a9fbf803a89bf11b59485a733c8f8/pypesto/visualize/waterfall.py#L133-L138

stephanmg:
Changes looks fine, just wondering if should be made a parameter of the function. Opinions? @plakrisenko

stephanmg (approved):

plakrisenko:
there is the `delete_nan_inf` function, maybe one could modify it a bit to also work in case `order_by_id=True`

plakrisenko (approved):

plakrisenko:
> Changes looks fine, just wondering if should be made a parameter of the function. Opinions? @plakrisenko

It was introduced in https://github.com/ICB-DCM/pyPESTO/pull/999/files and is also used in parameters plot. 
One could introduce a constant, similar to https://github.com/ICB-DCM/pyPESTO/blob/develop/pypesto/C.py#L196

stephanmg:
> > Changes looks fine, just wondering if should be made a parameter of the function. Opinions? @plakrisenko
> 
> It was introduced in https://github.com/ICB-DCM/pyPESTO/pull/999/files and is also used in parameters plot. One could introduce a constant, similar to https://github.com/ICB-DCM/pyPESTO/blob/develop/pypesto/C.py#L196

Thanks. I would go for a constant. Otherwise lgtm now!

stephanmg:
Triggering a re-run of CI / base...

PaulJonasJost (approved):

Issue #1017, opened 2023-02-09
Prior definition from the Petab table


aidinbii:
Hello, I have a parameters table with defined initializationPriorParameters. 
```
importer = pypesto.petab.PetabImporter(petab_problem, model_name=model_name)
obj = importer.create_objective()
problem = importer.create_problem(obj,sensi_orders=(0,1),)
```
But when I run an optimization
```
 optimize.minimize(problem=problem,
                           optimizer=optimizer_1,
                           n_starts=20)
```
it samples the starting parameters from the lower and upper bound values (also specified in the same table)

Also, if I run:
``print(importer.create_prior()) ``
it returns "None"

dweindl:
Hi @aidinbii,

> ```
>  optimize.minimize(problem=problem,
>                            optimizer=optimizer_1,
>                            n_starts=20)
> ```
> 
> it samples the starting parameters from the lower and upper bound values (also specified in the same table)

If the initialization prior distribution is wider than the lower and upper bounds, it will be clipped at the bounds.
If your prior distribution is narrower than the bounds, something seems to be wrong. Can you show an example for your parameter table?

> Also, if I run: `print(importer.create_prior()) ` it returns "None"

This is to be expected. The initialization prior will only enter the startpoint sampling. It will not be used as prior in your objective. For the latter, use objectivePrior* in PEtab.

aidinbii:
@dweindl, sure here is the table
![Screenshot_20230210_060644](https://user-images.githubusercontent.com/43513377/218005923-6dde99bf-6062-43e6-8bf0-d20ee8914cae.png)

dweindl:
Ah, got it. Your code needs 
``` 
optimize.minimize(..., startpoint_method=importer.create_startpoint_method())
```

Agreed that it is non-obvious that this needs to be passed on top of `problem`. 

@ICB-DCM/pypesto-maintainers : Was there a good reason to not include `startpoint_method` in  `Problem`? Since `Problem` has `x_guesses`, I'd consider it quite natural to also contain any other information for startpoint sampling.

yannikschaelte:
Possible reasons against including in Problem may have been:
* Ability to recover Problem from file. This is however anyway not what we want long-term, therefore obsolete.
* Decoupling of Problem and information only relevant for Optimization. As however the guesses are also mostly optimization-relevant, I don't really see a reason not to include it.

jvanhoefer:
[...]

PR #1093, opened 2023-07-11
Add startpoint_method to Problem

### LINKED ISSUE #1035: Add startpoint_method to Problem
It's not intuitive that `x_guesses` is part of `Problem`, but `startpoint_method`s are handled separately. See also discussion in #1017. Would be great to change that.

dweindl:
It's not intuitive that `x_guesses` is part of `Problem`, but `startpoint_method`s are handled separately. See also discussion in #1017.

This patch
* adds `startpoint_method` to `Problem`
* during PEtab import, sets `Problem.startpoint_method` based on the PEtab problem

To be discussed: Do we want to keep the existing `startpoint_method` argument in the long term or should it be removed (deprecated)?

Closes #1035

### DIFF
--- a/pypesto/problem/base.py
+++ b/pypesto/problem/base.py
@@ -1,12 +1,21 @@
 import copy
 import logging
-from typing import Iterable, List, Optional, SupportsFloat, SupportsInt, Union
+from typing import (
+    Callable,
+    Iterable,
+    List,
+    Optional,
+    SupportsFloat,
+    SupportsInt,
+    Union,
+)
 
 import numpy as np
 import pandas as pd
 
 from ..objective import ObjectiveBase
 from ..objective.priors import NegLogParameterPriors
+from ..startpoint import StartpointMethod, to_startpoint_method, uniform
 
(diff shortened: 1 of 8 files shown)


yannikschaelte (approved):
looks reasonable to me! I'd suggest deprecating the minimize.startpoint_method argument then.

dweindl:
On hold. Requires `dill>0.3.6` (current `master` works), otherwise the dynesty notebook fails at `sampler.save_internal_sampler("dynesty.dill")` with `TypeError: cannot pickle '_abc._abc_data' object`.

pypesto/problem/base.py
```diff
@@ -90,6 +99,7 @@ def __init__(
         lb_init: Union[np.ndarray, List[float], None] = None,
         ub_init: Union[np.ndarray, List[float], None] = None,
         copy_objective: bool = True,
+        startpoint_method: Union[StartpointMethod, Callable, bool] = None,
```
dilpath, line 102: doc

dilpath (approved):
:+1:

FFroehlich (approved):
👍

PR #1094, opened 2023-07-11
Use cloudpickle for pickling dynesty sampler

dweindl:
Cloudpickle handles ABC-derived classes properly, whereas the latest dill version on pypi (0.3.6) does not (#1093).

Cloudpickle is already installed with pypesto, so there seems to be no need to get dill on top.


dweindl:
@dilpath : Was there any specific reason for choosing dill over cloudpickle?

dilpath (approved):
Fine for me :+1: I chose `dill` out of familiarity, I haven't tested cloudpickle much.

Issue #1096, opened 2023-07-14
Default startpoint method, hierarchically optimized parameters


Doresic:
Connected to this PR https://github.com/ICB-DCM/pyPESTO/pull/1093
A possible issue with the new implementation of  ` startpoint_method ` in pypesto. 

When not assigning the standpoint method, it will default into the `self.create_startpoint_method()` (petab sampling) while creating problem, and will not default into the `uniform` one from pypesto anymore.

In case of hierarchically optimized noise, the `uniform` startpoint method was sampling correctly only non-hierarchically optimized parameters (in case of Boehm with hierarchically optimized noise, that was 6). But the petab sampling will sample all parameters in the parameter df with estimate==1, so including the noise (9 parameters). I can imagine this occurring with any hierarchically optimized parameters.

This then breaks when appending the sampled startpoints at https://github.com/ICB-DCM/pyPESTO/blob/9a754573a7b77d30d5dc1f67a8dc1be6c29f1640/pypesto/startpoint/base.py#L104C20-L104C20

Not sure what's the best fix. Is it better if create_startpoint_method defaults into uniform, as it did in the pypesto.optimize.minimize() function? (not sure if this breaks something else)

dweindl:
You are right, that case is not properly handled. It's concerning that zero tests failed. :thinking: 

> Is it better if create_startpoint_method defaults into uniform, as it did in the pypesto.optimize.minimize() function?

For any PEtab import, it should use what is specified in the PEtab problem. As this is mandatory in PEtab, we don't need any default there. In fact, a reason for #1093 was that so far, the PEtab settings for startpoint sampling would have been easily ignored.

> Not sure what's the best fix.

I'd say `PetabImporter.create_startpoint_method` should get a parameter IDs or parameter indices argument (optional, default to all parameters), that needs to be passed appropriately here: https://github.com/ICB-DCM/pyPESTO/blob/c7a88ead25385b043e0dea1a9b46a3bb4719c796/pypesto/petab/importer.py#L743
Would that address your concerns?

PR #1105, opened 2023-07-25
Fix startpoint sampling for hierarchical optimization

dweindl:
See #1096

### DIFF
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -758,7 +758,7 @@ class PetabImporter(AmiciObjectBuilder):
             x_scales=x_scales,
             x_priors_defs=prior,
             startpoint_method=self.create_startpoint_method(
-                x_ids=np.delete(x_ids, x_fixed_indices), **kwargs
+                x_ids=np.delete(x_ids, x_fixed_indices)
             ),
             **problem_kwargs,
         )
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -631,20 +631,37 @@ class PetabImporter(AmiciObjectBuilder):
         else:
             return None
 
-    def create_startpoint_method(self, **kwargs) -> StartpointMethod:
+    def create_startpoint_method(
+        self, x_ids: Sequence[str] = None, **kwargs
+    ) -> StartpointMethod:
         """Create a startpoint method.
 
         Parameters
         ----------
+        x_ids:
+            If provided, create a startpoint method that only samples the
+            parameters with the given IDs.
         **kwargs:
             Additional keyword arguments passed on to
             :meth:`pypesto.startpoint.FunctionStartpoints.__init__`.
         """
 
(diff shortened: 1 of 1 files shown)


pypesto/petab/importer.py
```diff
@@ -740,7 +757,9 @@ def create_problem(
             x_names=x_ids,
             x_scales=x_scales,
             x_priors_defs=prior,
-            startpoint_method=self.create_startpoint_method(),
+            startpoint_method=self.create_startpoint_method(
+                x_ids=np.delete(x_ids, x_fixed_indices)
+            ),
```
Doresic, line 762: This will not work with models that have hierarchical parameters and `ESTIMATE==0` parameters (the issue is in the order of removing parameters from x_ids). 

In case of Boehm with `['Epo_degradation_BaF3', 'k_exp_hetero', 'k_exp_homo', 'k_imp_hetero', 'k_imp_homo', 'k_phos', 'ratio', 'sd_pSTAT5A_rel', 'sd_pSTAT5B_rel', 'sd_rSTAT5A_rel', 'specC17']` parameters, the ones that are hierarchical are `'sd_pSTAT5A_rel', 'sd_pSTAT5B_rel', 'sd_rSTAT5A_rel'` and these will be removed in these lines [here](https://github.com/ICB-DCM/pyPESTO/blob/9a754573a7b77d30d5dc1f67a8dc1be6c29f1640/pypesto/petab/importer.py#L704-L713). So then the 
`x_ids=['Epo_degradation_BaF3', 'k_exp_hetero', 'k_exp_homo', 'k_imp_hetero', 'k_imp_homo', 'k_phos', 'ratio', 'specC17']`. 

But then in when trying to call `np.delete(x_ids, x_fixed_indices)` it will give index error, because `x_fixed_indices` are `[6, 10]`, but the length has been reduced to 8 when the hierarchical ones were removed. 

So probably the easiest fix is using `x_ids = [x for x in x_ids if x not in self.petab_problem.x_fixed_ids]`. Or changing order of the deletions, but taking care that `x_scales` has to be made for fixed parameters too (probably).

Doresic (approved):
Thank you for this fix. It looks like the best one. There's just a small bugfix that we need to take care of.

pypesto/petab/importer.py
dweindl: Hm, I don't think I can follow.

Around the line you referenced, `x_ids` is updated to only hold the outer parameters,  `x_fixed_indices` are updated to refer to this shortened list.
Later, for the sampling, we only need the outer parameters that are not fixed, which is what the `delete` is taking care of.

So, specifically, I don't think

> because x_fixed_indices are [6, 10], but the length has been reduced to 8

is true after `x_fixed_indices = list(x_ids.index, self.petab_problem.x_fixed_ids))`

Doresic: That's true. I missed the line where `x_fixed_indices` are updated.

FFroehlich (approved):

PR #1106, opened 2023-07-28
Hierarchical parameter plot fix

Doresic:
The issue was that the parameter plot was not showing hierarchical parameters in some cases.

If, in any case (because some inner optimization failed, simulation failed, or some other reason) there is an `optimize_result`dictionary in `result.optmize_result.list` which doesn't contain the `INNER_PARAMETERS` key and its item, the parameter plot would fail at line [381](https://github.com/ICB-DCM/pyPESTO/blob/9a754573a7b77d30d5dc1f67a8dc1be6c29f1640/pypesto/visualize/parameters.py#L381). 

This is a fix to make it robust to this. If any of the `optimize_result` objects have an `INNER_PARAMETER` key, then the other ones will be filled with NaN values and plotted.

### DIFF
--- a/pypesto/visualize/parameters.py
+++ b/pypesto/visualize/parameters.py
@@ -381,11 +381,10 @@ def handle_inputs(
         res.get(INNER_PARAMETERS, None) for res in result.optimize_result.list
     ]
     if any(inner_xs):
-        inner_xs_names = next(
-            list(inner_xs_idx.keys())
-            for inner_xs_idx in inner_xs
-            if inner_xs_idx is not None
-        )
+        for inner_xs_idx in inner_xs:
+            if inner_xs_idx is not None:
+                inner_xs_names = list(inner_xs_idx.keys())
+                break
         inner_xs = [
             [np.nan for i in range(len(inner_xs_names))]
             if inner_xs_idx is None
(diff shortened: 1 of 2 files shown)


stephanmg:
LGTM.

yannikschaelte (approved):

stephanmg (approved):

pypesto/visualize/parameters.py
```diff
@@ -377,15 +377,25 @@ def handle_inputs(
...
-        inner_xs = [list(inner_xs_idx.values()) for inner_xs_idx in inner_xs]
+    inner_xs = [
+        res.get(INNER_PARAMETERS, None) for res in result.optimize_result.list
+    ]
+    if any(inner_xs):
+        inner_xs_names = next(
+            list(inner_xs_idx.keys())
+            for inner_xs_idx in inner_xs
+            if inner_xs_idx is not None
+        )
```
dilpath, line 388:
```suggestion
        for inner_xs_idx in inner_xs:
            if inner_xs_idx is not None:
                inner_xs_names = list(inner_xs_idx.keys())
                break
```

dilpath (approved):

pypesto/visualize/parameters.py
```diff
@@ -377,15 +377,25 @@ def handle_inputs(
...
+        inner_xs = [
+            [np.nan for i in range(len(inner_xs_names))]
```
dweindl, line 390:
```suggestion
            np.full(inner_xs_names, np.nan)
```
?

dweindl (approved):

PR #1109, opened 2023-08-21
Fix #1108

### LINKED ISSUE #1108: waterfall plot y limits not working
It is not always possible to set the y limits of a waterfall plot using the `y_limits` argument, because
this condition:

https://github.com/ICB-DCM/pyPESTO/blob/2eb0d15f4aa636f873785a3c1e9e7afce281755f/pypesto/visualize/misc.py#L173

needs to be fulfilled in order to reach this line:

https://github.com/ICB-DCM/pyPESTO/blob/2eb0d15f4aa636f873785a3c1e9e7afce281755f/pypesto/visualize/misc.py#L189

m-philipps:
- enable setting the y limits of a waterfall plot by using the `y_limits` argument
- use default y limits for zoomed in starts
- fix #1108

### DIFF
--- a/pypesto/visualize/misc.py
+++ b/pypesto/visualize/misc.py
@@ -186,7 +186,7 @@ def process_y_limits(ax, y_limits):
                 )
 
             # set limits
-            ax.set_ylim(y_limits)
+        ax.set_ylim(y_limits)
 
     else:
         # No limits passed, but if we have a result list: check the limits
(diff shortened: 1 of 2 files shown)


pypesto/visualize/misc.py
```diff
@@ -186,7 +186,7 @@ def process_y_limits(ax, y_limits):
                 )
 
             # set limits
```
PaulJonasJost, line 188:
```suggestion
        # set limits
```

PaulJonasJost (approved):
Thanks, looks good 👍🏼

plakrisenko (approved):

PR #1110, opened 2023-08-22
SacessOptimizer: retry reading, delay deleting

dweindl:
* delete temporary files only after *all* were read successfully
* retry reading temp file upon error to work around potential filesystem latency issues

### DIFF
--- a/pypesto/optimize/ess/sacess.py
+++ b/pypesto/optimize/ess/sacess.py
@@ -173,10 +173,17 @@ class SacessOptimizer:
             tmp_result_filename = SacessWorker.get_temp_result_filename(
                 worker_idx
             )
-            tmp_result = read_result(
-                tmp_result_filename, problem=False, optimize=True
-            )
-            os.remove(tmp_result_filename)
+            try:
+                tmp_result = read_result(
+                    tmp_result_filename, problem=False, optimize=True
+                )
+            except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


yannikschaelte (approved):
lgtm

pypesto/optimize/ess/sacess.py
```diff
@@ -173,10 +173,17 @@ def _create_result(self, problem: Problem) -> pypesto.Result:
...
-                tmp_result_filename, problem=False, optimize=True
-            )
-            os.remove(tmp_result_filename)
+            try:
+                tmp_result = read_result(
+                    tmp_result_filename, problem=False, optimize=True
+                )
+            except FileNotFoundError:
+                # wait and retry, maybe the file wasn't found due to some filesystem latency issues
+                time.sleep(10)
```
yannikschaelte, line 182: 10s might be too long? Shouldn't any latency issues be resolved within fractions of a second?

dweindl: Well, depends on the latency :sweat_smile:
I can also change it to 5s, although I don't think it's really an issue. 
Ideally, this will never trigger, but if it does it means we had some latency issue and we are happy to have waited 10s instead of failed. If the file doesn't exist at all, then we will waste a couple of seconds, but if that happens, anyways something went severely wrong.

yannikschaelte: :+1:

PR #1112, opened 2023-08-23
SacessOptimizer: Fix logging with multiprocessing

dweindl:
Logging from SacessOptimizer didn't work properly, none of the worker log records were shown.
This PR introduces a separate process for collecting messages from sacess-workers and logging them.
Also introduces more descriptive logger names for the individual workers.


yannikschaelte (approved):
makes sense

vwiela (commented):
Looks good to me and improved some error messages, when the worker crashed.

PR #1113, opened 2023-08-24
Other platform tests

yannikschaelte:
* Also run the base test suite on half-eaten apples
 * Provide a very core suite of workflow tests for windows (to be extended if there is user demand; in particular an AMICI set up would require further work)
 
Aside:
* Run tests on Python 3.11 instead 3.10 as numba supports it now

### DIFF
--- a/test/base/test_workflow.py
+++ b/test/base/test_workflow.py
@@ -1,0 +1,123 @@
+"""Test a basic workflow, with minimum dependencies.
+
+These tests are not for correctness, but for basic functionality.
+"""
+
+import pypesto
+import pypesto.optimize as optimize
+import pypesto.profile as profile
+import pypesto.sample as sample
+import pypesto.visualize as visualize
+
+from ..util import CRProblem
+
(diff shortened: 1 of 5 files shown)


test/base/test_workflow.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # check that the results are sorted
+    fvals = result.optimize_result.fval
+    assert fvals == sorted(fvals)
+
+    # check that optimization was successful
+    assert fvals[0] < crproblem.get_fnllh()(crproblem.p_true)
+
+    # visualize the results
+    visualize.waterfall(result)
```
dweindl, line 59: use [`close_fig`](https://github.com/ICB-DCM/pyPESTO/blob/f29c848a1806c4e0fb10025978cf66bc9f9866bb/test/visualize/test_visualize.py#L27) for all tests involving plotting?

dweindl (approved):
Great, thanks!

PR #1115, opened 2023-08-31
SacessOptimizer: tmpdir option

dweindl:
SacessOptimizer:
* Add option to pass a custom directory for temporary files
* Make the default temporary file paths unique, to avoid name collisions when running multiple SacessOptimizer instances from within the same working directory


vwiela (approved):
Great additional option to use the sacess optimizer in job_arrays.

PR #1118, opened 2023-09-14
{id} in history storage filename

plakrisenko:
At the moment, if MultiProcessEngine is used and HistoryOptions storage_file contains "{id}" it lead to an error. 

It's a minor thing, but is inconvenient when switching between SingleCoreEngine and MultiProcessEngine


PaulJonasJost (approved):
Looks good to me

PR #1124, opened 2023-10-06
Update ensemble.py

vwiela:
Add an option for computing cutoff values for pointwise confidence regions. Needed for ensemble simulations.

### DIFF
--- a/pypesto/ensemble/ensemble.py
+++ b/pypesto/ensemble/ensemble.py
@@ -1204,7 +1204,11 @@ def get_percentile_label(percentile: Union[float, int, str]) -> str:
     return f'{PERCENTILE} {percentile}'
 
 
-def calculate_cutoff(result: Result, percentile: float = 0.95):
+def calculate_cutoff(
+    result: Result, 
+    percentile: float = 0.95,
+    cr_option: string = "simultaneous",
+):
     """
     Calculate the cutoff of the ensemble.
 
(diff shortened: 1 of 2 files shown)


pypesto/ensemble/ensemble.py
```diff
@@ -1220,6 +1224,10 @@ def calculate_cutoff(result: Result, percentile: float = 0.95):
         The percentile of the chi^2 distribution. Between 0 and 100.
         Higher values will result in a more lax cutoff. If the value is greater
         than 100, the cutoff will be returned as np.inf.
+    cr_option:
+        The type of confidence region, which determines the degree of freedom of
+        the chi^2 distribution for the cutoff value. It can take 'simultaneous' or
+        'pointwise'.
```
PaulJonasJost, line 1230: A small link to a paper to what the difference is would be nice for users.

PaulJonasJost (approved):

pypesto/ensemble/ensemble.py
dilpath: Agreed, but some internet article, e.g. Wikipedia, would also be fine for me

pypesto/ensemble/ensemble.py
```diff
@@ -1204,7 +1204,11 @@ def get_percentile_label(percentile: Union[float, int, str]) -> str:
     return f'{PERCENTILE} {percentile}'
 
 
-def calculate_cutoff(result: Result, percentile: float = 0.95):
+def calculate_cutoff(
+    result: Result,
+    percentile: float = 0.95,
+    cr_option: str = 'simultaneous',
```
dilpath, line 1210: Add
```python
SIMULTANEOUS = 'simultaneous'
POINTWISE = 'pointwise
```
to C.py in the ensemble section [1] and use here.

[1]
https://github.com/ICB-DCM/pyPESTO/blob/622145db3fdf3c3e2747c34ba3e563fb34d58406/pypesto/C.py#L11

Alternatively, change to boolean argument `simultaneous=True` and document the meaning of `True` and `False` in the docstring

dilpath (approved):

pypesto/ensemble/ensemble.py
vwiela: Done.

PR #1130, opened 2023-10-10
fix ensembles from optimization endpoints

plakrisenko:
If only the names of free parameters are taken https://github.com/ICB-DCM/pyPESTO/blob/main/pypesto/ensemble/ensemble.py#L663, 
then maybe the same should be true for the parameter values


vwiela (approved):
Fixes the issue of getting constant predictions.

PaulJonasJost (approved):

dilpath (approved):
Thanks!

dilpath:
@PaulJonasJost is this an issue in `Ensemble.from_optimization_history` too?

PR #1139, opened 2023-10-19
Visualization: Fix `legend` argument checking for waterfall/parameter/history plots 

m-philipps:
- before the fix a `legend_error` would be thrown when the user supplied the legends as a tuple
- Now handles all data types
- raises `TypeError` for single and multiple results
- raises `ValueError` in case of multiple results when lengths of results and legends do not match


pypesto/visualize/misc.py
```diff
@@ -80,18 +80,16 @@ def process_result_list(
                 legends.append('Result ' + str(i_leg))
         else:
             # legends were passed by user: check length
-            if isinstance(legends, list):
+            try:
+                if isinstance(legends, str):
+                    legends = [legends]
```
PaulJonasJost, line 85: This checks if legends is a string (returns False in case of tuple). If it is it will create a list of length 1 which should always throw an error then.

to catch tuples and and other sequence types, perhaps rather use `legends  list(legends)`.
```suggestion
                legends = list(legends)
```

pypesto/visualize/misc.py
```diff
@@ -80,18 +80,16 @@ def process_result_list(
...
                 if len(legends) != len(results):
-                    legend_error = True
-            else:
-                legend_error = True
```
PaulJonasJost, line 87: If you do not use `legend_error` anymore, also remove `legend_error` in line 52 👍🏼

PaulJonasJost (changes requested):
Thanks for noticing, looks good aside from some slight changes .

pypesto/visualize/misc.py
m-philipps: The idea was that a string of length `!= len(results)` would throw an error, but a string with the appropriate number of characters would be used for the legend. We could also raise an error for any string, but this would not be consistent with the type hint for the legend argument of `visualize.waterfall`: `legends: Optional[Union[Sequence[str], str]] = None,` because a string is a Sequence.

I don't know what the second suggestion is. A tuple works now, as it should. Any datatype for which the `length` function is not defined raises a `TypeError`.

PaulJonasJost: ahhh understood, thanks. Was a bit unclear for me.

m-philipps:
@PaulJonasJost what do you think of the suggestion for raising different errors for legend type and length, also in the case of a single result?

pypesto/visualize/misc.py
```diff
@@ -80,18 +84,19 @@ def process_result_list(
...
-        )
+                    raise ValueError(
+                        'List of results passed and list of labels do '
+                        'not have the same length but should. Stopping.'
+                    )
+            except TypeError:
+                legend_type_error = True
+
+    if legend_type_error:
+        raise TypeError("Unexpected legend type. Stopping")
```
PaulJonasJost, line 99:
```suggestion
        raise TypeError("Unexpected legend type.")
```
See above :)

PaulJonasJost (approved):

m-philipps:
ready to go, @plakrisenko

pypesto/visualize/misc.py
```diff
@@ -80,18 +84,19 @@ def process_result_list(
                 legends.append('Result ' + str(i_leg))
         else:
             # legends were passed by user: check length
-            if isinstance(legends, list):
+            try:
+                if isinstance(legends, str):
```
plakrisenko, line 88: if this is true it already means that 'List of results passed and list of labels do not have the same length.', right? As it's a multiple result case + only a string was passed as a legend

plakrisenko (approved):

pypesto/visualize/misc.py
m-philipps: The string could have matching length, e.g. "ABC" for 3 results. 

The idea was that a string of length != len(results) would throw an error, but a string with the appropriate number of characters would be used for the legend. We could also raise an error for any string, but this would not be consistent with the type hint for the legend argument of visualize.waterfall: legends: Optional[Union[Sequence[str], str]] = None, because a string is a Sequence.

PR #1146, opened 2023-10-23
Doc: pypesto.select xrefs + add select.postprocessors

dweindl:
Fixes formatting, adds proper cross-references, and add pypesto.select.postprocessors to the sphinx API doc.

:eyes: https://pypesto--1146.org.readthedocs.build/en/1146/api/pypesto.select.postprocessors.html


PaulJonasJost (approved):
Thanks for this, do you think we need to do this at one point for other modules as well?

dilpath (approved):
Thanks a lot! For some reason `pathlib.Path` is documented as `pypesto.select.postprocessors.Path` on RTD. I still need to go over the docs and fill in missing docstrings etc., so this can be left to another PR.

https://pypesto--1146.org.readthedocs.build/en/1146/api/pypesto.select.postprocessors.html#pypesto.select.postprocessors.Path

dweindl:
> do you think we need to do this at one point for other modules as well?

At some point, yes. Not a priority.

> For some reason `pathlib.Path` is documented as `pypesto.select.postprocessors.Path`

Resolved.

PR #1153, opened 2023-10-25
Fix problem overwrite of profiling

### LINKED ISSUE #1150: `pypesto.profile.parameter_profile` leaves `Problem` in modified state if an error occurs during profiling
What happens:

```python
fixed_parameters_before = problem.x_fixed_indices

parameter_profile(problem, profile_index=[0], ...) # <- fails for unrelated reasons

assert problem.x_fixed_indices == fixed_parameters_before # <- fails, because parameters were fixed during profiling but not unfixed

# retry computing profiles
parameter_profile(problem, profile_index=[0], ...) # <- nothing happens, because that parameter is still set to fixed
```

`parameter_profile` should either create a copy of `problem`, or make sure that it is in the original state when `parameter_profile` exists, independent of any potential errors.

PaulJonasJost:
closes #1150

### DIFF
--- a/pypesto/profile/profile.py
+++ b/pypesto/profile/profile.py
@@ -78,11 +78,11 @@ def parameter_profile(
         The profile results are filled into `result.profile_result`.
     """
     # Copy the problem to avoid side effects
-    problem_profiling = copy.deepcopy(problem)
+    problem = copy.deepcopy(problem)
     # Handling defaults
     # profiling indices
     if profile_index is None:
-        profile_index = problem_profiling.x_free_indices
+        profile_index = problem.x_free_indices
 
     # check profiling options
     if profile_options is None:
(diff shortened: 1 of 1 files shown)


PaulJonasJost:
So deep copy seems to work just fine, but I noticed that we do not have a deep copy function for `Problem` while we have one for `Objective`. Should we add one just in case? I do not think so, as the only thing within problem that's should not be a standard type, is the objective. Other opinions welcome.

pypesto/profile/profile.py
```diff
@@ -76,10 +77,12 @@ def parameter_profile(
     result:
         The profile results are filled into `result.profile_result`.
     """
+    # Copy the problem to avoid side effects
+    problem_profiling = copy.deepcopy(problem)
```
dweindl, line 81:
```suggestion
    problem = copy.deepcopy(problem)
```

Avoids the danger of still modifying the wrong `problem`?

dweindl (approved):
Thanks

PR #1155, opened 2023-10-25
Add warning, if fixed parameter is supposed to be profiled

### LINKED ISSUE #1154: Log a warning in case one tries to profile a fixed parameter.
Currently there is no mentioning of this, just a continue.

Initially talked about in #1150

PaulJonasJost:
closes #1154


dweindl (approved):
Thanks :+1:

PR #1156, opened 2023-10-25
Log traceback in case of exceptions during optimizations

### LINKED ISSUE #1147: Uninformative `'hess'` in `OptimizerResult.message`
Sometimes optimization results have `'hess'` as `OptimizerResult.message` which is not all that helpful. I encountered it with FidesOptimizer, but I can't say it's specific to that.  Not completely sure where it's coming from, but my guess would be https://github.com/ICB-DCM/pyPESTO/blob/883138a63a8911e51884be9dff2028d374ad69f6/pypesto/optimize/optimizer.py#L1340

~~As a start, it would be good to change that line to include the type of exception and stack trace.~~ 

EDIT: No, it's not that line and including the stack traces would currently mess up the messages for non-finite starting points and others.

dweindl:
So far, with `minimize(..., OptimizeOptions(allow_failed_starts=True))`
only the exception message was stored/logged, which is not always sufficiently
informative (see #1147).

Now we log the full exception type / message / traceback.

Closes #1147

### DIFF
--- a/pypesto/optimize/optimizer.py
+++ b/pypesto/optimize/optimizer.py
@@ -1337,11 +1337,8 @@ class FidesOptimizer(Optimizer):
         try:
             opt.minimize(x0)
             msg = self._convert_exitflag_to_message(opt)
-        except RuntimeError:
-            import sys
-            import traceback
-
-            msg = "\n".join(traceback.format_exception(*sys.exc_info()))
+        except RuntimeError as err:
+            msg = str(err)
 
         optimizer_result = OptimizerResult(
             x=opt.x_min,
--- a/pypesto/optimize/optimizer.py
+++ b/pypesto/optimize/optimizer.py
@@ -94,7 +94,12 @@ def history_decorator(minimize):
             )
         except Exception as err:
             if optimize_options.allow_failed_starts:
-                logger.error(f'start {id} failed: {err}')
+                import sys
+                import traceback
+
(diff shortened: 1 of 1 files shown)


PaulJonasJost (approved):
Does this include the whole traceback now? might be quite long, but good to have for this anyways.

Not sure on policies we have on importing, but possibly nicer to have them as default import for the sake of better overview?

dweindl:
> Does this include the whole traceback now? might be quite long, but good to have for this anyways.

Yes. One can always shorten it later. But previously this didn't contain sufficient information to debug anything.

> Not sure on policies we have on importing, but possibly nicer to have them as default import for the sake of better overview?

No clear policy. But if those modules are only required under exceptional circumstances, i would only import them under exceptional circumstances.
I would rather not go back to https://github.com/ICB-DCM/pyPESTO/issues/934

dweindl:
Okay, this is not quite what we want. This messes up handling of cases like `RuntimeError: Encountered non-finite function inf value at initial point.`.

dweindl:
> Okay, this is not quite what we want. This messes up handling of cases like `RuntimeError: Encountered non-finite function inf value at initial point.`.

Changed now. Leaving FidesOptimizer exception reporting as it was, just changing it for the non-caught exceptions. Only logging the full traceback, but keeping the result message short. Should be reasonable compromise.

@PaulJonasJost Please re-review.

PR #1157, opened 2023-10-25
Support valid AMICI noise distributions that are invalid in PEtab

dilpath:
The PEtab importer will throw "invalid PEtab" errors if a user requests e.g. a negative binomial noise distribution. The user can skip the validation check with `PetabImporter(..., validate_petab=False)`. However, AMICI will still complain about the invalid PEtab [1], because `validate_petab` isn't passed to AMICI. This PR fixes this, s.t. users can use any AMICI noise distribution [2].

[1] https://github.com/AMICI-dev/AMICI/blob/3c5e997df3655c26dde35705ef25b2a0f419fe8b/python/sdist/amici/petab_import.py#L565
[2] https://github.com/AMICI-dev/AMICI/blob/3c5e997df3655c26dde35705ef25b2a0f419fe8b/python/sdist/amici/import_utils.py#L182-L188

### DIFF
--- a/pypesto/optimize/optimizer.py
+++ b/pypesto/optimize/optimizer.py
@@ -94,7 +94,12 @@ def history_decorator(minimize):
             )
         except Exception as err:
             if optimize_options.allow_failed_starts:
-                logger.error(f'start {id} failed: {err}')
+                import sys
+                import traceback
+
+                trace = "\n".join(traceback.format_exception(*sys.exc_info()))
+
+                logger.error(f'start {id} failed:\n{trace}')
                 result = OptimizerResult(
                     x0=x0, exitflag=-1, message=str(err), id=id
                 )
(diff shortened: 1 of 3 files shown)


pypesto/petab/importer.py
```diff
@@ -280,7 +281,7 @@ def create_model(
             logger.info(
                 f"Compiling amici model to folder " f"{self.output_folder}."
             )
-            self.compile_model(**kwargs)
+            self.compile_model(validate=self.validate_petab, **kwargs)
```
dweindl, line 284: This fails for pysb models. Pass `validate` only if `from petab.models import MODEL_TYPE_SBML; petab_problem.model.type_id == MODEL_TYPE_SBML`.

dweindl (approved):
:+1:

FFroehlich (approved):

PR #1161, opened 2023-10-30
Fix `HierarchicalAmiciCalculator.__call__` not setting 'hess' in result

### LINKED ISSUE #1160: HierarchicalAmiciCalculator.__call__ missing 'hess' in returned dict upon simulation failure
I finally got to the bottom of #1147 

The problem is that in case of simulation failures  `HierarchicalAmiciCalculator.__call__` does not set `inner_result[HESS]` here, although the Hessian was requested:

https://github.com/ICB-DCM/pyPESTO/blob/883138a63a8911e51884be9dff2028d374ad69f6/pypesto/hierarchical/calculator.py#L113-L122

This results in the following error:

```python

  File "pypesto/optimize/task.py", line 59, in execute
    optimizer_result = self.optimizer.minimize(
                       ^^^^^^^^^^^^^^^^^^^^^^^^
  File "pypesto/optimize/optimizer.py", line 167, in wrapped_minimize
    result = minimize(
             ^^^^^^^^^
  File "pypesto/optimize/optimizer.py", line 136, in wrapped_minimize
    result = minimize(
             ^^^^^^^^^
  File "pypesto/optimize/optimizer.py", line 83, in wrapped_minimize
    result = minimize(
             ^^^^^^^^^
  File "pypesto/optimize/optimizer.py", line 1342, in minimize
    opt.minimize(x0)
  File "fides/minimize.py", line 333, in minimize
    funout_new = self.fevaler(x_new)
                 ^^^^^^^^^^^^^^^^^^^
  File "fides/minimize.py", line 40, in __call__
    ret = self.fun(x, **self.funargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pypesto/objective/amici/amici.py", line 395, in __call__
    return super().__call__(x, sensi_orders, mode, return_dict, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pypesto/objective/base.py", line 197, in __call__
    result = ObjectiveBase.output_to_tuple(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pypesto/objective/base.py", line 334, in output_to_tuple
    output += (kwargs[HESS],)
               ~~~~~~^^^^^^
KeyError: 'hess'
```

dweindl:
`HierarchicalAmiciCalculator.__call__` needs to set `'hess'` in the returned dict, also in case of simulation failures.

Closes #1160

### DIFF
--- a/pypesto/hierarchical/calculator.py
+++ b/pypesto/hierarchical/calculator.py
@@ -15,6 +15,7 @@ from ..C import (
     AMICI_SIGMAY,
     AMICI_Y,
     GRAD,
+    HESS,
     INNER_PARAMETERS,
     INNER_RDATAS,
     RDATAS,
@@ -115,9 +116,12 @@ class HierarchicalAmiciCalculator(AmiciCalculator):
         if any(rdata.status != amici.AMICI_SUCCESS for rdata in inner_rdatas):
             # if the gradient was requested, we need to provide some value
             # for it
+            dim = len(x_ids)
             if 1 in sensi_orders:
-                inner_result[GRAD] = np.full(
-                    shape=len(x_ids), fill_value=np.nan
+                inner_result[GRAD] = np.full(shape=dim, fill_value=np.nan)
+            if 2 in sensi_orders:
+                inner_result[HESS] = np.full(
+                    shape=(dim, dim), fill_value=np.nan
                 )
             return inner_result
 


Doresic (approved):
Yes, I think for simulation failures it should return a nan hessian. I think I encountered a situation where that caused an issue if fides (with BFGS hessian, not analytical) is used with hierarchical optimization. So the changes as they are are good.

However, just to mention: I believe that the true hessian is not the same as the one given by AMICI, because of the hierarchical optimization of inner parameters. The gradient from AMICI is correct as the hierarchical parameters are optimal, so the derivatives cancel out. But I'm not sure that happens for the Hessian. It could even be correct, thinking about it now briefly, since the inner problem optimization is convex so for the optimum both first and second-order derivatives should be 0. But I'm not sure, would have to look at it in detail.

dweindl:
> The gradient from AMICI is correct as the hierarchical parameters are optimal, so the derivatives cancel out. But I'm not sure that happens for the Hessian. It could even be correct, thinking about it now briefly, since the inner problem optimization is convex so for the optimum both first and second-order derivatives should be 0. But I'm not sure, would have to look at it in detail.

Would be great if you could double-check that. If not, we need to let `Objective` know that we can't provide the Hessian instead of happily providing a potentially incorrect one. 

That's independent of this PR - will merge this already.

PR #1163, opened 2023-10-30
ProfileOptions: add some basic integrity checking

dweindl:
I kept wondering why changing `min_step_size` did not change anything. Turned out it was nowhere checked whether `default_step_size` is below `min_step_size`...

Now `ProfileOptions.__init__` and `pypesto.profile.parameter_profile` check at least a subset of profiling options.


PaulJonasJost (approved):
Thanks for the test as well 👍🏼

PR #1166, opened 2023-10-30
Fix pypesto.profile.parameter_profile incorrectly assuming symmetric bounds

### LINKED ISSUE #1165: pypesto.profile.parameter_profile incorrectly assumes symmetric bounds
`pypesto.profile.parameter_profile` incorrectly assumes symmetric bounds (`-lb = ub`) here:

https://github.com/ICB-DCM/pyPESTO/blob/883138a63a8911e51884be9dff2028d374ad69f6/pypesto/profile/walk_along_profile.py#L64-L75

This causes major problems in case this is not fulfilled:
* `ub < -lb`: premature stopping, cropped profiles
* `ub > -lb`: infinite loop

Not sure how this could go unnoticed...

To reproduce:
```python

def test_broken_profiles():
    import pypesto
    from pypesto import profile
    from ..util import rosen_for_sensi

    obj = rosen_for_sensi(max_sensi_order=1)['obj']

    # for infinite loop
    lb = 6 * np.ones(5)
    ub = 10 * np.ones(5)

    # for cropped profile
    lb = -4 * np.ones(5)
    ub = 1 * np.ones(5)

    problem = pypesto.Problem(
        objective=obj,
        lb=lb,
        ub=ub,
    )

    optimizer = optimize.ScipyOptimizer(options={'maxiter': 10})
    result = optimize.minimize(
        problem=problem,
        optimizer=optimizer,
        n_starts=2,
        progress_bar=False,
    )
    profile.parameter_profile(
        problem=problem,
        result=result,
        optimizer=optimizer,
        next_guess_method='fixed_step',
        profile_index=[1],
        progress_bar=False,
        profile_options=profile.ProfileOptions(
            min_step_size=0.5,
            delta_ratio_max=0.05,
            default_step_size=0.5,
            ratio_min=0.01,
            whole_path=True,
        )
    )
    x_path = result.profile_result.list[0][1]['x_path'][1]

    print(x_path.min(), x_path.max())
    assert x_path.max() == ub[1]
    assert x_path.min() == lb[1]
```

dweindl:
Use the correct bounds for checking whether we are done computing the profile.

Add test.

Closes #1165

### DIFF
--- a/pypesto/profile/walk_along_profile.py
+++ b/pypesto/profile/walk_along_profile.py
@@ -62,14 +62,15 @@ def walk_along_profile(
         x_now = current_profile.x_path[:, -1]
 
         # check if the next profile point needs to be computed
-        if options.whole_path:
-            stop_profile = (
-                x_now[i_par] * par_direction >= problem.ub_full[[i_par]]
-            )
+        # ... check bounds
+        if par_direction == -1:
+            stop_profile = x_now[i_par] <= problem.lb_full[[i_par]]
+        elif par_direction == 1:
+            stop_profile = x_now[i_par] >= problem.ub_full[[i_par]]
         else:
-            stop_profile = (
-                x_now[i_par] * par_direction >= problem.ub_full[[i_par]]
-            ) or (current_profile.ratio_path[-1] < options.ratio_min)
+            raise AssertionError("par_direction must be -1 or 1")
+        if not options.whole_path:
+            stop_profile |= current_profile.ratio_path[-1] < options.ratio_min
 
         if stop_profile:
             break
--- a/test/profile/test_profile.py
+++ b/test/profile/test_profile.py
@@ -7,6 +7,7 @@ import warnings
 from copy import deepcopy
 
 import numpy as np
+import pytest
 from numpy.testing import assert_almost_equal
 
 import pypesto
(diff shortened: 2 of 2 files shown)


pypesto/profile/walk_along_profile.py
```diff
@@ -62,14 +62,15 @@ def walk_along_profile(
...
+            stop_profile = x_now[i_par] <= problem.lb_full[[i_par]]
+        elif par_direction == 1:
+            stop_profile = x_now[i_par] >= problem.ub_full[[i_par]]
         else:
-            stop_profile = (
-                x_now[i_par] * par_direction >= problem.ub_full[[i_par]]
-            ) or (current_profile.ratio_path[-1] < options.ratio_min)
+            raise AssertionError("par_direction must be -1 or 1")
+        if not options.whole_path:
+            stop_profile |= current_profile.ratio_path[-1] < options.ratio_min
```
PaulJonasJost, line 73: not entirely sure why not an = but should be fine.

PaulJonasJost (approved):
Thanks for noticing 👍🏼

pypesto/profile/walk_along_profile.py
dweindl: Because it's `stop_due_to_bounds or stop_due_to_ratio`, just `=` would ignore the `stop_due_to_bounds` part.

PR #1167, opened 2023-10-31
Cleanup pypesto/profile/profile_next_guess.py

dweindl:
Some cleanup / simplification of pypesto/profile/profile_next_guess.py

### DIFF
--- a/pypesto/profile/profile_next_guess.py
+++ b/pypesto/profile/profile_next_guess.py
@@ -62,14 +62,15 @@ def next_guess(
         return fixed_step(
             x, par_index, par_direction, profile_options, problem
         )
-    elif update_type == 'adaptive_step_order_0':
+
+    if update_type == 'adaptive_step_order_0':
         order = 0
     elif update_type == 'adaptive_step_order_1':
         order = 1
     elif update_type == 'adaptive_step_regression':
         order = np.nan
     else:
-        raise Exception(
+        raise ValueError(
             f'Unsupported `update_type` {update_type} for `next_guess`.'
         )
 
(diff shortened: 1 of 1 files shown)


pypesto/profile/profile_next_guess.py
```diff
@@ -263,37 +262,19 @@ def par_extrapol(step_length):
...
-            'increase',
-            par_extrapol,
-            next_obj,
-            next_obj_target,
-            clip_to_minmax,
-            clip_to_bounds,
-            par_index,
-            problem,
-            options,
-        )
```
PaulJonasJost, line 296: specific reason to remove this? Wouldn't this now prevent the stepsize to get bigger if the landscape allows for this?

PaulJonasJost (approved):
Thanks for the cleanup. If we remove the `increase` line search, perhaps we should also change the Literal/remove it? But not quite sure whether it is not used somewhere else.

pypesto/profile/profile_next_guess.py
dweindl: Handled by `"decrease" if next_obj_target < next_obj else "increase",`.
Same functionality as before, just getting rid of 12 redundant lines.

pypesto/profile/profile_next_guess.py
```diff
@@ -62,14 +62,15 @@ def next_guess(
         return fixed_step(
             x, par_index, par_direction, profile_options, problem
         )
-    elif update_type == 'adaptive_step_order_0':
```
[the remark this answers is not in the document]
dweindl: Do they really belong together? I'd say no. Fixed step: we are good to go. Adaptive step; we need to set `order`. I think this makes it clearer.

dweindl:
> If we remove the `increase` line search,

It's still there! See other comment.

pypesto/profile/profile_next_guess.py
PaulJonasJost: yes, only saw that now, sorry.

pypesto/profile/profile_next_guess.py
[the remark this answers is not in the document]
PaulJonasJost: one can argue like that as well. No strong preference here 👍🏼

PR #1176, opened 2023-11-16
ESSOptimizer: Check inputs and check stopping criteria more frequently

dweindl:
* Ensure we have some stopping criterion to not run infinitely long
* Avoid exceeding stopping criteria by checking them more frequently

### DIFF
--- a/pypesto/optimize/ess/ess.py
+++ b/pypesto/optimize/ess/ess.py
@@ -70,7 +70,7 @@ class ESSOptimizer:
     def __init__(
         self,
         *,
-        max_iter: int = 10**100,
+        max_iter: int = None,
         dim_refset: int = None,
         local_n1: int = 1,
         local_n2: int = 10,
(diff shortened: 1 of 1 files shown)


pypesto/optimize/ess/ess.py
```diff
@@ -122,6 +122,18 @@ def __init__(
...
+            # in this case, we'd run forever
+            raise ValueError(
+                "Either `max_iter`, `max_eval` or `max_walltime_s` have to be provided."
+            )
+        if max_eval is None:
+            max_eval = np.inf
+        if max_walltime_s is None:
+            max_walltime_s = np.inf
+        if max_iter is None:
+            max_iter = np.inf
```
PaulJonasJost, line 135: Can we have something like sensible default values somewhat? at least for max_iter?

PaulJonasJost (approved):

pypesto/optimize/ess/ess.py
dweindl: What is a sensible default value?

PaulJonasJost: I would not necessarily know, which was why I was asking :D but for general optimisers I think we have like a 1000 steps? 🤔 If you cannot think of one, also fine, would just think that having some kind of sensible default would be nicer than throwing an error.

dweindl: The point here was to have at least one exit criterion. If I want an optimization to run for 1h, and it manages to do more than 1000 iterations then that's fine. I think it doesn't hurt to think about what one wants to have as exit criterion :sweat_smile:.

PaulJonasJost: also fine I guess 😅

PR #1180, opened 2023-11-17
Fix check for allow_failed_starts

dweindl:
Previously it was incorrectly assumed that `optimize_options` is not None. Not it's checked before accessing it.


PaulJonasJost (approved):
Thanks for noticing 👍🏼

PR #1181, opened 2023-11-17
ESSOptimizer: allow passing an optimizer factory

dweindl:
Allow passing a optimizer factory to ESSOptimizer so that overall ESS walltime and function evaluation limits can be respected by the local optimizer.

### DIFF
--- a/pypesto/optimize/ess/ess.py
+++ b/pypesto/optimize/ess/ess.py
@@ -27,7 +27,7 @@ References
 import enum
 import logging
 import time
-from typing import List, Optional, Tuple
+from typing import Callable, List, Optional, Tuple, Union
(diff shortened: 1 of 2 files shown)


PaulJonasJost (approved):
Looks good to me. Should we add checks for the optimisers, whether they support wall time?

dweindl:
> Looks good to me. Should we add checks for the optimisers, whether they support wall time?

I had similar thoughts. For this application I went for passing a callable where the user needs to see whether/how a given optimizer can handle those constraints. That works. If there are more use cases, one could think about adding Optimizer.set_maxtime, Optimizer.set_maxiter, .... to have a common interface for common options. Not sure, though, how much that would be used.

PR #1182, opened 2023-11-17
Refactor ESSOptimizer

dweindl:
Fixes two log messages. The rest is just some cleanup and shuffling some code around for readability / reducing duplications.

### DIFF
--- a/pypesto/optimize/ess/ess.py
+++ b/pypesto/optimize/ess/ess.py
@@ -143,9 +143,7 @@ class ESSOptimizer:
         self.max_iter: int = max_iter
         self.max_eval: int = max_eval
         self.dim_refset: int = dim_refset
-        self.local_optimizer: Optional[
-            'pypesto.optimize.Optimizer'
-        ] = local_optimizer
+        self.local_optimizer = local_optimizer
         self.n_diverse: int = n_diverse
         if n_procs is not None and n_threads is not None:
             raise ValueError(
@@ -204,7 +202,7 @@ class ESSOptimizer:
             refset is not None and problem is not None
         ):
             raise ValueError(
-                "Either `refset` or `problem` has to be provided."
+                "Exactly one of `problem` or `refset` has to be provided."
             )
 
         # generate initial RefSet if not provided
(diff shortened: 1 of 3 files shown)


pypesto/optimize/ess/function_evaluator.py
```diff
@@ -261,3 +261,35 @@ def multiple(self, xs: Sequence[np.ndarray]) -> np.array:
...
+            problem=problem,
+            startpoint_method=startpoint_method,
+            n_procs=n_procs,
+        )
+
+    return FunctionEvaluatorMT(
+        problem=problem,
+        startpoint_method=startpoint_method,
+        n_threads=n_threads or 1,
+    )
```
PaulJonasJost, line 295: So the default is multithread with one thread. Just out of curiosity, why not Multi process with 1?

pypesto/optimize/ess/ess.py
```diff
@@ -189,24 +183,12 @@ def _initialize(self):
...
-        ----------
-        problem:
-            Problem to run ESS on.
-        startpoint_method:
-            Method for choosing starting points.
-            **Deprecated. Use ``problem.startpoint_method`` instead.**
-        refset:
-            The initial RefSet or ``None`` to auto-generate.
-        """
+    ):
```
PaulJonasJost, line 191: A short doctoring would still be nice, even if it is an internal `_foo` function.

PaulJonasJost: or rather a hint that doc is equivalent to `minimize`?

pypesto/optimize/ess/sacess.py
```diff
@@ -328,6 +328,7 @@ def submit_solution(
         sender_idx: Index of the worker submitting the results.
         elapsed_time_s: Elapsed time since the beginning of the sacess run.
         """
+        abs_change = fx - self._best_known_fx.value
```
PaulJonasJost, line 331: Naming a bit misleading as you use `abs` command below. But no idea for better naming, expect perhaps `change`

PaulJonasJost (approved):
Looks good to me.

pypesto/optimize/ess/function_evaluator.py
dweindl: I'd expect it to be more or less the same. The mp part would still create a pool of one, the mt part doesn't do any extra.

pypesto/optimize/ess/sacess.py
dweindl: I think gets relatively clear from the context. `change` would as ambiguous - (absolute) relative or (absolute) absolute?.

PR #1190, opened 2023-11-18
Quiet tests.

dweindl:
Don't print all the test output by default, only on failures.

### DIFF
--- a/test/base/test_prior.py
+++ b/test/base/test_prior.py
@@ -125,7 +125,7 @@ def test_mode(scale, prior_type_list):
 
             # flat functions don't have local minima, so dont check this
             # for uniform priors
-            print("XXXXX", result.optimize_result.list[0].fval)
+
             num_optim = result.optimize_result.list[0]['x'][~np.isnan(topt)]
             assert np.isclose(
                 num_optim, topt[~np.isnan(topt)], atol=1e-03
--- a/test/base/test_prior.py
+++ b/test/base/test_prior.py
@@ -125,7 +125,7 @@ def test_mode(scale, prior_type_list):
 
             # flat functions don't have local minima, so dont check this
             # for uniform priors
-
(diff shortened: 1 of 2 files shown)


test/base/test_prior.py
```diff
@@ -125,7 +125,7 @@ def test_mode(scale, prior_type_list):
 
             # flat functions don't have local minima, so dont check this
             # for uniform priors
-
```
PaulJonasJost, line 128: Is this an actual output we want? especially since it is a print?

tox.ini
```diff
@@ -110,15 +109,15 @@ deps =
...
-        test/hierarchical -s
+        test/hierarchical
 description =
     Test hierarchical optimization module
 
 [testenv:select]
 extras = test,amici,petab,select
 commands =
     pytest --cov=pypesto --cov-report=xml --cov-append \
-        test/select -s
```
PaulJonasJost, line 121: not sure but don't we become "less silent" by setting the capture method from `no` to `fd`?

PaulJonasJost (approved):

test/base/test_prior.py
dweindl: Ooops, definitely not.

test/base/test_prior.py
```diff
@@ -125,7 +125,7 @@ def test_mode(scale, prior_type_list):
...
+            print("XXXXX", result.optimize_result.list[0].fval)
```
dweindl, line 128:
```suggestion
```

tox.ini
dweindl: Compare:
* now: https://github.com/ICB-DCM/pyPESTO/actions/runs/6912941885/job/18809255758
* before: https://github.com/ICB-DCM/pyPESTO/actions/runs/6905429814/job/18788171857

PR #1192, opened 2023-11-19
Doc: Show typehints in description

dweindl:
Instead of in the signature. Much more readable.

E.g., before: https://pypesto.readthedocs.io/en/latest/api/pypesto.visualize.html#pypesto.visualize.waterfall
after: https://pypesto--1192.org.readthedocs.build/en/1192/api/pypesto.visualize.html#pypesto.visualize.waterfall

Also fixes some function documentation missing from sphinx docs due to lack of  `wraps`.

Also removes some circular dependencies.

### DIFF
--- a/pypesto/C.py
+++ b/pypesto/C.py
@@ -5,7 +5,7 @@ Package-wide consistent constant definitions.
 """
 
 from enum import Enum
-from typing import Callable, Literal, Tuple, Union
+from typing import Literal, Tuple, Union
 
 ###############################################################################
 # ENSEMBLE
@@ -274,13 +274,6 @@ CONDITION_IDS = 'condition_ids'
 
 CSV = 'csv'  # return file format
 H5 = 'h5'  # return file format
-
-
-###############################################################################
-# SELECT
-
-TYPE_POSTPROCESSOR = Callable[["ModelProblem"], None]  # noqa: F821
-
 
 ###############################################################################
 # VISUALIZE
--- a/pypesto/predict/task.py
+++ b/pypesto/predict/task.py
@@ -28,7 +28,7 @@ class PredictorTask(Task):
 
     def __init__(
         self,
-        predictor: 'pypesto.predict.Predictor',  # noqa: F821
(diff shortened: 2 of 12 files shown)


pypesto/C.py
```diff
@@ -275,13 +275,6 @@ class InnerParameterType(str, Enum):
 CSV = 'csv'  # return file format
 H5 = 'h5'  # return file format
 
-
-###############################################################################
-# SELECT
-
-TYPE_POSTPROCESSOR = Callable[["ModelProblem"], None]  # noqa: F821
```
dweindl, line 282: @dilpath : Had to shuffle some things around to break circular dependencies. Is that okay?

dilpath: Looks fine to me, thanks :+1:

dilpath (approved):

pypesto/predict/task.py
```diff
@@ -28,7 +28,7 @@ class PredictorTask(Task):
 
     def __init__(
         self,
-        predictor: 'pypesto.predict.Predictor',  # noqa: F821
+        predictor,  #: 'pypesto.predict.Predictor',  # noqa: F821
```
PaulJonasJost, line 31: is this also commented due to circular imports?

PaulJonasJost (approved):

pypesto/predict/task.py
dweindl: No, because it doesn't exist :expressionless: . This whole module is not used anywhere. I think it could/should be removed, but went for the more conservative approach.

dilpath: Ah, it should be `pypesto.predict.AmiciPredictor` I guess -- there is no parent `Predictor` class.

dweindl: Possible, but not sure what was intended there, as it's really not used anywhere.

dweindl: Was introduced in this monster, without usage https://github.com/ICB-DCM/pyPESTO/pull/490/

PR #1194, opened 2023-11-19
SacessOptimizer: More frequent cooperation

dweindl:
Check during every ESS iteration if cooperation/adaptation critera are met.

### DIFF
--- a/pypesto/optimize/ess/refset.py
+++ b/pypesto/optimize/ess/refset.py
@@ -160,6 +160,46 @@ class RefSet:
         """
         Add an attribute array to the refset members.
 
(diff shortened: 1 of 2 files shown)


PaulJonasJost (approved):
I do not exactly see how the cooperation became more frequent, but otherwise looks fine. As far as I recall, the `keep_going` part shortened due to a previous PR?

dweindl:
> I do not exactly see how the cooperation became more frequent, but otherwise looks fine. As far as I recall, the `keep_going` part shortened due to a previous PR?

Previously, a full scatter search was run (incorrectly) instead of only a single iteration. The previous PR was some refactoring to enable performing only one single iteration.

PR #1195, opened 2023-11-19
ESSOptimizer: Retry local search in case of non-finite value at startpoint

dweindl:
In ESSOptimizer, retry local searches if we encounter a non-finite value at the startpoint. Don't add points with non-finite objective values to list of local solutions.

### DIFF
--- a/pypesto/optimize/ess/ess.py
+++ b/pypesto/optimize/ess/ess.py
@@ -461,16 +461,14 @@ class ESSOptimizer:
         """
         if self.local_only_best_sol and self.x_best_has_changed:
             self.logger.debug("Local search only from best point.")
-            local_search_x0 = self.x_best
-            local_search_fx0 = self.fx_best
+            local_search_x0_fx0_candidates = ((self.x_best, self.fx_best),)
         # first local search?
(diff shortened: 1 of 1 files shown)


pypesto/optimize/ess/ess.py
```diff
@@ -492,43 +490,51 @@ def _do_local_search(
             priority = (
                 1 - self.balance
             ) * quality_order + self.balance * diversity_order
-            chosen_child_idx = np.argmin(priority)
-            local_search_x0 = x_best_children[chosen_child_idx]
-            local_search_fx0 = fx_best_children[chosen_child_idx]
+            local_search_x0_fx0_candidates = (
+                (x_best_children[i], fx_best_children[i])
+                for i in np.argsort(priority)
+            )
```
PaulJonasJost, line 496: Why change here from "best" to just all of them but sorted?

PaulJonasJost (approved):

pypesto/optimize/ess/ess.py
dweindl: That same point is tried initially. Previously, we would just use the result from that point, even in non-finite. Now, we try the second-best choice if the best one fails (and so on).

Issue #1179, opened 2023-11-16
test/hierarchical/test_ordinal.py:99: AssertionError


dweindl:
https://github.com/ICB-DCM/pyPESTO/actions/runs/6890727459/job/18744349158

```
______________________ test_optimization[inner_options1] _______________________

inner_options = [{'interval_constraints': 'max-min', 'method': 'standard', 'reparameterized': False}, {'interval_constraints': 'max-mi...reduced', 'reparameterized': False}, {'interval_constraints': 'max-min', 'method': 'reduced', 'reparameterized': True}]

    def test_optimization(inner_options: List[Dict]):
        """Check that optimizations finishes without error."""
        petab_problem = petab.Problem.from_yaml(example_ordinal_yaml)
    
        optimizer = pypesto.optimize.ScipyOptimizer(
            method='L-BFGS-B', options={'maxiter': 10}
        )
        for option in inner_options:
            problem = _create_problem(petab_problem, option)
            result = pypesto.optimize.minimize(
                problem=problem, n_starts=1, optimizer=optimizer
            )
            # Check that optimization finished without infinite or nan values.
            assert np.isfinite(result.optimize_result.list[0]['fval'])
            assert np.all(np.isfinite(result.optimize_result.list[0]['x']))
            assert np.all(np.isfinite(result.optimize_result.list[0]['grad'][2:]))
            # Check that optimization finished with a lower objective value.
>           assert (
                result.optimize_result.list[0]['fval']
                < result.optimize_result.list[0]['fval0']
            )
E           assert 0.09495139837343385 < 0.09495139837343385

test/hierarchical/test_ordinal.py:99: AssertionError
```

@Doresic : Can you please check that? Changing `<` to `<=` would fix it. I am not sure, though, if it makes sense that this test seems to start already at the optimum.

Doresic:
Sure, will take a look. Seems like you're right, it should be a <=, but not sure what changed that it started failing. Will look today and fix.

Doresic:
Put up a PR to solve. https://github.com/ICB-DCM/pyPESTO/pull/1199
Decided to set a random seed so it never happens again due to random drawing. It seems it finally drew a start-point in a minimum.

PR #1199, opened 2023-11-20
 test/hierarchical/test_{ordinal,spline}.py::test_optimization: Set seed for reproducibility

Doresic:
The tests were sometimes failing if the start point chosen was such that it was already in the optimum (or close). Then the end objective function was the same as it was at the start point. Setting a seed so it does not happen again.
Solution to this issue: https://github.com/ICB-DCM/pyPESTO/issues/1179


dweindl (approved):
:+1:

PR #1201, opened 2023-11-20
SacessOptimizer: Fix passing parameters

dweindl:
* Fixes a bug during the cooperation step where best parameters wouldn't be shared correctly due to incorrect assignment to `SyncManager.Array`.
* Fixes a bug where free_indices are ignored, and thus, a wrong parameter vector is stored in local solutions and sent to the manager.

### DIFF
--- a/pypesto/optimize/ess/ess.py
+++ b/pypesto/optimize/ess/ess.py
@@ -528,10 +528,15 @@ class ESSOptimizer:
                 f"{optimizer_result.exitflag}: {optimizer_result.message}"
             )
             if np.isfinite(optimizer_result.fval):
-                self.local_solutions.append(optimizer_result.x)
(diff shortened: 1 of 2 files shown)


pypesto/optimize/ess/ess.py
```diff
@@ -528,10 +528,15 @@ def _do_local_search(
...
+                    optimizer_result.free_indices
+                ]
+                local_solution_fx = optimizer_result.fval
+
+                self.local_solutions.append(local_solution_x)
 
                 self._maybe_update_global_best(
-                    optimizer_result.x, optimizer_result.fval
+                    local_solution_x, local_solution_fx
                 )
```
PaulJonasJost, line 540: Conceptually, why was this changed? what was not working here?

PaulJonasJost (approved):

PR #1202, opened 2023-11-20
SacessOptimizer: Update worker settings

dweindl:
* Most importantly fix default RefSet sizes (too big before)
* Add default settings for >10 workers

### DIFF
--- a/pypesto/optimize/ess/sacess.py
+++ b/pypesto/optimize/ess/sacess.py
@@ -312,7 +312,9 @@ class SacessManager:
             ].copy()
             for setting in ["local_n2", "balance", "dim_refset"]:
                 if setting in leader_options:
-                    self._ess_options[worker_idx] = leader_options[setting]
+                    self._ess_options[worker_idx][setting] = leader_options[
+                        setting
+                    ]
             return self._ess_options[worker_idx].copy()
 
     def submit_solution(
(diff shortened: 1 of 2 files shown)


pypesto/optimize/ess/sacess.py
```diff
@@ -700,90 +701,166 @@ def get_default_ess_options(num_workers: int, dim: int) -> List[Dict]:
...
     Parameters
     ----------
     num_workers: Number of configurations to return.
     dim: Problem dimension.
     """
-    min_dimrefset = 3
+    min_dimrefset = 5
+
+    def dim_refset(x):
+        return max(min_dimrefset, ceil((1 + sqrt(4 * dim * x)) / 2))
```
PaulJonasJost, line 714: Is this also based on [the reference](https://bitbucket.org/DavidPenas/sacess-library/src/508e7ac15579104731cf1f8c3969960c6e72b872/src/method_module_fortran/eSS/parallelscattersearchfunctions.f90#lines-929). Could not find the function `MOD(ids...)`

PaulJonasJost (approved):

pypesto/optimize/ess/sacess.py
dweindl: Yes. The `MOD` is replaced by `cycle` at the end of this function.

PR #1203, opened 2023-11-20
Handle message and exitflag in histories

### LINKED ISSUE #1191: MemoryHistory does not store exitflag and message
Needs to override `HistoryBase.finalize` and store those values.

dweindl:
* Add properties `message` and `exitflag` to `HistoryBase`
* Implement them in derived classes where missing
* Store them in `CountHistoryBase`

Resolves #1191

### DIFF
--- a/pypesto/history/base.py
+++ b/pypesto/history/base.py
@@ -118,6 +118,16 @@ class HistoryBase(ABC):
     def start_time(self) -> float:
         """Return start time."""
 
+    @property
+    @abstractmethod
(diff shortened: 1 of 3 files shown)


m-philipps (approved):

dweindl:
> Should exit flag and message also be listed as possible entries here?

No, those are only the fields we want to save on every iteration / function evaluation.

FFroehlich (approved):

Issue #1077, opened 2023-06-06
Frequent timeouts for CI / base (3.9)


dweindl:
e.g. https://github.com/ICB-DCM/pyPESTO/actions/runs/5187107504/jobs/9349064642?pr=1073

Increase timeouts or speed up tests?

dweindl:
Still happening. Now also for julia tests (e.g. https://github.com/ICB-DCM/pyPESTO/actions/runs/5959628067/job/16166590355)

dweindl:
Also for the mac tests: e.g. https://github.com/ICB-DCM/pyPESTO/actions/runs/6034032186/job/16371678336

PaulJonasJost:
base should be faster in newer version now.

dweindl:
> base should be faster in newer version now.

Times out less frequently, but still does occasionally: https://github.com/ICB-DCM/pyPESTO/actions/runs/6697723646/job/18198209071

dweindl:
While I think it's good to keep the test suite as compact as possible, I would also suggest to just extend the time limits. With re-running all the timed-out tests the current limits waste more time than they save.

PaulJonasJost:
Yes, think that is a good idea. What I also noticed is that the tests seem to run out if too many actions run at the same time. Partially this might be due to two reasons:
* Merge develop into most/all open PRs
* Multiple commits into a single PR as this triggers an action but does not cancel the previous one

dweindl:
Also for notebook tests, e.g. https://github.com/ICB-DCM/pyPESTO/actions/runs/6903646092/job/18782766099

PR #1208, opened 2023-11-21
Speed up test.sample.test_sample.test_pipeline

dweindl:
There is really no need to run `pypesto.visualize.sampling.sampling_1d_marginals` over 20 times.

It seems this was the cause of #1077. In some cases, plotting takes over 10min (cancelled at some point, not sure it would ever complete). Maybe somebody using this function would like to investigate that further...

### DIFF
--- a/test/sample/test_sample.py
+++ b/test/sample/test_sample.py
@@ -2,7 +2,6 @@
 
 import os
 
-import matplotlib.pyplot as plt
 import numpy as np
 import petab
 import pytest
@@ -13,7 +12,6 @@ import pypesto
 import pypesto.optimize as optimize
 import pypesto.petab
 import pypesto.sample as sample
-import pypesto.visualize as visualize
 from pypesto.sample.pymc import PymcSampler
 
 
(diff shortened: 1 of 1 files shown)


dweindl:
For example:

904.16s call     test/sample/test_sample.py::test_pipeline[Emcee-gaussian]

image

dilpath (approved):
Agreed, I don't see a need for any testing of visualization here, it's tested elsewhere.

> 904.16s call test/sample/test_sample.py::test_pipeline[Emcee-gaussian]

I'm surprised this took so long. It should be plotting just 1000 (`n_walkers * n_samples`) samples in total, for a single parameter...

dweindl:
> I'm surprised this took so long. It should be plotting just 1000 (`n_walkers * n_samples`) samples in total, for a single parameter...

Ditto. Might be good to dig deeper. The two times I was able to reproduce long run times were both with emcee. The second time was "only" 210.49s, though. Maybe seaborn / mpl is doing something funny there, not sure.

dilpath:
> > I'm surprised this took so long. It should be plotting just 1000 (`n_walkers * n_samples`) samples in total, for a single parameter...
> 
> Ditto. Might be good to dig deeper. The two times I was able to reproduce long run times were both with emcee. The second time was "only" 210.49s, though. Maybe seaborn / mpl is doing something funny there, not sure.

Looks like it, I can plot a histogram with matplotlib almost immediately, but seaborn takes a lot of time, when using a vector of parameter samples from `test/sample/test_sample.py::test_pipeline[Emcee-gaussian]`. Might be worth switching to matplotlib then.

Regarding the variability, this seems to be dependent on the number of burn-in (Geweke test) samples removed from the visualization.

PR #1211, opened 2023-11-22
Add `Hdf5History.from_history()`

### LINKED ISSUE #1196: Add Hdf5History.from_history()
It would be convenient if one could save, e.g., a MemoryHistory to hdf5 later on.

dweindl:
Adds `Hdf5History.from_history()` which allows saving other histories as HDF5 later on.

Closes #1196


test/base/test_history.py
```diff
@@ -716,3 +717,53 @@ def test_trim_history():
...
+    # write a second time to test `overwrite` argument
+    with pytest.raises(RuntimeError, match="already exists"):
+        pypesto.Hdf5History.from_history(
+            history, hdf5_file, id_="0", overwrite=False
+        )
+    pypesto.Hdf5History.from_history(
+        history, hdf5_file, id_="0", overwrite=True
+    )
+
+    copied = pypesto.Hdf5History(file=hdf5_file, id="0")
```
PaulJonasJost, line 736:
```suggestion
    copied = pypesto.Hdf5History.from_history(
        history, hdf5_file, id_="0", overwrite=True
    )
```

PaulJonasJost (approved):
Looks good to me. Thanks for the test as well.

test/base/test_history.py
dweindl: This was intentional to ensure that the file is closed and read anew. Will add a comment.

PaulJonasJost: Don't you already ensure this in lines 725 together with 729?

dweindl: Not for the overwrite=True case. But might be sufficient.

PR #1212, opened 2023-11-23
 (Sac)ESSOptimizer: Create history of best objective values

### LINKED ISSUE #974: (C)ESS optimizer: Create history for global best
So far, the ESS optimizers don't record any trace/history. It would be good to create/return a trace for the best parameters/fval in each iteration. This would be helpful for performance comparison and would leave some useful result in case optimization is canceled prematurely.

dweindl:
Store traces of the best values/parameters for the `ESSOptimizer` and the best
values of each worker for the `SacessOptimizer`.

Closes #974


pypesto/optimize/ess/ess.py
```diff
@@ -167,6 +171,7 @@ def _initialize(self):
         self.exit_flag: ESSExitFlag = ESSExitFlag.DID_NOT_RUN
         self.evaluator: Optional[FunctionEvaluator] = None
         self.starttime: Optional[float] = None
+        self.history: MemoryHistory = MemoryHistory()
```
PaulJonasJost, line 174: Do I see correctly, that History is now mandatory? Would at least set a flag or something to enable/disable history.

PaulJonasJost (approved):
Some testing would be good I guess, otherwise looks good.

pypesto/optimize/ess/ess.py
dweindl: It is tracked automatically, yes. The current history is usually quite short and I don't see a reason to not collect it. I see that as some basic diagnostic tool with low cost that should always be there.

At some point, there might be some option to track additional values for ensemble purposes which should not be done by default.

PR #1213, opened 2023-11-23
Fix time trajectory simulation for hierarchical solver

vwiela:
Provide additional `simulation_edatas` for simulating and visualizing time trajectories with an hierarchical solver. Before this was not possible since the hierarchical calculator does not allow for changes in the original edatas.


pypesto/hierarchical/calculator.py
```diff
@@ -159,3 +175,31 @@ def __call__(
...
+            experimental condition.
+        """
+
+        if simulation_timepoints is None:
+            self.simulation_edatas = None
+            return
+        else:
+            simulation_edatas = copy.deepcopy(edatas)
+            for i, edata in enumerate(simulation_edatas):
+                edata.setTimepoints(simulation_timepoints[i])
```
dweindl, line 203: Note that just changing the timepoints will still keep the values of the measurements/stddevs which will then be associated with the wrong timepoints.

pypesto/hierarchical/problem.py
```diff
@@ -164,6 +164,48 @@ def check_edatas(self, edatas: List[amici.ExpData]) -> bool:
...
+            A data set. Will be checked against the data set provided to the
+            constructor.
+
+        Returns
+        -------
+        Whether the data sets are consistent.
+        """
+        # TODO: refine check to check for data positions, not only ordering and value of data.
+        original_data = [
+            amici.numpy.ExpDataView(edata)['observedData'][
```
dweindl, line 187:
```suggestion
            amici.numpy.ExpDataView(edata).observedData[
```

pypesto/hierarchical/problem.py
```diff
@@ -164,6 +164,48 @@ def check_edatas(self, edatas: List[amici.ExpData]) -> bool:
 
         return True
 
+    def check_simulation_edatas(
```
dweindl, line 167: What's the intention here? Avoiding that one accidentally passes different datasets?

There would be some members to be checked, see https://amici.readthedocs.io/en/latest/generated/amici.amici.html#amici.amici.ExpData

pypesto/objective/amici/amici.py
```diff
@@ -599,8 +603,13 @@ def set_custom_timepoints(
...
 
-        amici_objective.custom_timepoints = custom_timepoints
-        amici_objective.apply_custom_timepoints()
+        if type(amici_objective.calculator) is HierarchicalAmiciCalculator:
+            amici_objective.calculator.set_simulation_edatas(
+                amici_objective.edatas, custom_timepoints
+            )
+        else:
+            amici_objective.custom_timepoints = custom_timepoints
+            amici_objective.apply_custom_timepoints()
```
dweindl, line 612: I strongly prefer keeping HierarchicalAmiciCalculator out of here. Could you briefly explain the problem here? The ExpDatas are modified here, and then they are not used by the HierarchicalAmiciCalculator when created copy of the objective is used? This should probably be fixed somewhere else.

pypesto/objective/amici/amici.py
```diff
@@ -599,8 +603,13 @@ def set_custom_timepoints(
                 for _ in range(len(amici_objective.edatas))
             ]
 
-        amici_objective.custom_timepoints = custom_timepoints
-        amici_objective.apply_custom_timepoints()
+        if type(amici_objective.calculator) is HierarchicalAmiciCalculator:
```
dweindl, line 606: This will fail on instances of derived classes of `HierarchicalAmiciCalculator` - use `isinstance` instead.

dweindl (commented):
> Provide additional `simulation_edatas` for simulating and visualizing time trajectories with an hierarchical solver. Before this was not possible since the hierarchical calculator does not allow for changes in the original edatas.

While I see the use case, I am not so much in favor of entangling simulation/visualization needs with the AmiciObjective / Amici/Calculator code which is already sufficiently complex. I think this whole topic of simulation of PEtab-defined conditons should rather be handled more conveniently by AMICI.

pypesto/hierarchical/problem.py
vwiela: The idea is to check that the only difference between simulation_edatas and the original edatas is that simulation edatas might consist of more timepoints than the original ones in order to obtain a more dense simulation. I am not sure whether it is necessary to have this check since I would not expect someone to provide simulation edatas by themselves.

pypesto/objective/amici/amici.py
vwiela: In general, I put the modifications into the calculator since this was the best way to change the least amount of existing code and it is only needed for the hierarchical setup (so far). Since for a non-hierarchical setup it is okay to use the `apply_custom_timepoints` function from the AmiciObjective, which will change the edatas. 
But for the hierarchical setup, this would change the calculation of inner parameters and is therefore not allowed so far. 

If one would only want the `apply_custom_timepoints` function to work, it should be enough to modify the `check_edatas` function to allow for NaN values that correspond to just having more dense timepoints. 
I did not do that, since I would not know how to precisely check this and the simulation edatas might be valuable for future modification of simulation/visualization routines.

Doresic: In short, if one wants a copy of an objective with more timepoints that is also hierarchical, the edatas is not allowed to change. But then the simulation edatas (the one passed to AMICI after the optimal scaling+offset are calculated) is changed.

vwiela:
Based on the comments by @dweindl and some others, I changed the fix to be just a change in the code for the visualization routines.
This circumvents changes in the main parts of pypesto and can be undone in the future, when new plotting routines are available. 
Until then, this should fix the issue and one can use the `time_trajectory_model` function with hierarchical models now.

pypesto/visualize/model_fit.py
```diff
@@ -174,24 +176,72 @@ def time_trajectory_model(
...
-    # evaluate objective with return dic = True to get data
+    # get optimization result
     parameters = result.optimize_result.list[start_index]['x']
+
     # reduce vector to only include free indices. Needed downstream.
     parameters = problem.get_reduced_vector(parameters)
-    ret = obj(parameters, mode='mode_fun', sensi_orders=(0,), return_dict=True)
+
+    # simulate with custom timepoints for hierarchical model
+    if isinstance(problem.objective.calculator, HierarchicalAmiciCalculator):
```
m-philipps, line 187: could also outsource everything under this condition to a separate helper function that retrieves the rdatas

m-philipps (approved):
looks fine to me

pypesto/visualize/model_fit.py
```diff
@@ -174,24 +176,72 @@ def time_trajectory_model(
...
+            edatas=edatas,
+            problem_parameters=x_dct,
+            scaled_parameters=True,
+            parameter_mapping=parameter_mapping,
+            amici_model=amici_model,
+        )
+
+        rdatas = amici.runAmiciSimulations(
+            amici_model,
+            amici_solver,
```
dweindl, line 220: Depending on the settings on amici_solver, this might compute sensitivities. Create copy of amici_solver and disable computing sensitivities?

dweindl (approved):
Thanks.

pypesto/visualize/model_fit.py
vwiela: Does the computation of sensitivities create any harm? Or why would it be desirable not to compute them?

dweindl: For non-small problems, the extra computation time is potentially quite annoying.

pypesto/visualize/model_fit.py
vwiela: Liked that idea, so it is also usable for other functions or to the users. Put a function for that into the `misc.py`.

pypesto/visualize/misc.py
```diff
@@ -1,7 +1,9 @@
+import copy
 import warnings
 from numbers import Number
-from typing import Iterable, List, Optional, Union
+from typing import Iterable, List, Optional, Sequence, Union
 
+import amici
```
Doresic, line 6: Tests are probably failing because of this. As much as I understand pyPESTO testing, for some tests, the virtual environment doesn't have amici installed. So if this file is imported, even if the functions using amici are not used, those tests will fail.
To fix this, I think the standard is to include a try: except: for the import, like this:
```
try:
    import amici
except ImportError:
    pass
```

In all hierarchical files, we do this for both amici and petab.

Doresic (approved):
Nice change in the implementation, thanks!

PR #1214, opened 2023-11-27
Scatter search documentation / use local optimizer by default

dweindl:
* Extended scatter search documentation
* Changed `get_default_ess_options` to default to fides for local optimization instead of no local optimizer
* Added a wrapper for fides to forward walltime and function evaluation limits


pypesto/optimize/ess/sacess.py
```diff
@@ -866,7 +901,76 @@ def dim_refset(x):
...
+        elif local_optimizer is not False:
+            cur_settings['local_optimizer'] = local_optimizer
+
     return [
         settings[0],
         *(itertools.islice(itertools.cycle(settings[1:]), num_workers - 1)),
     ]
+
+
+class SacessFidesFactory:
```
PaulJonasJost, line 918: As far as I can see, the factory is a new addition, providing ess with fitting fides optimisers? Is it a lot more effort to generalise this to more than just fides?

PaulJonasJost (approved):
Looks fine to me, thanks for the enhanced documentation :)

PR #1219, opened 2023-11-29
Fix flake8 warning

dweindl:
The recent flake8 fails with `pypesto/objective/amici/amici.py:624:33: B026 Star-arg unpacking after a keyword argument is strongly discouraged, because it only works when the keyword parameter is declared after all parameters supplied by the unpacked sequence, and this change of ordering can surprise and mislead readers.`

Fixed here.


PaulJonasJost (approved):
not pretty but I guess it makes sense

PR #1223, opened 2023-11-29
Set/fix typehints in visualize/profiles.py / visualize/reference_poin…

dweindl:
…ts.py

And remove redundant / contradictory types from docstrings

### DIFF
--- a/pypesto/visualize/profiles.py
+++ b/pypesto/visualize/profiles.py
@@ -1,4 +1,4 @@
-from typing import Sequence, Tuple, Union
+from typing import Optional, Sequence, Union
 from warnings import warn
 
 import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


Doresic (approved):
Nice changes, thank you.
For my information, descriptions like `profile_indices: list of integer values` are not nice, so that's why you're removing and adding them to the function input like `Sequence[int]`?

pypesto/visualize/profiles.py
```diff
@@ -15,50 +15,50 @@ def profiles(
     results: Union[Result, Sequence[Result]],
     ax=None,
     profile_indices: Sequence[int] = None,
-    size: Sequence[float] = (18.5, 6.5),
+    size: tuple[float, float] = (18.5, 6.5),
     reference: Union[ReferencePoint, Sequence[ReferencePoint]] = None,
     colors=None,
```
Doresic, line 20: Should this be `colors: Sequence[tuple]` or something similar?

Doresic (approved):

dweindl:
> For my information, descriptions like `profile_indices: list of integer values` are not nice, so that's why you're removing and adding them to the function input like `Sequence[int]`?

They are not bad per se. However, proper type annotations can be used for static code analysis, those other comments cannot. And redundancy is bad. Also, as you can see in some of the changes, those comments don't reflect the actually expected types.

pypesto/visualize/profiles.py
dweindl: "or something similar" :sweat_smile: 
I am not completely sure which types are supported.

PR #1231, opened 2023-12-01
Julia caching more efficient.

PaulJonasJost:
Make Julia caching more efficient through `Julia-actions`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,11 @@ env:
   # use all available cores for compiling amici models
   AMICI_PARALLEL_COMPILE: ""
 
+# needed to allow julia-actions/cache to delete old caches that it has created
+permissions:
+  actions: write
+  contents: read
+
 # jobs
 jobs:
   base:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -15,6 +15,11 @@ env:
   # use all available cores for compiling amici models
   AMICI_PARALLEL_COMPILE: ""
 
+# needed to allow julia-actions/cache to delete old caches that it has created
+permissions:
+  actions: write
+  contents: read
```
dweindl, line 21: Should be sufficient to add that for the `julia-actions/cache` step?

.github/workflows/ci.yml
```diff
@@ -185,21 +190,15 @@ jobs:
       with:
         version: 1.9
 
-    - name: Cache
-      uses: actions/cache@v3
-      with:
-        path: |
-          ~/.cache
-          .tox/
```
dweindl, line 193: Unless explicitly incompatible with `julia-actions/cache`, continue caching `.tox/` and `~/.cache/pip/`

PaulJonasJost: well Julia_actions/cache does not support the path argument

dweindl: I meant continue caching those with `actions/cache`

dweindl (approved):
:+1:

PR #1241, opened 2023-12-07
Add option for hierarchical models

vwiela:
Add option for hierarchical models in model_selection.

### DIFF
--- a/pypesto/select/misc.py
+++ b/pypesto/select/misc.py
@@ -19,6 +19,7 @@ def model_to_pypesto_problem(
     model: Model,
     objective: Objective = None,
     x_guesses: Iterable[Dict[str, float]] = None,
+    hierarchical: Bool = False
 ) -> Problem:
     """Create a pyPESTO problem from a PEtab Select model.
 
@@ -50,9 +51,13 @@ def model_to_pypesto_problem(
             x_guesses=x_guesses,
             model=model,
             petab_problem=petab_problem,
+            hierarchical=hierarchical
         )
 
-    importer = PetabImporter(petab_problem)
+    importer = PetabImporter(
+        petab_problem, 
+        hierarchical=hierarchical
+    )
     if objective is None:
         amici_model = importer.create_model(
             non_estimated_parameters_as_constants=False,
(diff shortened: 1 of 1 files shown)


pypesto/select/misc.py
```diff
@@ -66,11 +73,47 @@ def model_to_pypesto_problem(
...
+        x_guesses=x_guesses,
+        hierarchical=hierarchical
+    )
+    return pypesto_problem
 
 def correct_x_guesses(
     x_guesses: Iterable[Dict[str, float]],
     model: Model,
     petab_problem: petab.Problem = None,
+    hierarchical: bool = False
```
dilpath, line 116:
```suggestion
    hierarchical: bool = False,
```

dilpath (approved):
Nice, thanks! Let's merge this after you see that it works in your case.

pypesto/select/misc.py
```diff
@@ -66,11 +73,47 @@ def model_to_pypesto_problem(
...
+    Problem
+        The pyPESTO select problem.
+    """
+    pypesto_problem = model_to_pypesto_problem(
+        model=model,
+        objective=objective,
+        x_guesses=x_guesses,
+        hierarchical=hierarchical
+    )
+    return pypesto_problem
```
dilpath, line 110:
```suggestion
def hier_model_to_pypesto_problem(*args, **kwargs) -> Problem:
    """Create a hierarchical pyPESTO problem from a PEtab Select model.
    
    See `model_to_pypesto_problem`.
    """
    pypesto_problem = model_to_pypesto_problem(
        *args,
        **kwargs,
        hierarchical=True,
    )
    return pypesto_problem
```

pypesto/select/misc.py
```diff
@@ -36,6 +38,8 @@ def model_to_pypesto_problem(
         Values in `x_guess` for parameters that are not estimated will be
         ignored and replaced with their value from the PEtab Select model, if
         defined, else their nominal value in the PEtab parameters table.
+    hierarchical:
+        Whether the problem involves hierarchical optimization.
```
dweindl, line 42: How about directly forwarding any `**kwargs` to `PetabImporter()`?

vwiela: Could be done as well. 
I personally found it more transparent in this way, instead of hiding it in `**kwargs`.

dilpath: These methods (`model_to_pypesto_problem` and `model_to_hierarchical_pypesto_problem`) are used as an argument [1].

A separate method was provided so that the user doesn't need to setup with `partial` like
```python
from functools import partial
model_to_pypesto_problem_method = partial(model_to_pypesto_problem, hierarchical=True)
pypesto_select_problem.select_to_completion(..., model_to_pypesto_problem_method=model_to_pypesto_problem_method)
```

I think forwarding `**kwargs` is a good idea, but they will need to be forwarded to two calls: `correct_x_guesses` and `PetabImporter`

[1] https://github.com/ICB-DCM/pyPESTO/blob/b4a187f8ac10ce0cc12a22f49cd75325f3811d8c/pypesto/select/method.py#L250

PR #1242, opened 2023-12-08
Hierarchical log space startpoint sampling

Doresic:
The numerical inner solver of the scaling+offset parameters has a multi-start option. However, the multi-start sampling was done in linear space. In this version, bounds of scaling and offset are not possible. Because of that [-1e20, 1e20] dummy bounds were introduced. But that resulted in the linear sampling having huge values, mostly in the order of 10^19. These startpoints would never converge, rendering the multi-start approach practically useless.

I've implemented sampling in log space for the numerical solver. The values can be negative, so I couldn't directly use pypesto's log sampling. So I used a symmetric logarithm function `np.sign(x) * np.log10(np.abs(x) + 1)` transformation which supports negative values, sampled in that space, and transformed the values back with its inverse `np.sign(x) * (10 ** np.abs(x) - 1)`.

### DIFF
--- a/pypesto/hierarchical/solver.py
+++ b/pypesto/hierarchical/solver.py
@@ -7,6 +7,7 @@ from ..C import InnerParameterType
 from ..objective import Objective
 from ..optimize import minimize
 from ..problem import Problem
+from .parameter import InnerParameter
 from .problem import InnerProblem, scale_value_dict
 from .util import (
     apply_offset,
@@ -231,6 +232,8 @@ class NumericalInnerSolver(InnerSolver):
 
         ub = np.full(shape=len(pars), fill_value=self.dummy_ub)
 
+        x_guesses = self.sample_startpoints(problem, pars)
+
         x_names = [x.inner_parameter_id for x in pars]
         data = problem.data
 
@@ -263,14 +266,9 @@ class NumericalInnerSolver(InnerSolver):
             objective, lb=lb, ub=ub, x_names=x_names, **self.problem_kwargs
         )
 
-        if self.x_guesses is not None:
-            pypesto_problem.set_x_guesses(
-                self.x_guesses[:, pypesto_problem.x_free_indices]
-            )
-        else:
-            pypesto_problem.set_x_guesses(
-                [list(problem.get_dummy_values(scaled=False).values())]
-            )
+        pypesto_problem.set_x_guesses(
+            x_guesses[:, pypesto_problem.x_free_indices]
+        )
 
         # perform the actual optimization
         result = minimize(pypesto_problem, **self.minimize_kwargs)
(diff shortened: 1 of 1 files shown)


pypesto/hierarchical/solver.py
```diff
@@ -298,3 +296,73 @@ def fun(x):
             x_opt = scale_value_dict(x_opt, problem)
 
         return x_opt
+
+    def sample_startpoints(
+        self, problem: InnerProblem, pars: List[InnerParameter]
+    ):
```
dweindl, line 302:
```suggestion
    )  -> np.array:
```

pypesto/hierarchical/solver.py
```diff
@@ -298,3 +296,73 @@ def fun(x):
...
+        Samples the startpoints for the numerical optimization from a
+        log-uniform distribution using the symmetric logarithmic scale.
+
+        Parameters
+        ----------
+        problem:
+            The inner problem to solve.
+        pars:
+            The inner parameters to sample startpoints for.
+        Returns
```
dweindl, line 314:
```suggestion

        Returns
```

dweindl (approved):
Sounds reasonable to me.

pypesto/hierarchical/solver.py
```diff
@@ -298,3 +296,74 @@ def fun(x):
...
+        def inverse_symlog(x):
+            return np.sign(x) * (np.power(10, np.abs(x)) - 1)
+
+        # Sample startpoints from a log-uniform distribution
+        startpoints = np.random.uniform(
+            low=symlog(lb),
+            high=symlog(ub),
+            size=(n_samples, len(pars)),
+        )
+        startpoints = inverse_symlog(startpoints)
```
dilpath, line 350:
```suggestion
        def symlog10(x):
            return np.sign(x) * np.log10(np.abs(x) + 1)

        def inverse_symlog10(x):
            return np.sign(x) * (np.power(10, np.abs(x)) - 1)

        # Sample startpoints from a log-uniform distribution
        startpoints = np.random.uniform(
            low=symlog10(lb),
            high=symlog10(ub),
            size=(n_samples, len(pars)),
        )
        startpoints = inverse_symlog10(startpoints)
```

pypesto/hierarchical/solver.py
```diff
@@ -298,3 +296,74 @@ def fun(x):
...
+
+        if n_samples <= 0:
+            return self.x_guesses
+
+        lb = np.array(
+            [x.lb if x.lb != -np.inf else self.dummy_lb for x in pars]
+        )
+        ub = np.array(
+            [x.ub if x.ub != np.inf else self.dummy_ub for x in pars]
+        )
```
dilpath, line 336: Will replace `nan`s as well (`nan` -> `0`), but I guess the bounds have already been checked to not contain `nan` at this point, since it would be an issue for either implementation.

```suggestion
        lb = np.nan_to_num([x.lb for x in pars], neginf=self.dummy_lb)
        ub = np.nan_to_num([x.ub for x in pars], posinf=self.dummy_ub)
```

pypesto/hierarchical/solver.py
```diff
@@ -298,3 +296,74 @@ def fun(x):
...
+                    self.x_guesses,
+                    startpoints,
+                )
+            )
+        else:
+            startpoints = np.vstack(
+                (
+                    np.array(
+                        [list(problem.get_dummy_values(scaled=False).values())]
+                    ),
```
dilpath, line 365: Why are the dummy values only used in the case where the user doesn't supply `x_guesses`? I think it makes sense to use.

If so, then the highlighted code is currently used at a two points in the code, so it could be computed in advance (doesn't reduce computation, just lines of code).

dilpath (approved):
Looks good! Fine as-is.

pypesto/hierarchical/solver.py
Doresic: Hmmm probably, but it's good to be safe. Nice ty

PR #1243, opened 2023-12-08
delete_nan_inf can handle list input type

### LINKED ISSUE #1107: Parameters plot error when result has `None`
**Bug description**
`visualize.parameters` fails with `ValueError` when some of the optimization starts failed, i.e.  `result.optimize_result[i]['x']` is `None`:

```
---> 17 pypesto.visualize.parameters(
     18     results=result, 
     19     start_indices="all", 
     20     reference=reference,
     21 )

File [~/Documents/pyPESTO/pypesto/visualize/parameters.py:121], in parameters(results, ax, parameter_indices, lb, ub, size, reference, colors, legends, balance_alpha, start_indices, scale_to_interval, plot_inner_parameters)
    118     lb, ub, xs = map(scale_parameters, (lb, ub, xs))
    120     # call lowlevel routine
--> 121     ax = parameters_lowlevel(
    122         xs=xs,
    123         fvals=fvals,
    124         lb=lb,
    125         ub=ub,
    126         x_labels=x_labels,
    127         ax=ax,
    128         size=size,
    129         colors=colors[j],
    130         legend_text=legends[j],
    131         balance_alpha=balance_alpha,
    132     )
    134 # parse and apply plotting options
    135 ref = create_references(references=reference)

File [~/Documents/pyPESTO/pypesto/visualize/parameters.py:272], in parameters_lowlevel(xs, fvals, lb, ub, x_labels, ax, size, colors, linestyle, legend_text, balance_alpha)
    238 """
    239 Plot parameters plot using list of parameters.
    240 
   (...)
    269     The plot axes.
    270 """
    271 # parse input
--> 272 xs = np.array(xs)
    273 fvals = np.array(fvals)
    274 # remove nan or inf values in fvals and xs

ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (100,) + inhomogeneous part.
```

**Expected behavior**
I would expect the parameters plot to be generated without the failed starts

**Comment**
In the lower level [process_start_indices](https://github.com/ICB-DCM/pyPESTO/blob/develop/pypesto/visualize/misc.py#L290) `delete_nan_inf` is only used when `start_indices` is set to `'all_clustered'` or `'first_cluster'`

m-philipps:
- fix #1107 
- do not attempt to convert parameter list with `None` to numpy array
- filter out `None` from parameter vector within `delete_nan_inf`
- add a test that covers optimizer result with `x=None`


pypesto/visualize/optimization_stats.py
```diff
@@ -375,7 +375,7 @@ def stats_lowlevel(
         The plot axes.
     """
     fvals = result.optimize_result.fval
-    values = [res[property_name] for res in result.optimize_result.list]
+    values = [[res[property_name]] for res in result.optimize_result.list]
```
PaulJonasJost, line 378: Why is making that specifically a list within a list required?

PaulJonasJost (approved):

pypesto/visualize/optimization_stats.py
m-philipps: This is making the use of `delete_nan_inf` consistent. 
The values are given over to `delete_nan_inf` in place of the optimization result parameters. These are parameter vectors within a list, i.e. one list (or `None`) per start.

m-philipps:
Does anyone know where this condition is needed? It seems counter-intuitive because the function deletes nan values from the given list.
https://github.com/ICB-DCM/pyPESTO/blob/c012c19faad9ccd53de3711dd7f22e40d2dd9c4c/pypesto/util.py#L282-L296

dweindl:
> Does anyone know where this condition is needed? It seems counter-intuitive because the function deletes nan values from the given list.

No idea where it's needed. Yes, it's deletes the nans, but it makes sure that the dimensions of `x` still match (an np.array with shape (0, 0) is different from (0, 2), although they are both empty).

pypesto/visualize/parameters.py
```diff
@@ -121,6 +121,17 @@ def scale_parameters(x):
...
+        # parse fvals and parameters
+        fvals = np.array(fvals)
+        # remove nan or inf values
+        xs, fvals = delete_nan_inf(
+            fvals=fvals,
+            x=xs,
+            xdim=len(ub) if ub is not None else 1,
+            magnitude_bound=WATERFALL_MAX_VALUE,
+        )
+
```
Doresic, line 134: Moving the `delete_nan_inf` to here makes it so the plotting of reference points doesn't go through it. Is that an issue? I've never used reference points, but they also possibly might contain infinite/nan values, no?

m-philipps: Good point, it's not a problem though, nans are just not shown in the line plot.

stephanmg (approved):
Looks fine, branch needs update.

Doresic (commented):
Looks good! Thank you

PR #1244, opened 2023-12-11
Fix hierarchical parameter plotting for all optimizers

Doresic:
The hierarchical parameters were being plotted in the parameter plot only if a SciPy optimizer was used. Fixed this by making all other optimizers save the inner parameters in the result object as well.

A second issue was that the pypesto result couldn't be stored in the hdf5 format because of saving the inner parameter as a dictionary in the `OptimizerResult`. Now it's saved as a list of values instead. The `inner_x_names` are obtained via the inner problem.


Doresic:
@m-philipps I made the hierarchical part in `handle_inputs()` of the parameter plotting much shorter. It can still be put into a separate function if it is still too long or would. A name could be `handle_inner_inputs()`.

Doresic:
I tested some optimizers: L-BFGS-B, SLSQP, fides, pyswarm, powell. For all of those, the parameter plot works nicely.
Had some trouble installing ipopt.

pypesto/optimize/optimizer.py
```diff
@@ -41,6 +41,16 @@ def __init__(self, optimizer: str):
...
 
 
+def add_inner_parameters(
+    objective: Objective, optimizer_result: OptimizerResult
+):
+    """Add inner parameters from objective to the optimizer result."""
+    if hasattr(objective, INNER_PARAMETERS) and any(
+        objective.inner_parameters
+    ):
+        optimizer_result[INNER_PARAMETERS] = objective.inner_parameters
```
dweindl, line 51: I don't really like the idea of the `Optimizer` being aware of hierarchical optimization. In my opinion, this should all be hidden inside the given `Problem`.
Instead of having an optimizer call `add_inner_parameters`, I'd prefer the optimizer calling sth like `Problem.postprocess_optimization_result`. That's also not great - open for better suggestions - because `Problem` shouldn't have to care about whether it's being optimizer, sampled, or whatever, but I'd still prefer it that way around.

test/visualize/test_visualize.py
```diff
@@ -457,15 +459,26 @@ def test_parameters_hist():
 @pytest.mark.parametrize("scale_to_interval", [None, (0, 1)])
 @close_fig
 def test_parameters_hierarchical(scale_to_interval):
-    # create the necessary results
-    result = create_optimization_result()
+    # bbtain a petab problem with hierarchical parameters
```
dweindl, line 462:
```suggestion
    # obtain a petab problem with hierarchical parameters
```

dweindl (commented):
Thanks for fixing that. I'd prefer some way that doesn't require every single optimizer to take care of it itself. Related to the other comment. 
How about renaming the current `xxxOptimizer.minimize` functions to `xxxOptimizer._minimize`, and changing `Optimizer.minimize` to ` def minimize(...): res = self._minimize(...); return self.postprocess_result(...)`?
Would require a few more changes here, but would be more future-proof.

pypesto/optimize/optimizer.py
Doresic: Ohhh I like the final comment idea you had. Renaming the minimize function of each individual optimizer `_minimize`, and the minimize call being called from the base (parent) object which would postprocess it as well. Definitely sounds nicer, and more consistent with respect to the current objective implementation (unprocessed call etc...). Would like to hear the opinions of others on what's best to do.

FFroehlich (approved):
only looked at `pypesto/objective/amici/amici.py`

pypesto/optimize/optimizer.py
dilpath: One alternative is to implement this as a wrapper, like `history_decorator`. I agree with Daniel, it's not clear to me whether it's better for `Optimizer` to be aware of `inner_parameters`, or for `Problem` to be aware of `OptimizerResult`. Maybe "`Optimizer` knows `inner_parameters`" is less circular, since the `inner_parameters` are actually output from the objective, and `Optimizer` already knows `problem.objective`.

Somewhat related: since there are so many decorators applied to `minimize`, maybe a single decorator that collects all other decorators will make the code cleaner.

> Renaming the minimize function of each individual optimizer _minimize, and the minimize call being called from the base (parent) object which would postprocess it as well.

This sounds like an alternative to the currect decorators, but if the current decorators are replaced with this `_minimize`+`minimize` setup, then you would require a `preprocess` step too -- or, will you keep the current decorators, and just have this setup only for inner parameters? I think sticking to decorators is probably the cleanest option.

Does the current code also allow users to save inner parameters from profiling? If not, is this planned, and how would it be done?

pypesto/visualize/parameters.py
```diff
@@ -386,41 +380,30 @@ def handle_inputs(
     inner_xs = [
         res.get(INNER_PARAMETERS, None) for res in result.optimize_result.list
     ]
-    if any(inner_xs):
-        inner_xs_names = next(
-            list(inner_x.keys()) for inner_x in inner_xs if inner_x is not None
+
+    from ..hierarchical.calculator import HierarchicalAmiciCalculator
```
dilpath, line 384: Does this break the parameters plot for users without AMICI? Looks like imports are handled appropriately with `try-except` blocks in the different scripts this import depends on, but it would be important to ensure.

dilpath (approved):
Looks cleaner :+1:

pypesto/optimize/optimizer.py
dweindl: Decorators: As good.

> ...  or for `Problem` to be aware of `OptimizerResult`

I think at some point we might need something in that direction (e.g. plans for some PEtab result file). Maybe we can get around knowing about `OptimizerResult` specifically by making all `*Result` handle some kind of `ProblemInternals`.
This is probably out of scope here. I would be happy with a postprocessing solution in `Optimizer` (based on decorators or methods), so we have that handled in a single location. That should make pushing things from Optimizer to Problem much easier later on.

pypesto/optimize/optimizer.py
```diff
@@ -41,6 +41,16 @@ def __init__(self, optimizer: str):
         )
 
 
+def add_inner_parameters(
```
dweindl, line 44:
```suggestion
def _add_inner_parameters(
```

Not meant to be used elsewhere, right?

doc/example/hierarchical.ipynb
```diff
@@ -87,10 +89,20 @@
...
    "metadata": {},
    "source": [
-    "The PEtab observable table contains placeholders for scaling parameters $s$ (`observableParameter1_{pSTAT5A_rel,pSTAT5B_rel,rSTAT5A_rel}`), offsets $b$ (`observableParameter2_{pSTAT5A_rel,pSTAT5B_rel,rSTAT5A_rel}`), and noise parameters $\\sigma^2$ (`noiseParameter1_{pSTAT5A_rel,pSTAT5B_rel,rSTAT5A_rel}`) that are overridden by the `{observable,noise}Parameters` column in the measurement table. When using hierarchical optimization, the nine overriding parameters `{offset,scaling,sd}_{pSTAT5A_rel,pSTAT5B_rel,rSTAT5A_rel}` are to be estimated in the inner problem."
+    "To convert a non-hierarchical PEtab model to a hierarchical one, the observable_df, the measurement_df and the parameter_df have to be changed accordingly.\n",
+    "The PEtab **observable table contains** placeholders for scaling parameters $s$ (`observableParameter1_{pSTAT5A_rel,pSTAT5B_rel,rSTAT5A_rel}`), offsets $b$ (`observableParameter2_{pSTAT5A_rel,pSTAT5B_rel,rSTAT5A_rel}`), and noise parameters $\\sigma^2$ (`noiseParameter1_{pSTAT5A_rel,pSTAT5B_rel,rSTAT5A_rel}`) that are overridden by the `{observable,noise}Parameters` column in the **measurement table**."
```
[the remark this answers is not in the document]
Doresic: They can even be timepoints specific. The only requirements for the scaling+offset thing is that they are only coupled to one other scaling/offset parameter and the sigma parameters can be coarser but each scaling/offset can be connected to only one sigma.

I've never tried to put the inner parameters in the observable formulae directly. Did you? If it works, it's definitely good to include it in the notebook.

pypesto/objective/amici/amici.py
```diff
@@ -456,7 +456,7 @@ def call_unprocessed(
 
         nllh = ret[FVAL]
         rdatas = ret[RDATAS]
-        if INNER_PARAMETERS in ret and ret[INNER_PARAMETERS]:
+        if INNER_PARAMETERS in ret and any(ret[INNER_PARAMETERS]):
             self.inner_parameters = ret[INNER_PARAMETERS]
```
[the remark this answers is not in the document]
Doresic: Hmmm this was a mistake. It should be `ret[INNER_PARMETERS] is not None`. It can be `None` if the simulation fails, or if the `inner_calculator_collector` doesn't have any interpretable inner parameters (basically noise, spline or ordinal parameters are not easily interpretable by just number).

pypesto/visualize/parameters.py
Doresic: Hmmm True. I also don't like how it's just there. It should be separated: in the sense that it should be imported only if there are non-None `inner_xs`.

Doresic: I changed it, and separated the inner_xs part into a separate function `handle_inner_inputs`.

doc/example/hierarchical.ipynb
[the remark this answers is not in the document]
dilpath: Thanks, makes sense, timepoint-specific is required for ordinal data.

> I've never tried to put the inner parameters in the observable formulae directly. Did you?
I'm not sure I've tried it, but the format validation in `hierarchical/petab.py` is setup to support it I think...

pypesto/optimize/optimizer.py
Doresic: Seems like the cleanest thing now to do would be to add it as a decorator and collapse all decorators into one. I'll try that now. If it doesn't fit well, I can change it to the `_minimize` + `minimize` option.

pypesto/objective/amici/amici.py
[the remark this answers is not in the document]
dilpath: Ah OK, then can be `if ret.get(INNER_PARAMETERS, None) is not None:`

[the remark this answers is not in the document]
Doresic: Ohh that's even nicer :D 
Will make it into that.

pypesto/visualize/parameters.py
dilpath: Looks nice! Thanks

pypesto/visualize/parameters.py
```diff
@@ -382,45 +376,10 @@ def handle_inputs(
     # retrieve results
     fvals = result.optimize_result.fval
     xs = result.optimize_result.x
+
     # retrieve inner parameters if available
```
m-philipps, line 380:
```python
# retrieve inner parameters in case of hierarchical optimisation
```

pypesto/visualize/parameters.py
```diff
@@ -471,6 +430,58 @@ def handle_inputs(
     return lb, ub, x_labels, fvals_out, xs_out
 
 
+def _handle_inner_inputs(
+    result: Result,
+):
```
m-philipps, line 435: return type would be

```python
(Union[None, list(float])
```
for each of inner_xs, inner_xs_names, inner_lb, inner_ub?

m-philipps (approved):
thanks :+1:

pypesto/visualize/parameters.py
Doresic: I think it's 
```
Union[
    Tuple[None, None, None, None],
    Tuple[list[np.ndarray], list[str], np.ndarray, np.ndarray],
    ]
```
Since it either returns 4 None's or 4 things

pypesto/optimize/optimizer.py
Doresic: Done, seems to be working nicely :)

PaulJonasJost (approved):
Did not look at the hierarchical parts. Looks good to me. Potential thought for the future (not needed here) is to allow a selection of wrappers.

stephanmg (approved):

pypesto/objective/amici/amici.py
```diff
@@ -215,7 +215,7 @@ def __init__(
         self.custom_timepoints = None
 
         # Initialize the dictionary for saving of inner parameters.
```
dweindl, line 217:
```suggestion
```

dweindl (approved):
:+1:

PR #1250, opened 2023-12-15
Vis: Sacess history plot

dweindl:
Added `pypesto.visualize.optimizer_history.sacess_history` to visualize SacessOptimizer histories. In principle, they could be plotted the same as any other history, but here a step plot is usually preferable, and we show the global best in addition to all workers.


stephanmg:
Thanks for the enhancements. LGTM and can be merged.

stephanmg (approved):

m-philipps (approved):
Thanks :+1:

Issue #1252, opened 2023-12-16
Importing pypesto hides all warnings


dweindl:
```sh
$ python -c "import warnings; warnings.warn('bla');"
<string>:1: UserWarning: bla
$ python -c "import pypesto; import warnings; warnings.warn('bla');"
$
```

This can't be intentional. Not sure yet whether it's pypesto or any of its dependencies.

dweindl:
Turns out it's AMICI.

PaulJonasJost:
as it is amici, will be closing it here.

PR #1253, opened 2023-12-16
Use warnings.warn instead of logging.warn when loading Problem from H…

dweindl:
…DF5 without an Objective

Allows using warnings filters to get rid of this message.

e.g.,

```python
import warnings
warnings.filterwarnings("ignore", category=UserWarning, message=".*You are loading a problem.*", module="pypesto.store.read_from_hdf5")
```

**Wait until next amici release for #1252 to be fixed.**


pypesto/store/read_from_hdf5.py
```diff
@@ -120,10 +121,9 @@ def read(self, objective: ObjectiveBase = None) -> Problem:
         if objective is None:
             objective = Objective()
             # raise warning that objective is not loaded.
-            logger.info(
-                'WARNING: You are loading a problem.\nThis problem'
-                ' is not to be used without a separately created'
-                ' objective.'
+            warnings.warn(
```
PaulJonasJost, line 124: why not logger.warn? 🤔

PaulJonasJost (approved):

pypesto/store/read_from_hdf5.py
dweindl: Because then I can't easily silence just this warning

PR #1254, opened 2023-12-16
Fix test.util.load_amici_objective / AmiciVersionError

dweindl:
So far, after updating amici, test.util.load_amici_objective would fail with AmiciVersionError. In that case, we just want to re-import the model. That's what's achieved here.


PaulJonasJost (approved):
That is gold, thanks a lot 👍🏼

PR #1260, opened 2023-12-19
GHA: Clear PR cache after merge

dweindl:
Might resolve frequent cache misses for Julia tests.


Doresic (approved):
Can't comment much as I don't know much about caching. But looks good! :D 
Small typo

PR #1261, opened 2023-12-19
GHA: Reduce cache size

dweindl:
We have limit cache size. Only cache what's really relevant.


Doresic (approved):
What does ~/.cache/pip change? The tests cache somewhere else? Or cache less? (for my info)

dweindl:
> What does ~/.cache/pip change? The tests cache somewhere else? Or cache less? (for my info)

It's just the directory to be cached. Previously is was all under ~/.cache, now it's only ~/.cache/pip.

PR #1263, opened 2023-12-21
Extended Amici history

plakrisenko:


pypesto/history/amici.py
```diff
@@ -0,0 +1,232 @@
...
+        self,
+        id: str,
+        file: Union[str, Path],
+        options: Union[HistoryOptions, dict, None] = None,
+    ):
+        super().__init__(id, file, options=options)
+
+    @staticmethod
+    def _simulation_to_values(x, result, used_time):
+        values = Hdf5History._simulation_to_values(x, result, used_time)
```
dilpath, line 46: Here it's `@staticmethod`, but n `CsvAmiciHistory` it's `super()...`. Just curious, why?

pypesto/history/amici.py
```diff
@@ -0,0 +1,232 @@
...
+        return columns + [
+            (c, np.nan)
+            for c in [
+                CPU_TIME_TOTAL,
+                PREEQ_CPU_TIME,
+                PREEQ_CPU_TIME_B,
+                POSTEQ_CPU_TIME,
+                POSTEQ_CPU_TIME_B
+            ]
+        ]
```
dilpath, line 157: Should this also exist for `Hdf5AmiciHistory`?

pypesto/history/csv.py
```diff
@@ -73,7 +73,7 @@ def __init__(
                 # transform strings to np.ndarrays
                 trace[col] = trace[col].apply(string2ndarray)
 
-            self._trace = trace
+            self._trtraceace = trace
```
dilpath, line 76: :thinking:

dilpath (commented):
Looks good :+1: I'm not familiar with the differences between the implementations for CSV and HDF5 history, so some of my feedback might be trivial.

pypesto/history/csv.py
plakrisenko: whoops :open_mouth:

pypesto/history/amici.py
plakrisenko: This is for `_init_trace which is only needed in Csv case, I think

pypesto/history/amici.py
plakrisenko: In `CsvHistory` it's not static as it uses dynamically updated attributes (e.g. `_n_fval`), in `Hdf5History` it's not the case, so I made the methods static both in `Hdf5History` and `Hdf5AmiciHistory`.

pypesto/history/base.py
```diff
@@ -249,7 +249,7 @@ def get_time_trace(
         trim: bool = False,
     ) -> Union[Sequence[float], float]:
         """
-        Cumulative execution times.
+        Cumulative execution times [s].
```
dweindl, line 252: Since pypesto already reports these in seconds, wouldn't it be better to convert the newly added quantities to seconds instead of ms?

m-philipps (approved):

pypesto/optimize/optimizer.py
```diff
@@ -106,6 +106,7 @@ def wrapped_minimize(
             id=id,
             x_names=[problem.x_names[ix] for ix in problem.x_free_indices],
             options=history_options,
+            amici_objective=isinstance(objective, AmiciObjective)
```
dweindl: I'd strongly prefer to keep AMICI out of pypesto.{optimize,history}.
How about something like:
```python
from pypesto.history.generate import create_history
class ObjectiveBase:
   def create_history(...):
       create_history(...)
```
and doing the amici-specific magic in `AmiciObjective.create_history`?

plakrisenko: Maybe break create_history in two parts: without storage file and with storage file, and implement both in ObjectiveBase (then pypesto.history.generate isn't needed)? Then these can be re-implemented in Objectives as needed.

dweindl: > then pypesto.history.generate isn't needed

That could/should be done in the medium term. For now, I'd just leave it there and re-use in ObjectiveBase.create_history, just in case somebody is actually using pypesto.history.generate.create_history directly.

test/base/test_history.py
```diff
@@ -689,6 +691,57 @@ def test_hdf5_history_mp():
...
+                progress_bar=False,
+            )
+            assert isinstance(
+                result2.optimize_result.list[0].history, amici_history_class
+            )
+            result2.optimize_result.list[0].history.get_cpu_time_total_trace()
+            result2.optimize_result.list[0].history.get_preeq_time_trace()
+            result2.optimize_result.list[0].history.get_preeq_timeB_trace()
+            result2.optimize_result.list[0].history.get_posteq_time_trace()
+            result2.optimize_result.list[0].history.get_posteq_timeB_trace()
```
dweindl, line 742:
```suggestion
            history = result2.optimize_result.list[0].history
            assert isinstance(history, amici_history_class)

            assert np.all(
                history.get_cpu_time_total_trace()
                >= history.get_preeq_time_trace()
            )
            assert np.all(
                history.get_cpu_time_total_trace()
                >= history.get_preeq_timeB_trace()
            )
            assert np.all(
                history.get_cpu_time_total_trace()
                >= history.get_posteq_time_trace()
            )
            assert np.all(
                history.get_cpu_time_total_trace()
                >= history.get_posteq_timeB_trace()
            )
```

test/base/test_history.py
```diff
@@ -689,6 +691,57 @@ def test_hdf5_history_mp():
...
+                n_starts=1,
+                history_options=history_options_mp,
+                progress_bar=False,
+            )
+            assert not isinstance(
+                result1.optimize_result.list[0].history, amici_history_class
+            )
+
+            # optimizing with amici history saved in hdf5
+            result2 = pypesto.optimize.minimize(
```
dweindl, line 728: `ValueError: ID "0" is already used in history file`

-> delete file before second minimize or use a different file

pypesto/optimize/optimizer.py
```diff
@@ -18,7 +18,7 @@
     OptimizerHistory,
     create_history,
 )
-from ..objective import Objective
+from ..objective import AmiciObjective, Objective
```
dweindl, line 21:
```suggestion
from ..objective import Objective
```

dweindl (approved):
Thx

pypesto/history/amici.py
```diff
@@ -0,0 +1,260 @@
...
+        file: Union[str, Path],
+        options: Union[HistoryOptions, dict, None] = None,
+    ):
+        super().__init__(id, file, options=options)
+
+    @staticmethod
+    def _simulation_to_values(x, result, used_time):
+        values = Hdf5History._simulation_to_values(x, result, used_time)
+        # default unit for time in amici is [ms], converted to [s]
+        values |= {
```
FFroehlich, line 49:
```suggestion
        values |= {
            key: sum(
                 [rdata[key] for rdata in result[RDATAS]]
             ) * 0.001
             for key in (CPU_TIME_TOTAL, PREEQ_CPU_TIME, PREEQ_CPU_TIME_BACKWARD, POSTEQ_CPU_TIME, POSTEQ_CPU_TIME_BACKWARD)
         }
```

pypesto/history/amici.py
```diff
@@ -0,0 +1,260 @@
...
+                PREEQ_CPU_TIME_BACKWARD,
+                POSTEQ_CPU_TIME,
+                POSTEQ_CPU_TIME_BACKWARD,
+            ]
+        ]
+
+    def _simulation_to_values(self, result, used_time):
+        values = super()._simulation_to_values(result, used_time)
+        # default unit for time in amici is [ms], converted to [s]
+        values |= {
```
FFroehlich, line 178: see comment on `Hdf5AmiciHistory`

FFroehlich (approved):

test/base/test_history.py
```diff
@@ -689,6 +691,70 @@ def test_hdf5_history_mp():
...
+                >= history.get_preeq_timeB_trace()
+            )
+            assert np.all(
+                history.get_cpu_time_total_trace()
+                >= history.get_posteq_time_trace()
+            )
+            assert np.all(
+                history.get_cpu_time_total_trace()
+                >= history.get_posteq_timeB_trace()
+            )
```
PaulJonasJost, line 755: Wouldn't a better assertion actually be to assert the sum over the different traces to be smaller equal than the total time?

PaulJonasJost (approved):
Thanks :)

PR #1246, opened 2023-12-13
Execute all notebooks / clear notebook output

Doresic:
All of the example notebooks are stored on GitHub with their output generated which is a lot of figures and text, taking up storage unnecessarily. I've removed the output, but I still need to check the notebook output is being generated for the online docs.

PR #1269, opened 2024-01-05
petab: Enable Importer passing verbose to create_model

Doresic:
Enable the `create_objective` and `create_problem` methods of the `PetabImporter` to pass verbose to `create_model` and `sbml2amici`. Coming from #1246

### DIFF
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -252,6 +252,7 @@ class PetabImporter(AmiciObjectBuilder):
     def create_model(
         self,
         force_compile: bool = False,
+        verbose: bool = True,
         **kwargs,
     ) -> amici.Model:
         """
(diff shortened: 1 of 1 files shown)


pypesto/petab/importer.py
```diff
@@ -428,7 +428,10 @@ def create_objective(
 
         # create model
         if model is None:
-            model = self.create_model(force_compile=force_compile)
+            verbose = kwargs.pop('verbose', True)
```
dweindl, line 431: It's good to be able to pass `verbose=False`. However, it's potentially confusing to do that via `kwargs` (`**kwargs:
            Additional arguments passed on to the objective.`). I'd prefer adding (and documenting) it as a separate keyword argument to `create_objective`.

Doresic: True, will change.

Doresic: Additionally, I extended the same thing to all methods of the `PetabImporter` (create_edatas, create_solver etc...). To be uniform in changes, but it might be too much?

dweindl (approved):
Thx

pypesto/petab/importer.py
dweindl: > To be uniform in changes, but it might be too much?

No strong opinion. I don't think it will be used frequently via those other functions, but I don't think it hurts having it there.

Doresic: Same feeling

PR #1270, opened 2024-01-05
ESSOptimizer: Include results for local searches in OptimizeResult

dweindl:
Include results for local searches in OptimizeResult. Those might be useful for creating ensembles or waterfall plots.


PaulJonasJost (approved):
looks good to me, thanks :)

PR #1271, opened 2024-01-08
Doc: Clear petab notebook output

dweindl:
A subset of #1246.

:eye: https://pypesto--1271.org.readthedocs.build/en/1271/example/petab_import.html

PR #1272, opened 2024-01-08
Refactor progress bars

dweindl:
tqdm progress bars are used in a couple of places. They don't play so well in non-interactive jobs, testing, ... . They also cause trouble with nbspinx (#1246, #1271).
Progress bars can be disabled for specific tasks, but not globally (or at least not very conveniently).
Since recently, tqdm can be controlled via environment variables (e.g., disabling all progress bars or changing update frequency). However, this works by changing the argument defaults, so it only works if we don't pass explicit `disable=...`. Therefore, this PR introduces some wrapper that checks whether the user explicitly enabled/disabled progress bars. If not, we go with the tqdm default, which means showing all progress bars unless globally disabled. An additional `enabled` argument is added for convenience.


dilpath (approved):
:+1:

vwiela (approved):
Looks good to me.

PR #1275, opened 2024-01-09
Doc: Clear amici and conversion reaction notebook output

dweindl:
A subset of #1246.

After #1272.


PaulJonasJost (approved):
Could not inspect it well, as it is a fork, but first few code spaces seem fine.

PR #1277, opened 2024-01-09
Doc: Use clang on RTD

dweindl:
... and parallelize model import for faster notebook execution.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -21,7 +21,7 @@ python:
 build:
   os: "ubuntu-22.04"
   apt_packages:
-    - gfortran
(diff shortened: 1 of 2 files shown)


Doresic (approved):
Clang is faster than gfortran? Or are you changing it for another reason? 
Nice change for parallel compile.

dweindl:
> Clang is faster than gfortran? Or are you changing it for another reason? 

We shouldn't need gfortran, but a while a ago there was some issue with gfortran not being found. Testing if that's still an issue. Clang is faster than g++. Therefore, installing clang.

PR #1283, opened 2024-01-16
PetabImporter: version-specific amici model directories

dweindl:
Use different directories for models from different amici versions (e.g., `amici_models/0.20.0/{model_name}` instead of `amici_models/{model_name}`). Allows switching between amici versions without version errors or re-importing the model.

(We could also include the architecture, if considered helpful.)


FFroehlich (approved):
👍

PR #1285, opened 2024-01-16
Set only active maintainers as default codeowners

PaulJonasJost:
This way only the three mentioned will be asked for a review by default.


FFroehlich (approved):

FFroehlich:
thanks!

Doresic (approved):

PR #1287, opened 2024-01-18
Select: use petab_select method to set estimated parameters

dilpath:
Possibly a breaking change, if users expect scaled values at `petab_select.Model.estimated_parameters`

### DIFF
--- a/pypesto/select/model_problem.py
+++ b/pypesto/select/model_problem.py
@@ -165,16 +165,22 @@ class ModelProblem:
         self.model.set_criterion(Criterion.NLLH, float(self.best_start.fval))
         self.model.compute_criterion(criterion=self.criterion)
 
-        self.model.estimated_parameters = {
+        estimated_parameters = {
             id: float(value)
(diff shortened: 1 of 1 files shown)


pypesto/select/model_problem.py
```diff
@@ -165,16 +165,22 @@ def set_result(self, result: Result):
...
-                zip(
-                    self.pypesto_problem.x_names,
-                    self.best_start.x,
-                )
+                dict(
+                    zip(
+                        self.pypesto_problem.x_names,
+                        self.best_start.x,
+                    )
+                ).items()
```
Doresic, line 176: Just out of curiosity, why is using `dict` here better than just using `zip` only as before?

Doresic (approved):
Left a small question. Otherwise looks good. Thanks!

pypesto/select/model_problem.py
dilpath: Hm, not sure why that was added, it's removed now, thanks!

PR #1293, opened 2024-01-25
EnsemblePrediction: remove "no predictor" warning

dweindl:
It's unclear why `EnsemblePrediction.predictor` would be needed. This PR removes a seemingly useless log message.


dilpath (approved):
> It's unclear why `EnsemblePrediction.predictor` would be needed.

Not sure it ever found a use
https://github.com/ICB-DCM/pyPESTO/pull/557#discussion_r562612360

PR #1297, opened 2024-01-29
Semiquantitative: Robust regularization calculation

Doresic:
Making the analytical calculation of the optimal regularization robust to the simulation range. In case the range was 0, the optimal beta was infinite due to division by 0.


Doresic:
Not sure why the nbqa black run is failing. Locally it works. Will check again.

stephanmg:
Looks like two formatters are run

stephanmg (approved):

stephanmg:
Related to #1300

PR #1305, opened 2024-02-14
Include pymc in the documentation.

PaulJonasJost:
Pymc was not listed in the documentation.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -24,6 +24,7 @@ API reference
    pypesto.profile.profile_next_guess
    pypesto.result
    pypesto.sample
+   pypesto.sample.pymc
    pypesto.select
    pypesto.select.postprocessors
    pypesto.startpoint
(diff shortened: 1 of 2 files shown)


pypesto/sample/pymc.py
```diff
@@ -18,6 +18,14 @@
 
 logger = logging.getLogger(__name__)
 
+# __all__ definition for Sphinx
```
dweindl, line 21:
```suggestion
```

This is somewhat misleading, since `__all__` is not really specifically a sphinx thing but has broader implications.

m-philipps (approved):

pypesto/result/sample.py
```diff
@@ -48,6 +20,34 @@ def __init__(
         effective_sample_size: float = None,
         message: str = None,
     ):
+        """
+        Initialize the McmcPtResult.
```
dilpath, line 24: Why are these now documented as constructor parameters? Attributes seems more appropriate

dilpath (approved):

PR #1306, opened 2024-02-14
Variational inference with PyMC

arrjon:
I added a wrapper for variational inference with `PyMC`. The wrapper uses the `PymcSampler` as basis and directly supports all functions and methods from `PyMC` [see here for details on VI](https://www.pymc.io/projects/examples/en/latest/variational_inference/variational_api_quickstart.html).

To include this properly in `pyPESTO` one would need to add proper testing and write a `Result Object` to save the estimated parameters of the variational distribution.

### DIFF
--- a/pypesto/sample/pymc.py
+++ b/pypesto/sample/pymc.py
@@ -216,6 +216,81 @@ class PymcSampler(Sampler):
 
         self.data = data
 
+    def fit(
+        self,
+        n_iterations: int,
+        method: str = 'advi',
+        random_seed: int = None,
+        start_sigma=None,
+        beta: float = 1.0,
+    ):
+        """
+        Sample the problem.
+
+        Parameters
+        ----------
+        n_iterations:
+            Number of iterations.
+        method: str or :class:`Inference`
+            string name is case insensitive in:
+            -   'advi'  for ADVI
+            -   'fullrank_advi'  for FullRankADVI
+            -   'svgd'  for Stein Variational Gradient Descent
+            -   'asvgd'  for Amortized Stein Variational Gradient Descent
+        random_seed: int
+            random seed for reproducibility
+        start_sigma: `dict[str, np.ndarray]`
+            starting standard deviation for inference, only available for method 'advi'
+        """
+        try:
+            import pymc
+        except ImportError:
+            raise SamplerImportError("pymc")
+
+        problem = self.problem
+        log_post = PymcObjectiveOp.create_instance(problem.objective, beta)
+
+        x0 = None
+        x_names_free = problem.get_reduced_vector(problem.x_names)
+        if self.x0 is not None:
+            x0 = {
+                x_name: val
+                for x_name, val in zip(problem.x_names, self.x0)
+                if x_name in x_names_free
+            }
+
+        # create model context
+        with pymc.Model():
+            # parameter bounds as uniform prior
+            _k = [
+                pymc.Uniform(x_name, lower=lb, upper=ub)
+                for x_name, lb, ub in zip(
+                    x_names_free,
+                    problem.lb,
+                    problem.ub,
+                )
+            ]
+
+            # convert parameters to PyTensor tensor variable
+            theta = pt.as_tensor_variable(_k)
+
+            # define distribution with log-posterior as density
+            pymc.Potential("potential", log_post(theta))
+
+            # record function values
+            pymc.Deterministic("loggyposty", log_post(theta))
+
+            # perform the actual sampling
+            data = pymc.fit(
+                draws=int(n_iterations),
+                method=method,
+                start=x0,
+                random_seed=random_seed,
+                start_sigma=start_sigma,
+            )
+
+        self.data = data
+
     def get_samples(self) -> McmcPtResult:
         """Convert result from pymc to McmcPtResult."""
         # dimensions
--- a/pypesto/sample/pymc.py
+++ b/pypesto/sample/pymc.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import logging
-from typing import Union
+from typing import Optional, Union
 
 import arviz as az
 import numpy as np
@@ -220,10 +220,11 @@ class PymcSampler(Sampler):
         self,
         n_iterations: int,
         method: str = 'advi',
-        random_seed: int = None,
-        start_sigma=None,
-        inf_kwargs=None,
+        random_seed: Optional[int] = None,
+        start_sigma: Optional = None,
+        inf_kwargs: Optional = None,
         beta: float = 1.0,
+        **kwargs,
     ):
         """
         Sample the problem.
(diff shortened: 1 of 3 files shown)


pypesto/sample/pymc.py
```diff
@@ -216,6 +216,89 @@ def sample(self, n_samples: int, beta: float = 1.0):
...
+            # perform the actual sampling
+            data = pymc.fit(
+                n=int(n_iterations),
+                method=method,
+                random_seed=random_seed,
+                start=x0,
+                start_sigma=start_sigma,
+                inf_kwargs=inf_kwargs,
+                **kwargs,
+            )
```
PaulJonasJost, line 298: As far as I could see (might have missed something), this is the only difference aside from Line 171 `trace = self.trace` to the functionality of `sample`. In this case, it might make sense to wrap this into the function either with an additional parameter or with a decorator?

pypesto/sample/variational_inference.py
```diff
@@ -0,0 +1,153 @@
...
+    # result.vi_result = vi_result
+    #
+    # autosave(
+    #     filename=filename,
+    #     result=result,
+    #     store_type="variational",
+    #     overwrite=overwrite,
+    # )
+    #
+    # return result
```
PaulJonasJost, line 129: Comments need to be removed and we need to build the variational result object as you said.

pypesto/sample/variational_inference.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def eval_variational_log_density(
+    x_points: np.ndarray, vi_approx
```
PaulJonasJost, line 133: vi_approx should have a type I guess?

pypesto/sample/variational_inference.py
```diff
@@ -0,0 +1,153 @@
...
+    Call to do parameter sampling.
+
+    Parameters
+    ----------
+    problem:
+        The problem to be solved. If None is provided, a
+        :class:`pypesto.AdaptiveMetropolisSampler` is used.
+    n_iterations:
+        Number of iterations for the optimization.
+    method: str or :class:`Inference`
```
PaulJonasJost, line 41: Unless we link our documentation to pymc, we have no class inference I think 🤔  Should also be indicated in the type of `method`.

arrjon:
Mainly tests are missing, but otherwise ready

pypesto/variational/pymc.py
```diff
@@ -0,0 +1,190 @@
...
+            # perform the actual sampling
+            data = pymc.fit(
+                n=int(n_iterations),
+                method=method,
+                random_seed=random_seed,
+                start=x0,
+                start_sigma=start_sigma,
+                inf_kwargs=inf_kwargs,
+                **kwargs,
+            )
```
Doresic, line 117: Not that common with VI, but I'm a bit confused as to why the return object from the fitting is called data? 
I thought the output would be some sort of model (collection of distributions that were fit to the posterior).
Probably I just don't understand it.

pypesto/variational/variational_inference.py
```diff
@@ -0,0 +1,134 @@
...
+    # TODO: if needed, we can add a result object for variational inference methods
+    result.variational_result = variational
+    (
+        result.sample_result.variational_parameters_names,
+        result.sample_result.variational_parameters,
+    ) = variational.get_variational_parameters()
+    if filename is not None:
+        logger.warning(
+            "Variational parameters are not saved in the hdf5 file. You have to save them manually."
+        )
```
Doresic, line 132: Nice warning. Good to have.

pypesto/variational/variational_inference.py
```diff
@@ -0,0 +1,134 @@
...
+    t_elapsed = process_time() - t_start
+    logger.info("Elapsed time: " + str(t_elapsed))
+
+    # extract results and save samples to pypesto result
+    if n_samples is None or n_samples == 0:
+        # constructing a McmcPtResult object with nearly empty trace_x
+        n_samples = 1
+
+    result.sample_result = variational.sample(n_samples)
+    result.sample_result.time = t_elapsed
```
Doresic, line 113: `t_elapsed` does not include the time it takes to sample with `variational.sample(...)`. Is it negligible?

pypesto/variational/variational_inference.py
```diff
@@ -0,0 +1,134 @@
...
+    autosave(
+        filename=filename,
+        result=result,
+        store_type="sample",
+        overwrite=overwrite,
+    )
+
+    # make pymc object available in result
+    # TODO: if needed, we can add a result object for variational inference methods
+    result.variational_result = variational
```
Doresic, line 124: I guess this will not be saved if the result is saved as there is no `VariationalResultHDF5Writer` class. Would it be possible to save it and would you want it? If not, it might be nice to add a warning in the `write_result()`function of `pypesto.store.save_to_hdf5.py` that check whether there is a variational result and says it's not being saved if there is one.

Doresic: It might be good to add a `variational_result` attribute to the `Result` class, with documentation saying what kind of object will be saved in this attribute.

pypesto/variational/pymc.py
```diff
@@ -0,0 +1,190 @@
+import logging
```
Doresic, line 1: Module-level docstrings would be nice. Short descriptions of what all the `.py` files are there for. Good example in `pypesto.store` modules

Doresic (approved):
All in all, very nice implementation! Thanks.
I left some questions/recommendations that would come up to my mind if I would try to use the module myself.

pypesto/variational/variational_inference.py
arrjon: yes it is

pypesto/variational/pymc.py
arrjon: that is an oddity of pymc and so far I know, their usual name. But I can change that, as it is indeed misleading!

pypesto/variational/variational_inference.py
arrjon: I do not want to add this in the result class, since it is not saved at the moment. But I added a warning in the `write_result()` function now

pypesto/variational/pymc.py
Doresic: Oh ok, no in that case, good to keep consistent.

test/variational/test_variational.py
```diff
@@ -0,0 +1,87 @@
...
+    # create problem
+    problem = create_petab_problem()
+
+    result = variational_fit(
+        problem,
+        n_iterations=100,
+        method="advi",
+        x0=np.array([3, -4]),
+    )
+    return result
```
PaulJonasJost, line 28: What is this used for? Do not see it in the tests anywhere else. If not used, would remove it.

PaulJonasJost (approved):

test/variational/test_variational.py
arrjon: true, this was an old test I wrote, not necessary anymore

PR #1307, opened 2024-02-21
Ruff codechecks

PaulJonasJost:
Most important to check would be 
+ pre-commit-config.yaml
+ pyproject.toml

can flake8 be deleted from tox and .flake8 as well?


.pre-commit-config.yaml
```diff
@@ -54,12 +31,19 @@ repos:
...
-- repo: local
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.1.11
   hooks:
-  - id: style
-    name: Check style
-    description: Check style
-    pass_filenames: false
-    entry: tox -e project,flake8
```
dweindl, line 63: Did you remove `project` on purpose?

If we aren't using flake8 any longer, I'd also remove the respective tox config from `tox.ini`.

pypesto/engine/base.py
```diff
@@ -24,5 +24,6 @@ def execute(
             List of tasks to execute.
         progress_bar:
             Whether to display a progress bar.
+
```
dweindl, line 27: Any good reason for those trailing empty lines?

pyproject.toml
```diff
@@ -11,10 +11,80 @@ build-backend = "setuptools.build_meta"
...
+"test/*" = [
+    "T201",
+    "S101",
+    "D",
+]
+"pypesto/util.py" = [
+    "D400",
+    "D205",
+    "D107",
+]
```
dweindl, line 81: Why different docstring rules in different files?

dweindl (commented):
Thanks. Mostly looks good.

doc/example/workflow_comparison.ipynb
```diff
@@ -468,7 +195,7 @@
     "        \"\"\"\n",
     "        Turn array of parameters to dictionary usable for objective call.\n",
     "        \"\"\"\n",
-    "        return {x_id: x_i for x_id, x_i in zip(self.x_ids, x)}\n",
+    "        return {zip(self.x_ids, x)}\n",
```
dweindl, line 198: fixme: dict vs set

pyproject.toml
```diff
@@ -11,10 +11,80 @@ build-backend = "setuptools.build_meta"
...
 line-length = 79
-target-version = ['py38', 'py39']
 skip-string-normalization = true
 
 [tool.isort]
 profile = "black"
 line_length = 79
 multi_line_output = 3
+
+[tool.ruff]
```
dilpath, line 21: Remove `tool.isort`, `tool.black` above?

dilpath (approved):

stephanmg:
> Most important to check would be
> 
> * pre-commit-config.yaml
> * pyproject.toml
> 
> can flake8 be deleted from tox and .flake8 as well?

Looks good to me. Would remove .flake8.

.pre-commit-config.yaml
PaulJonasJost: As far as I can see in the code, project is just used to initialise the project to be able to run flake8. Not needed for any part of the tests if we remove the flake8 test. Yes, agreed, will remove from tox.ini

pyproject.toml
PaulJonasJost: * I do not really care about the documentation within `examples`, in many cases I think it is not supposed to be documented the same the code base (hence ignore `D` there)
* Same goes for testing and Jupyter notebooks
* All other files (both */init and the other ones). are the ones that have in their heading the following string which is considered docstring:
```
Header
=====

Short description of either the module or the file
```

This produces exactly D400 and D205, which is why in only those files it is excluded, in all other files it should be checked.

dweindl: Ok

dweindl (approved):

FFroehlich:
bit late now, but might make sense to add the PR of this commit to a `.git-blame-ignore-revs`  file (see https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)

PR #1311, opened 2024-02-27
Fix CMAES installation hint

dilpath:
Currently instructs users to install `cma` with `pip install pypesto[cma]`, but this isn't configured in "setup.py".

Now refactored so everything is in terms of `cma` and `CmaOptimizer`.


PaulJonasJost (approved):
Perhaps would be nicer to rename to `cma` in setup?

dilpath:
> Perhaps would be nicer to rename to `cma` in setup?

Agreed, as well as renaming `CmaesOptimizer` to `CmaOptimizer`, since it looks like the other optimizers are named after the package rather than the method. What do you think? Would be an unnecessary breaking change then.

PaulJonasJost:
> > Perhaps would be nicer to rename to `cma` in setup?
> 
> Agreed, as well as renaming `CmaesOptimizer` to `CmaOptimizer`, since it looks like the other optimizers are named after the package rather than the method. What do you think? Would be an unnecessary breaking change then.

I am fine with that, we can have a period where we still allow the usage of `CmaesOptimizer` but then with a deprecation warning? Would not be breaking per se but we can then remove it later?

PaulJonasJost (approved):
Thanks 🙏🏼

PR #1313, opened 2024-02-28
Semiquant: Fix spline inner parameter initial values

Doresic:
Initial values of spline inner parameters of each optimization are set to end values of the previous inner optimization, if possible. However, the saving of previous inner parameters was saving the spline knot values, instead of the spline knot differences, which are the inner parameters.

This wasn't a big issue, as the initial values of the inner optimization are not that important (the problem is convex). But in some cases, they were so different that the inner solver stopped early.

Updated the visualization to work with the change.


stephanmg (commented):
LGTM. Feel free to merge.

stephanmg (approved):

PR #1314, opened 2024-02-29
Semiquant: Add spline knots to the optimization result

### LINKED ISSUE #1308: Make spline parameters available for user for non-linear semi-quantitative data
**Feature description**
When performing hierarchical approximation using splines as the observable mapping one should include spline knots and values for each observable in the `result.optimize_result` object.

**Motivation/Application**
When using non-linear semi-quantitative data and a spline appoximation for the observabale mapping one might like to use this mapping further. For example, in the function `pypesto.visualize.plot_splines_from_pypesto_result` these values are extracted but not given to the user.

Doresic:
Fixes #1308 by saving the spline knots in the `OptimizerResult` object under `spline_knots` in the form of `list[list[np.array[float], np.array[float]]]`.  Each list in the first level corresponds to a semiquantitative observable. Each of these lists contains two arrays: the first array contains the spline bases, the second array contains the spline knot values. The ordering of the observable lists is the same as in `pypesto.problem.hierarchical.semiquant_observable_ids`.

This also allows the visualization `plot_splines_from_pypesto_result` to just extract the spline knot values from the pypesto result object instead of solving the inner problem to get them. The latter could even a bit wrong due to possibly different initial conditions of the inner optimization problem.

Except this, there is some random cleanup here and there.


FFroehlich (approved):
only looked at changes in pypesto/objective/amici/amici.py

PaulJonasJost (approved):
Do we have a test yet, saving and loading this results? If not, please add before merging. Otherwise looks good to me

Doresic:
> Do we have a test yet, saving and loading this results? If not, please add before merging. Otherwise looks good to me

Added a test to `test/hierarchical/test_spline.py`
I'm just running a simple optimization with 2 starts, saving the result, loading it, and checking the spline knots before and after are the same.

m-philipps (approved):
I don't see any issues in the visualize/spline_approximation.py. For the future it would be useful to have you as a codeowner of that file.

PR #1315, opened 2024-03-05
Relative: Fix return of relative calculator if sim fails

Doresic:
In case of simulation fail, the inner relative calculator did not return inner parameters to be added to the inner result. This caused some `ValueError: too many values to unpack (expected 2)` as a tuple was expected.


pypesto/hierarchical/inner_calculator_collector.py
```diff
@@ -476,7 +476,10 @@ def __call__(
                 snllh += inner_result[GRAD]
 
             all_inner_pars.update(inner_result[X_INNER_OPT])
-            if INNER_PARAMETERS in inner_result:
+            if (
+                INNER_PARAMETERS in inner_result
+                and inner_result[INNER_PARAMETERS] is not None
+            ):
                 interpretable_inner_pars.extend(inner_result[INNER_PARAMETERS])
```
dweindl, line 481:
```suggestion
            interpretable_inner_pars.extend(inner_result.get(INNER_PARAMETERS) or [])
```

dilpath (approved):

dweindl (approved):
thx :+1:

pypesto/hierarchical/inner_calculator_collector.py
Doresic: Ohh that's cleaner, ty :D

dweindl:
```python
In [1]: {}.get(1, None) or []
Out[1]: []

In [2]: {1:2}.get(1, None) or []
Out[2]: 2
```

dweindl: Okay, if `get` produces a numpy array, this raises a DeprecationWarning. Not ideal then.

Doresic: Hmm something else is breaking as well in the notebook, will find a solution fixing both.

dweindl: fine to keep the original code

Doresic: Made something in between, not a one-liner but clean enough :D

PR #1319, opened 2024-03-06
Fix unwanted amici model recompilation

### LINKED ISSUE #1318: PetabImporter does not check for existing amici models in location specified by output_folder arg.
**Bug description**
When using the `PetabImporter` the amici model is re-compiled although it is already available at the location specified by the `output_folder` argument.

Before compiling the amici model the `PetabImporter` checks if an amici module is already available, but does not check within subfolders of the current path:
https://github.com/ICB-DCM/pyPESTO/blob/ee852cb216a65b12bf2d1e16edc8c0fba6c3e714/pypesto/petab/importer.py#L355-L356
This does not make use of the `output_folder` argument.

**Expected behavior**
Since by default the amici models are compiled to the folder `amici_models/` I would expect those to be found and not recompiled.

m-philipps:
changes:
- `PetabImporter` checks for existing amici model in the location specified by `output_folder` before re-compiling.
- closes #1318

### DIFF
--- a/pypesto/petab/importer.py
+++ b/pypesto/petab/importer.py
@@ -353,7 +353,11 @@ class PetabImporter(AmiciObjectBuilder):
         # try to import (in particular checks version)
         try:
             # importing will already raise an exception if version wrong
-            importlib.import_module(self.model_name)
+            # importlib.import_module(self.model_name)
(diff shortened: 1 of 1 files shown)


dweindl (approved):
:+1: thx

PR #1320, opened 2024-03-08
Remove CESSOptimizer

dweindl:
`SacessOptimizer` is superior to `CESSOptimizer` and I wouldn't recommend using it. I think it's best to just remove it. No value, just extra maintenance effort. I don't think anybody was using it yet anyways, so this shouldn't actually be a breaking change for anybody. If not, please let me know.


PaulJonasJost (approved):
fine by me, thanks 🙏🏼

PR #1324, opened 2024-03-11
Dynesty warnings added

arrjon:
The wrapper for the dynesty sampler differs significantly to the other samplers in pyPESTO. Usually the samplers expect the objective to be the posterior and `x_priors` to be defined, from which they compute the likelihood if needed. I added the argument `objective_type` to the sampler to switch between posterior and likelihood with default set to posterior. I also added checks in the `initialize` function to very if `x_priors` is defined in the problem or not.

Furthermore, I added a warning that by default, a uniform prior with bounds specified in the problem is assumed. This warning was missing, but already mentioned in the `prior_transform` function.

### DIFF
--- a/pypesto/sample/dynesty.py
+++ b/pypesto/sample/dynesty.py
@@ -176,13 +176,13 @@ class DynestySampler(Sampler):
         else:
             # if objective is the negative log likelihood, we will ignore x_priors even if they are defined
             if self.problem.x_priors is not None:
-                logger.info(
-                    f"Warning: Assuming '{OBJECTIVE_NEGLOGLIKE}' as objective. "
+                logger.warning(
+                    f"Assuming '{OBJECTIVE_NEGLOGLIKE}' as objective. "
                     f"'x_priors' defined in the problem will be ignored."
                 )
 
         # if priors are uniform, we can use the default prior transform (assuming that bounds are set correctly)
-        logger.info(
+        logger.warning(
             "Assuming 'prior_transform' is correctly specified. If 'x_priors' is not uniform, 'prior_transform'"
             " has to be adjusted accordingly."
         )
--- a/pypesto/sample/dynesty.py
+++ b/pypesto/sample/dynesty.py
@@ -30,6 +30,7 @@ from typing import List, Union
 
 import numpy as np
 
+from ..C import OBJECTIVE_NEGLOGLIKE, OBJECTIVE_NEGLOGPOST
 from ..problem import Problem
 from ..result import McmcPtResult
 from .sampler import Sampler, SamplerImportError
@@ -62,7 +63,7 @@ class DynestySampler(Sampler):
         sampler_args: dict = None,
         run_args: dict = None,
         dynamic: bool = True,
-        objective_type: str = "neglogpost",
+        objective_type: str = OBJECTIVE_NEGLOGPOST,
     ):
         """
         Initialize sampler.
@@ -78,8 +79,8 @@ class DynestySampler(Sampler):
         dynamic:
             Whether to use dynamic or static nested sampling.
         objective_type:
-            The objective to optimize (as defined in  pypesto.problem). Either "neglogpost" or "negloglike".
-            If neglogpost, x_priors have to be defined in the problem.
+            The objective to optimize (as defined in  pypesto.problem). Either "neglogpost" or
+            "negloglike". If "neglogpost", x_priors have to be defined in the problem.
         """
         # check dependencies
         import dynesty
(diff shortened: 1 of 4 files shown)


pypesto/sample/dynesty.py
```diff
@@ -62,6 +62,7 @@ def __init__(
         sampler_args: dict = None,
         run_args: dict = None,
         dynamic: bool = True,
+        objective_type: str = "neglogpost",
```
dilpath, line 65: Add some string constants to `pypesto/C.py` for the options here

pypesto/sample/dynesty.py
```diff
@@ -130,7 +138,13 @@ def loglikelihood(self, x):
         if any(x < self.problem.lb) or any(x > self.problem.ub):
             return -np.inf
         # invert sign
-        # TODO this is possibly the posterior if priors are defined
+        if self.objective == "neglogpost":
```
dilpath, line 141:
```suggestion
        if self.objective_type == "neglogpost":
```

pypesto/sample/dynesty.py
```diff
@@ -130,7 +138,13 @@ def loglikelihood(self, x):
...
+            # problem.objective returns negative log-posterior
+            # compute log-likelihood by subtracting log-prior
+            return -1.0 * (
+                self.problem.objective(x) - self.problem.x_priors(x)
```
dilpath, line 145: If not already possible, please open an issue so this is easier via e.g. `self.problem.objective(x, priors=False)`, seems like it would be a common thing to check likelihood independently of priors in other contexts

pypesto/sample/dynesty.py
```diff
@@ -149,6 +163,27 @@ def initialize(
...
+            if self.problem.x_priors is not None:
+                logger.info(
+                    "Warning: Assuming 'negloglike' as objective. 'x_priors' defined in the problem will be ignored."
+                )
+
+        # if priors are uniform, we can use the default prior transform (assuming that bounds are set correctly)
+        logger.info(
+            "Assuming 'prior_transform' is correctly specified. If 'x_priors' is not uniform, 'prior_transform'"
+            " has to be adjusted accordingly."
+        )
```
dilpath, line 185: I'll try to make this nicer, we should be able to support e.g. Gaussian transform at least

pypesto/sample/dynesty.py
```diff
@@ -149,6 +163,27 @@ def initialize(
...
+                )
+        else:
+            # if objective is the negative log likelihood, we will ignore x_priors even if they are defined
+            if self.problem.x_priors is not None:
+                logger.info(
+                    "Warning: Assuming 'negloglike' as objective. 'x_priors' defined in the problem will be ignored."
+                )
+
+        # if priors are uniform, we can use the default prior transform (assuming that bounds are set correctly)
+        logger.info(
```
dilpath, line 182:
```suggestion
        warnings.warn(
```

dilpath (changes requested):
Thanks!

pypesto/sample/dynesty.py
arrjon: defining the `prior_transfromation`, if mean and cov of the Gaussian prior are known, would be easy. However, I am not so sure how to extract these information out of the current pyesto problem/objective.

vwiela:
Do we want to add this flag only in the dynesty sampler? I think it would be better to add it on the objective level.

pypesto/sample/dynesty.py
dilpath: OK, then let's wait until pyPESTO has better objective attributes

dilpath (approved):
Feel free to merge with just the warning and leave the `objective_type` things to a new PR where objective attributes are added.

Doresic (approved):
Looks good. Good temporary solution for dynesty.

I have a general question, possibly stupid. 
As much as I understand, the OBJECTIVE_NEGLOGLIKE and NEGLOGPOST constants are only passed to the dynesty sampler upon construction such that it knows whether it needs to remove the priors from the objective function when calculating the likelihood.

My question is whether the constants are necessary? Why not define the loglikelihood of dynesty to always remove the prior in cases a prior exists? The only case when it should not remove the prior from the objective is when it does not exist, right? I might be missing something.

arrjon:
> My question is whether the constants are necessary? Why not define the loglikelihood of dynesty to always remove the prior in cases a prior exists? The only case when it should not remove the prior from the objective is when it does not exist, right? I might be missing something.

If you load a petab problem, that should work fine. But if you define your objective yourself (which could be the nllh) and you define `x_priors` as well, you would compute something wrong without giving any warnings. That is not only a problem in dynesty, but also with the other samplers, see #1325.

PR #1326, opened 2024-03-13
Adding Thermodynamic Integration

arrjon:
I added a small function `compute_log_evidence` to the `ParallelTemperingSampler` which allows one to compute the log-evidence of a model after sampling. This is based on thermodynamic integration, which computes the integral over the mean log likelihood, where the mean is taken over the samples from the tempered posteriors. Usually a different temperature schedule is used with thermodynamic integration, which can be added as an option during sampling by
```
sampler = sample.ParallelTemperingSampler(
            internal_sampler=sample.AdaptiveMetropolisSampler(),
            options={"beta_init": "beta_decay"},
            n_chains=n_chains,
)
```

The evidence can be calculated by
```
 log_evidence = compute_log_evidence(result)
```

### DIFF
--- a/pypesto/sample/parallel_tempering.py
+++ b/pypesto/sample/parallel_tempering.py
@@ -4,6 +4,7 @@ from typing import Dict, List, Sequence, Union
 
 import numpy as np
 
+from ..C import BETA_DECAY, EXPONENTIAL_DECAY
 from ..problem import Problem
 from ..result import McmcPtResult, Result
 from ..util import tqdm
@@ -46,13 +47,13 @@ class ParallelTemperingSampler(Sampler):
         # set betas
         if (betas is None) == (n_chains is None):
             raise ValueError("Set either betas or n_chains.")
-        if betas is None and self.options["beta_init"] == "exponential_decay":
+        if betas is None and self.options["beta_init"] == EXPONENTIAL_DECAY:
             betas = near_exponential_decay_betas(
                 n_chains=n_chains,
                 exponent=self.options["exponent"],
                 max_temp=self.options["max_temp"],
             )
-        elif betas is None and self.options["beta_init"] == "beta_decay":
+        elif betas is None and self.options["beta_init"] == BETA_DECAY:
             betas = beta_decay_for_beta(
                 n_chains=n_chains, alpha=self.options["alpha"]
             )
@@ -78,7 +79,7 @@ class ParallelTemperingSampler(Sampler):
             "exponent": 4,
             "temper_log_posterior": False,
             "show_progress": None,
-            "beta_init": "exponential_decay",
+            "beta_init": BETA_DECAY,  # replaced in adaptive PT
             "alpha": 0.3,
         }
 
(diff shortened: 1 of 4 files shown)


pypesto/sample/parallel_tempering.py
```diff
@@ -164,6 +173,72 @@ def swap_samples(self) -> Sequence[bool]:
...
+def beta_decay_for_beta(n_chains: int, alpha: float) -> np.ndarray:
+    """Initialize betas to the (j-1)th quantile of a Beta(alpha, 1) distribution.
+
+    Parameters
+    ----------
+    n_chains:
+        Number of chains to use.
+    alpha:
+        Tuning parameter that modulates the skew of the distribution over the temperatures.
+        For alpha=1 we habe a uniform distribution, and as alpha decreases towards zero,
```
dilpath, line 227:
```suggestion
        For alpha=1 we have a uniform distribution, and as alpha decreases towards zero,
```

pypesto/sample/parallel_tempering.py
```diff
@@ -164,6 +173,72 @@ def swap_samples(self) -> Sequence[bool]:
...
+        temperatures become positively skewed. Xie et al. (2011) propose alpha=0.3 as a good start.
+
+
+    Proposed by Xie et al. (2011) to be used for thermodynamic integration.
```
dilpath, line 231: This kind of statement is usually above the Parameters section, in pyPESTO docs

pypesto/sample/parallel_tempering.py
```diff
@@ -71,6 +78,8 @@ def default_options(cls) -> Dict:
             "exponent": 4,
             "temper_log_posterior": False,
             "show_progress": None,
+            "beta_init": "exponential_decay",
```
dilpath, line 81: Is one of `beta_decay` or `exponential_decay` better in general? Since `beta_decay` is better for evidence, should we change the default to `beta_decay`?

pypesto/sample/parallel_tempering.py
```diff
@@ -43,12 +46,16 @@ def __init__(
...
         if (betas is None) == (n_chains is None):
             raise ValueError("Set either betas or n_chains.")
-        if betas is None:
+        if betas is None and self.options["beta_init"] == "exponential_decay":
             betas = near_exponential_decay_betas(
                 n_chains=n_chains,
                 exponent=self.options["exponent"],
                 max_temp=self.options["max_temp"],
             )
+        elif betas is None and self.options["beta_init"] == "beta_decay":
```
dilpath, line 55: Some constant in `C.py` for `"beta_decay"` and `"exponential_decay"` would be nice

dilpath (approved):
Thanks :+1:

pypesto/sample/parallel_tempering.py
arrjon: If you use the adaptive parallel tempering then you should use `exponential_decay`, that seems to be what Vousden et al. 2016. proposes. I do not think anybody will use the non-adaptive version, if not for evidence estimation so we can switch the default here.

pypesto/sample/parallel_tempering.py
```diff
@@ -164,6 +173,72 @@ def swap_samples(self) -> Sequence[bool]:
...
+            )
+        else:
+            raise ValueError(
+                f"Unknown method {method}. Choose 'trapezoid' or 'simpson'."
+            )
+
+        return log_evidence
+
+
+def beta_decay_for_beta(n_chains: int, alpha: float) -> np.ndarray:
```
vwiela, line 219: Maybe change function name to `beta_decay_betas` to be more similar and consistent with the other function name `near_exponential_decay_betas`.

vwiela (approved):
Thanks, looks good to me in general.

pypesto/sample/parallel_tempering.py
dilpath: Sounds good :+1: The choice of decay method should be communicated to the user somehow, e.g. docs or logs. Probably easiest to do some `logging.info` message that prints the options used, either always or only when the user doesn't specify the decay method

Doresic (approved):
Very nice. Thank you!

PR #1327, opened 2024-03-18
Hierarchical optimization: fix unnecessary simulation

dweindl:
Currently model simulation is run twice, even when no sensitivities are requested and the optimal inner parameters could be applied directly.

### DIFF
--- a/pypesto/hierarchical/relative/calculator.py
+++ b/pypesto/hierarchical/relative/calculator.py
@@ -132,10 +132,10 @@ class RelativeAmiciCalculator(AmiciCalculator):
             )
 
         if (
-            amici_solver.getSensitivityMethod()
-            == amici.SensitivityMethod_adjoint
-            or 2 in sensi_orders
-        ):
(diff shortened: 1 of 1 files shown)


Doresic (approved):
Yes, nice change. Didn't think of this case in the implementation.
To confirm we're thinking the same: the only thing this changes is that if the sensitivity method is adjoint, then the `sensi_orders = (0,)` calls of the objective will be called through `self.calculate_directly(...)` instead of `self.call_amici_twice(...)`, right?

dweindl:
> To confirm we're thinking the same: the only thing this changes is that if the sensitivity method is adjoint, then the `sensi_orders = (0,)` calls of the objective will be called through `self.calculate_directly(...)` instead of `self.call_amici_twice(...)`, right?

Correct.

It was necessary because checking for sensitivity method alone is insufficient. sensitivity method and sensitivity order jointly decide whether sensitivities are to be computed.

PR #1328, opened 2024-03-20
ESSOptimizer: Respect local_n2 in case of failed initial local search

dweindl:
Fixes an issue where ESSOptimizer would try to repeatedly perform a local search from the best known value, instead of trying other promising points, in case the initial local search yields a non-finite result.

### DIFF
--- a/pypesto/optimize/ess/ess.py
+++ b/pypesto/optimize/ess/ess.py
@@ -172,6 +172,7 @@ class ESSOptimizer:
         # Index of current iteration
         self.n_iter: int = 0
         # ESS iteration at which the last local search took place
+        # (only local searches with a finite result are counted)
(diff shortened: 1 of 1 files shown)


pypesto/optimize/ess/ess.py
```diff
@@ -172,6 +172,7 @@ def _initialize(self):
         # Index of current iteration
         self.n_iter: int = 0
         # ESS iteration at which the last local search took place
+        # (only local searches with a finite result are counted)
```
PaulJonasJost, line 175: That's a good addition here, I assume (looking at the code) that the problem occurred since neither local solution was set nor did self.n_iter go up? In this case a similar info on `self.local_solutions` might also be good?

PaulJonasJost (approved):

PR #1331, opened 2024-03-20
dynesty: new method to save only the raw dynesty sample results

dilpath:
With this change, things like AMICI models that exist in the dynesty loglikelihood, are not saved in the dynesty internal result.

Dummy pyPESTO objects (e.g. for analysis or plotting) can be created by supplying the `raw_results` to `get_original_dynesty_samples` or `get_mcmc_like_dynesty_samples`.


arrjon:
Thanks Dilan! I check your branch and something seems not quite right yet

- some results seem to be discared: `sampler.results.logz` does not have the same size before and after loading it 
- Moreover, if I call `restore_internal_sampler` I get `AttributeError: 'DynestySampler' object has no attribute 'results'`.  However, using directly `dynesty.utils.restore_sampler` works fine. (this issue was already there before your updates)

dilpath:
> * some results seem to be discared: `sampler.results.logz` does not have the same size before and after loading it

Works fine for me, `logz` is the same shape before and after using `save_raw_results` and `load_raw_results`. However, these methods do not save the `sampler` object. Could you try e.g.
```python
print(sampler.sampler.results.logz.shape)
pypesto.sample.dynesty.save_raw_results(sampler, "dynesty.raw")
raw_results = pypesto.sample.dynesty.load_raw_results("dynesty.raw")
print(raw_results.logz.shape)
```


> * Moreover, if I call `restore_internal_sampler` I get `AttributeError: 'DynestySampler' object has no attribute 'results'`.  However, using directly `dynesty.utils.restore_sampler` works fine. (this issue was already there before your updates)

Also cannot reproduce, e.g. with
```python
print(sampler_original.sampler.results.logz.shape)
sampler_original.save_internal_sampler("dynesty.dill")
sampler = pypesto.sample.DynestySampler(objective_type="negloglike")
sampler.restore_internal_sampler("dynesty.dill")
print(sampler.sampler.results.logz.shape)
```

> `AttributeError: 'DynestySampler' object has no attribute 'results'`

Maybe you are trying to access the pyPESTO sampler object, instead of the dynesty sampler object?
```python
# pyPESTO sampler object
sampler = pypesto.sample.DynestySampler()
# dynesty sampler object
sampler.sampler
```

`sampler.sampler` will have `sampler.sampler.results`.

arrjon:
Now this is fixed, thanks!

arrjon (approved):
LGSTM now :)

PR #1333, opened 2024-03-21
Relative: Fix return of inner parameters on objective call

Doresic:
There was as bad implementation of inner parameter returning on call of objective.
Previously, the `InnerCalculatorCollector` would take note of `self.best_fval` such that it returns `inner_parameters` only in cases a better objective function value was found. This was ok during optimization, but then if one wanted to get inner parameters for some other outer parameter vector for which the objective function would not be better, the inner parameters would not be returned. 

This is bad practice: to save current states in the calculator and objective objects. So now, after an optimization start is finished, the `hierarchical_decorator` instead calls the objective once more with the best outer parameters found to obtain the inner parameters for that outer vector.

There some small update in the visualize of `model_fit.time_trajectory_model`. Was not working with the current implementation due to recent hierarchical structure changes.

### DIFF
--- a/pypesto/hierarchical/inner_calculator_collector.py
+++ b/pypesto/hierarchical/inner_calculator_collector.py
@@ -112,7 +112,6 @@ class InnerCalculatorCollector(AmiciCalculator):
 
     def initialize(self):
         """Initialize."""
-        self.best_fval = np.inf
         for calculator in self.inner_calculators:
             calculator.initialize()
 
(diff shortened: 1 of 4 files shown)


dweindl (approved):
Thanks. Looks good to me.

stephanmg (approved):

FFroehlich (approved):
LGTM

PaulJonasJost (approved):

m-philipps (approved):

PR #1338, opened 2024-03-26
Select: support user-provided calibration results

dilpath:
Currently, the design is that:
- a user can provide previously-calibrated models as `user_calibrated_models`
- the design of PEtab Select was changed to accommodate this. A model selection iteration now involves two stages: "calibration" and "post-calibration". pyPESTO will not see the user-provided calibration results until the "post-calibration" stage of each iteration.
- ~~PEtab Select will put these models into the candidate space, instead of the uncalibrated versions~~
- ~~pyPESTO still needs to skip these models~~

The purpose is that the path through model space is still reproducible, because models are not skipped.
~~TODO: fully handle this in PEtab Select, so pyPESTO never sees already-calibrated models, except when it asks PEtab Select for the final results.~~


PaulJonasJost (approved):
LGTM 👍🏼 Is the functionality provided in petab select already?

Doresic (approved):
Nice feature, thanks!

dilpath:
Now handled on the PEtab Select side

PR #1346, opened 2024-03-28
fix jax deprecations

FFroehlich:
- Replaces experimental `host_callback` calls in jax, which are now deprecated, with `pure_callback` calls.
- Replaces caching with appropriate calls to the inner objective as caching results in technically non-pure. (Caching the way it was implemented might work anyways, but  that's a bit iffy). This required removal of second order code, but I am doubtful that part was functional anyways.
- Removes superfluous calls to `jax.jit` in `Jax.Objective`.
- Added compatibility with `jax.vmap`. This required quite a bit of refactoring as the base objective assumes inputs/outputs to be numpy arrays, which is incompatible with jax batch tracers.
- Refactored tests to actually test interoperability with `jax.grad` and `jax.vmap`.


pypesto/objective/jax/base.py
```diff
@@ -214,7 +166,44 @@ def check_sensi_orders(self, sensi_orders, mode: ModeType) -> bool:
...
+    def __call__(
+        self,
+        x: jnp.ndarray,
+        sensi_orders: Tuple[int, ...] = (0,),
+        mode: ModeType = MODE_FUN,
+        return_dict: bool = False,
+        **kwargs,
+    ) -> Union[jnp.ndarray, Tuple, ResultDict]:
+        """
+        See `ObjectiveBase` for more documentation.
```
PaulJonasJost, line 183:
```suggestion
        See :class:`ObjectiveBase` for more documentation.
```
To create a link to base class?

pypesto/objective/jax/base.py
```diff
@@ -226,46 +215,26 @@ def call_unprocessed(
         """
         See `ObjectiveBase` for more documentation.
```
PaulJonasJost, line 216:
```suggestion
        See :class:`ObjectiveBase` for more documentation.
```
Same as above

test/base/test_objective.py
```diff
@@ -214,42 +214,55 @@ def test_aesara(max_sensi_order, integrated):
...
-            )
+        jaxfun = _obj
+        if max_sensi_order == 1:
+            jaxfun = jax.grad(jaxfun)
+        # check compatibility with vmap and jit
+        vmapped = jax.vmap(jaxfun)
+        rvals_jax = vmapped(xx)
+        atol = 0
+        # also need to account for roundoff errors in input, so we are not
+        # we can't use rtol = 1e-8 for 32bit
```
PaulJonasJost, line 257:
```suggestion
        # also need to account for roundoff errors in input, so we
        # can't use rtol = 1e-8 for 32bit
```
Or some other typo probably?

PaulJonasJost (approved):
LGTM, not clear on the very details of the implementation though, but the tests checks out 👌🏼  thanks for the update

FFroehlich:
Now no longer requires passing of input function. Extended tests to demonstrate jax transformations of inputs and outputs.

PR #1347, opened 2024-04-02
Added deepcpy functionality and small test for roadrunner.

PaulJonasJost:
Deepcopy functionality is needed for profiling.


pypesto/objective/roadrunner/road_runner.py
```diff
@@ -69,6 +70,26 @@ def __init__(
         self.solver_options = solver_options
         super().__init__(x_names=x_names)
 
+    def __deepcopy__(
+        self, memodict: Optional[dict] = None
+    ) -> "RoadRunnerObjective":
+        """Deepcopy function for RoadRunner objective.
+
+        Needed for e.g. profiling.
+        """
```
dweindl, line 79:
```suggestion
        """Deepcopy function for RoadRunner objective."""
```

pypesto/objective/roadrunner/road_runner.py
```diff
@@ -69,6 +70,26 @@ def __init__(
...
+        import roadrunner
+
+        other = self.__class__.__new__(self.__class__)
+
+        for key in set(self.__dict__.keys()) - {"roadrunner_instance"}:
+            other.__dict__[key] = copy.deepcopy(self.__dict__[key])
+        other_rr = roadrunner.RoadRunner()
+        state = self.roadrunner_instance.saveStateS()
+        other_rr.loadStateS(state=state)
+        other.roadrunner_instance = other_rr
```
dweindl, line 89: Maybe you can check with the RR devs, if `__deepcopy__` could be provided by RR.

dweindl (approved):

pypesto/objective/roadrunner/road_runner.py
dilpath: also can use their `__setstate__` and `__getstate__`

https://github.com/sys-bio/roadrunner/blob/92d063b8877d121c2c1f878fcc1b5ee093b7ec0b/wrappers/Python/roadrunner/roadrunner.i#L1386-L1392

dweindl: Actually, if we remove

https://github.com/ICB-DCM/pyPESTO/blob/321ee9dbd0223da0c08172fc08d61796e0d176aa/pypesto/objective/base.py#L59-L64

this `__deepcopy__` function here wouldn't be needed. 

Does anybody recall the purpose of `ObjectiveBase.__deepcopy__`?

dilpath: There's an option that `deepcopy`s the objective at `Problem` instantiation.
https://github.com/ICB-DCM/pyPESTO/blob/34e89b3bc88d2052ca808da43c11c57da75bec04/pypesto/problem/base.py#L107-L108

But I'm not sure why `__deepcopy__` is defined, I would expect a default `deepcopy(...)` to do what is defined in `ObjectiveBase.__deepcopy__` already.

dweindl: > I would expect a default `deepcopy(...)` to do what is defined in `ObjectiveBase.__deepcopy__` already.

Exactly.

PaulJonasJost: agreed. Will remove the default deep copy and see whether all tests still work, then merge for now. Might be reverted when RR developers add a deepcopy function.

dweindl: > Will remove the default deep copy and see whether all tests still work, then merge for now.

Not only `ObjectiveBase.__deepcopy__`, but also `RoadRunnerObjective.__deepcopy__`.

> Might be reverted when RR developers add a deepcopy function.

No, that wouldn't require any change then.

PaulJonasJost: but RR does not have a deep copy functionality yet afaik. 🤔

dweindl: `deepcopy` will fall back to `__getstate__`/ `__setstate__`, which exist (see Dilan's comment above)

PaulJonasJost: good 👍🏼 will remove it but keep the tests, as I think we should still make sure it is working?

test/base/test_roadrunner.py
```diff
@@ -104,3 +106,33 @@ def _execute_case_rr(case, model_type, version):
         )
 
     logger.info(f"Case {version}/{model_type}/{case} passed.")
+
+
+def test_deepcopy():
+    "Test that deepcopy works as intended"
```
dweindl, line 112:
```suggestion
    """Test that deepcopy works as intended"""
```

PR #1348, opened 2024-04-02
Added a brief roadrunner example.

PaulJonasJost:
To be extended with multiprocessing, profiling and sampling, when tested through (e.g. profiling when #1347 is merged)


Doresic (approved):
Looks good! Shows all the important stuff, works nicely, and is very fast.

PR #1351, opened 2024-04-02
Ensembles: don't expect `OptimizerResult.id` to be convertible to `int`

### LINKED ISSUE #1349: `Ensemble.from_optimization_{history,endpoints}` assumes that `OptimizerResult.id` is always convertible to `int`
`pypesto.ensemble.ensemble.Ensemble.from_optimization_{history,endpoints}` assumes that `OptimizerResult.id` is always convertible to `int`

Here: 
https://github.com/ICB-DCM/pyPESTO/blob/82b7d5da1273b9d442a634e5aabbda9685279d77/pypesto/ensemble/ensemble.py#L672
and here:
https://github.com/ICB-DCM/pyPESTO/blob/82b7d5da1273b9d442a634e5aabbda9685279d77/pypesto/ensemble/ensemble.py#L803

Since the attribute is called `id` (and not `index` or anything the like), I don't think this is valid.

Those values are used to populate `Ensemble.vector_tags: Sequence[tuple[int, int]]` which I don't see being used anywhere. Therefore, I'd suggest to just use the id string without attempting conversion to int and changing the type of `Ensemble.vector_tags` to `Sequence[Any]`.

dweindl:
Fixes `Ensemble.from_optimization_{history,endpoints}`, which incorrectly assumed that `OptimizerResult.id` is always convertible to `int`.

Closes #1349.

### DIFF
--- a/pypesto/ensemble/ensemble.py
+++ b/pypesto/ensemble/ensemble.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 from functools import partial
-from typing import TYPE_CHECKING, Callable, Optional, Sequence, Union
+from typing import TYPE_CHECKING, Any, Callable, Optional, Sequence, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


dilpath (approved):
Agreed, looks like the IDs are used as a string elsewhere too

https://github.com/ICB-DCM/pyPESTO/blob/34e89b3bc88d2052ca808da43c11c57da75bec04/pypesto/result/optimize.py#L335-L339

https://github.com/ICB-DCM/pyPESTO/blob/34e89b3bc88d2052ca808da43c11c57da75bec04/pypesto/optimize/util.py#L169-L171

m-philipps (approved):
Looks fine to me, thanks

PaulJonasJost (approved):
LGTM, thanks 🙏🏼

PR #1352, opened 2024-04-02
pre-commit: pyupgrade

dweindl:
Stop using deprecated type annotations and imports.


dilpath (approved):
Add to ".git-blame-ignore-revs"?

dweindl:
> Add to ".git-blame-ignore-revs"?

Yes, but only after merging.

FFroehlich (approved):
👍 please add 64efde1 to https://github.com/ICB-DCM/pyPESTO/blob/develop/.git-blame-ignore-revs and use ruff (UP, https://docs.astral.sh/ruff/rules/#pyupgrade-up)

.pre-commit-config.yaml
```diff
@@ -47,3 +47,10 @@ repos:
       args:
         - --config
         - pyproject.toml
+
+- repo: https://github.com/asottile/pyupgrade
+  rev: v3.15.0
+  hooks:
+  - id: pyupgrade
+    args: ["--py39-plus"]
```
FFroehlich, line 55: Is this not automatically picked up from `python-requires` in `setup.cfg`?

dweindl: At least I didn't find that documented.

dweindl:
> 👍 please add [64efde1](https://github.com/ICB-DCM/pyPESTO/commit/64efde11b78921e35b8d1323e13fbac69224ed4a) to https://github.com/ICB-DCM/pyPESTO/blob/develop/.git-blame-ignore-revs 

After merge. Hash will change.

> and use ruff (UP, https://docs.astral.sh/ruff/rules/#pyupgrade-up)

:+1: Didn't notice it was available through ruff.

pypesto/visualize/sampling.py
```diff
@@ -1,7 +1,8 @@
 import logging
 import warnings
+from collections.abc import Sequence
```
PaulJonasJost, line 3: Any particular reason to use this one? Seems like just an extra import to me.

PaulJonasJost (approved):

pypesto/visualize/sampling.py
dweindl: Yes. The former is deprecated since python 3.9.

pypesto/hierarchical/inner_calculator_collector.py
```diff
@@ -89,9 +90,9 @@ class InnerCalculatorCollector(AmiciCalculator):
     def __init__(
         self,
         data_types: set[str],
-        petab_problem: "petab.Problem",
+        petab_problem: petab.Problem,
         model: AmiciModel,
-        edatas: list["amici.ExpData"],
+        edatas: list[amici.ExpData],
```
Doresic, line 95: I thought that adding " " on petab and amici objects was good for tests not to break if the module is imported, but petab and amici are not installed. I think I tried having it like this, but then tests were failing. Why is it ok to change now?

Doresic (approved):
Thank you! It's nice to have it changed everywhere.

pypesto/hierarchical/inner_calculator_collector.py
dweindl: tl;dr: `from __future__ import annotations`

details: https://peps.python.org/pep-0563/

.pre-commit-config.yaml
dweindl: > but we could anyways move all of the information to a single file, right?

Yes, it would be good switch to fully switch to `pyproject.toml`, but that's something for another day.

PR #1363, opened 2024-04-09
Add support for complex noise formulae in roadrunner

PaulJonasJost:
This currently assumes two things:
* The noise formula for one observable is always the same
* There is no time dependent changes in the parameters of the formula

I think both assumptions are things we make implicitly in PEtab/pyPESTO?

### DIFF
--- a/pypesto/objective/roadrunner/petab_importer_roadrunner.py
+++ b/pypesto/objective/roadrunner/petab_importer_roadrunner.py
@@ -103,12 +103,6 @@ class PetabImporterRR:
                 pattern = r"noiseParameter1_(.*?)($|\s)"
                 observable_name = re.search(pattern, noise_formula).group(1)
                 to_change.append((i_edata, j_formula, observable_name))
-
-                # raise NotImplementedError(
-                #     "Noise formulae must be either constants or single "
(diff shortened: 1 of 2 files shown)


pypesto/objective/roadrunner/petab_importer_roadrunner.py
```diff
@@ -85,20 +86,44 @@ def _check_noise_formulae(
...
+                self.rr.addAssignmentRule(
+                    f"noiseFormula_{obs_name}",
+                    original_formula,
+                    forceRegenerate=False,
                 )
+                self.rr.regenerateModel()
+                formulae_changed.append((obs_name, original_formula))
+            except RuntimeError as e:
+                if (obs_name, original_formula) not in formulae_changed:
+                    raise e
```
Doresic, line 126: Might be nicer to move the if clause up there where try: is and remove the try-except.

Doresic (approved):
Looks good!

dweindl:
> * There is no time dependent changes in the parameters of the formula
> 
> 
> I think both assumptions are things we make implicitly in PEtab/pyPESTO?

I don't think pypesto should assume that without verification.
For PEtab/amici it's just a limitation of amici.

PR #1364, opened 2024-04-09
Added option to sample startpoints of a problem, from the problem directly

PaulJonasJost:
Found it very unintuitive to call `problem.startpoint_method(n_starts, problem)`.


Doresic (approved):
This is a nice addition, will be useful to have in some cases. Thanks!

m-philipps (approved):
Thanks!

PR #1365, opened 2024-04-09
Correlation plot with nans

PaulJonasJost:
Added a test. There was previously an error, when "all" start indices were selected, where no general checks were run.


stephanmg (approved):
LGTM.

pypesto/visualize/misc.py
```diff
@@ -359,6 +359,13 @@ def process_start_indices(
         if start_index < len(result.optimize_result)
     ]
 
+    # filter out the indices that are not finite
+    start_indices = [
+        start_index
+        for start_index in start_indices
+        if np.isfinite(result.optimize_result[start_index].fval)
+    ]
```
Doresic, line 367: Should we check for `np.isnan` here as well?

Doresic (commented):
The `process_start_indices` function is used by multiple visualization routines. And some of them also have specific ways of dealing with infinite values. I think those should then be updated as well accordingly if `process_start_indices` doesn't return start indices with infinite values anymore. At least removed, so that the start indices are not checked for infinite values.

We should also think about adding a warning or logging entry that some start indices were removed from plotting due to infinite values (if that's the case). That's, for instance, done in the `parameters.py` plotting routine.

pypesto/visualize/misc.py
stephanmg: np.isfinite tests for `Not a Number` already.

stephanmg:
> We should also think about adding a warning or logging entry that some start indices were removed from plotting due to infinite values (if that's the case). That's, for instance, done in the `parameters.py` plotting routine.

Could that be simply a string message with: "Removed XY entries due to infinite or NaN value"?

m-philipps (approved):
Looks good to me, I would update the `process_start_indices` description, e.g.:
```
Create an array of indices if a number was provided, checks that the
indices do not exceed the max_index and removes starts with non-finite fval.
```

PR #1371, opened 2024-04-15
Removing Benchmark collection for deploy. Update action versions.

PaulJonasJost:
Deploy did not work due to the benchmark collection being a direct link

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -13,10 +13,15 @@ jobs:
 
     steps:
     - name: Check out repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
+
+    - name: Remove direct dependencies from setup.cfg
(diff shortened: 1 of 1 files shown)


dweindl (approved):

.github/workflows/deploy.yml
```diff
@@ -13,10 +13,15 @@ jobs:
 
     steps:
     - name: Check out repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
+
+    - name: Remove direct dependencies from setup.cfg
+      # Remove any "git+https"-based dependencies that are not supported
+      #  by PyPI and are only required for testing.
+      run: sed -i '/git+https/d' python/sdist/setup.cfg
```
dweindl, line 21: Oops. Should be `      run: sed -i '/git+https/d' setup.cfg`

PR #1373, opened 2024-04-16
fixed wrong filepath for config file.

PaulJonasJost:
I had a wrong file path in here. Needed fixing.


Doresic (approved):
Seems good. The setup.cfg file is just in the root of the repository, so I think this is correct.

Issue #1375, opened 2024-04-17
Hierachical optimization with conditions specific for observables fails


arrjon:
**Bug description**
Hierachical optimization with multipe conditions fails in the specific case if one has multiple conditions and observables specific to that condition. Amici returns zeros for those observables which are not seen in a condition and it seems like this is not correctly handled within pyPESTO.

These error occur
```
pypesto/hierarchical/relative/util.py:458: RuntimeWarning: divide by zero encountered in log
  0.5 * np.nansum(np.log(2 * np.pi * sigma_i**2))
```
```
pypesto/hierarchical/inner_calculator_collector.py:558: RuntimeWarning: divide by zero encountered in log
  np.log(2 * np.pi * sigma_i**2) + (data_i - sim_i) ** 2 / sigma_i**2
```
and not even one single start fails or stops. It just runs forever. I debugged the `nllh` value and it is always `-inf`.

**To reproduce**
I attached a minimal petab problem to recreate the issue.
[test.zip](https://github.com/ICB-DCM/pyPESTO/files/15007699/test.zip)

It is a minimal parameter estimation problem with two conditions. The model produces an observable `y` which is then mapped via an observable mapping to either `obs_y_1 = scaling_obs_y_1 * y + offset_obs_y_1` or `obs_y_2 = scaling_obs_y_2 * y + offset_obs_y_2`. Measurements for `obs_y_1` are only given by under the condition `condition1` and `obs_y_2` only for `condition2`.

**Environment**
- Operating system:  macOS 14.4.1
- `pypesto` version: current develop branch
- Python version: `python3.9`

arrjon:
That the `nllh` is `inf` seems also to be the case if one disables the hierachical optimization!

arrjon:
The problem lies in the observable mapping: parameters which are replaced by parameters from the petab measurement file cannot change with the condition. If you define them explicitly and replace them using the petab condition file, it works. However, then hierachical optimization is not supported anymore.

However, the behaviour of pypesto is very strange in that case and I would expect failure at some point since the opimizations only returns `inf`, but continues anyway.

Doresic:
> The problem lies in the observable mapping: parameters which are replaced by parameters from the petab measurement file cannot change with the condition.

To confirm, we showed this is not true, right?

The issue with hierarchical optimization was coming up due to some non-intended logarithms of 0, look at #1377 

[...]

PR #1377, opened 2024-04-17
Relative: fix log of zero for default 0 sigma values

Doresic:
If an observable has hierarchical noise, but has no measurement at some simulation timepoint for which the model is simulated, then the `edata.getObservedData` will be `nan` but the `rdata.sigmay` for that condition, observable and timepoint will be 0.

This was causing an issue in the direct calculation of the NLLH as we were taking the `log(2 * np.pi * sigma_i**2)` where `sigma_i` was `rdata.sigmay`. This gave `RuntimeWarning: divide by zero encountered in log` errors. 

The fix is simple: mask the `sigma_i` array with the mask for which the data is not `nan`. Already available in `problem.data_mask`.

Was raised in issue #1375

### DIFF
--- a/pypesto/hierarchical/relative/solver.py
+++ b/pypesto/hierarchical/relative/solver.py
@@ -502,7 +502,7 @@ class NumericalInnerSolver(RelativeInnerSolver):
                         f"`{par.inner_parameter_type}`."
                     )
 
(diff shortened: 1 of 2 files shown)


arrjon (approved):
LGTM, thank you!

dweindl (approved):
:+1: thx

Issue #1325, opened 2024-03-11
Easy switching between likelihood and posterior


arrjon:
**Feature description**
Depending on the context, the pypesto objective can be a likelihood or a posterior. It would be nice to have easy acess to both, e.g. if I have defined the objective as the posterior and added `x_priors` to get the likelihood by something like `self.problem.objective(x, priors=False)`. 

**Motivation/Application**
Depending on whether one is using optimisation or sampling, either the likelihood, the posterior or both might be requiered. Beeing able to quickly switch between them would make code more readable, see for instance #1324.

vwiela:
Agreed, one could maybe add a flag or method to the `objective`, to distinguish different kind of objectives. 

Additionally, it seems like that naming and comments in some example notebooks needs some revision then as well.

dweindl:
Independent of switching between types, it would be great to know what type of objective one is dealing with. Then things like profile likelihoods should refuse to work with random objective functions. 

> Agreed, one could maybe add a flag or method to the `objective`, to distinguish different kind of objectives.

:+1:

Either some flag, or something like `PosteriorObjective(ObjectiveBase)` that has a `.getNegLogLikelihood()`.

PR #1378, opened 2024-04-24
Calling prior in sampling with fixed parameters does not work

arrjon:
The `NegLogParameterPriors` objective does not properly account for fixed parameters when invoked during sampling. When it's integrated into the `AggregatedObjective`, optimization functions correctly; however, issues arise during sampling because `call_unprocessed` is triggered with the reduced parameter vector, causing a failure.

Maybe there is a better way to fix this, as I just updated the call of `neglogprior`, but I feel this might break stuff during optimzation.


arrjon:
I reverted the inital fix since I found a better solution. If you now call `problem.x_priors` it behaves exactly as `problem.objective`, which is the behaviour I would expect.

dilpath (approved):
Fine for me for now :+1:

I imagine this will be obsolete if #1325 is resolved by implementing objective type flags.

vwiela (approved):
Thanks for spotting. 

Think this is a good fix to it.

arrjon:
After discussion, @PaulJonasJost and I removed the deepcopy of the class. Otherwise, copying the prior does not work. This happens in the parallel tempering sampler, when the prior is copied to the different chains.

PR #1379, opened 2024-04-25
Require and test python>=3.10

vwiela:
As seen in #1378 petab tests are failing due to use of syntax in AMICI that is not supported in python version <=3.9.

I guess, we should think about dropping the python support for python <=3.9. 
Other packages as numpy already dropped support for 3.9 as well.
https://numpy.org/neps/nep-0029-deprecation_policy.html

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -366,7 +366,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.12']
+        python-version: ['3.11']
 
     steps:
     - name: Check out repository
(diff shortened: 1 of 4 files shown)


PaulJonasJost (approved):
LGTM but also added a commit so probably should not be last one to approve.

.github/workflows/ci.yml
```diff
@@ -226,7 +226,7 @@ jobs:
     strategy:
       matrix:
         # ipopt does not work on 3.9 (https://github.com/mechmotum/cyipopt/issues/225)
```
dweindl, line 228:
```suggestion
```

PR #1385, opened 2024-05-03
Update `test_thermodynamic_integration`

PaulJonasJost:
adjusted `test_thermodynamic_integration`, by lowering the number of chains by factor of 5 and only running 10 chains, as the comparison of more chains is not something that we should have to test imo.

In local tests (Mac) it went from ~200sek to 20


arrjon:
True, this is not neceassary as a test. It was included to check convergence of the evidence estimation. Okay for me to remove it to speed up testing time.

arrjon (approved):

Doresic (approved):

dilpath (approved):

vwiela (approved):

Issue #1383, opened 2024-05-02
Frequent timeouts for `mac (3.11)` workflow runs


dweindl:
There are currently frequent timeouts for `mac (3.11)` workflow runs. 

The by far [slowest](https://github.com/ICB-DCM/pyPESTO/actions/runs/8782850154/job/24097808694#step:6:240) test is:
 `239.30s call     test/sample/test_sample.py::test_thermodynamic_integration`

Can this be reduced to <1min?

PaulJonasJost:
it reliably times out at `test/profile/test_validation_intervals.py` from what I could see. will have a look 👍🏼

dweindl:
> The by far [slowest](https://github.com/ICB-DCM/pyPESTO/actions/runs/8782850154/job/24097808694#step:6:240) test is:
> `239.30s call test/sample/test_sample.py::test_thermodynamic_integration`

I take that back. This, as well as the timeouts in general, was a consequence of the issue addressed in #1388. 
After fixing that, the duration is okayish:
`35.52s call     test/sample/test_sample.py::test_thermodynamic_integration`

PaulJonasJost:
I think this should be fixed right now.

PR #1388, opened 2024-05-03
Fix `pypesto.sample.geweke_test.spectrum` for nfft<=3

dweindl:
For `nfft<=3`, this function computed `k` by converting infinity to int64.

For the last [4 years](https://github.com/ICB-DCM/pyPESTO/commit/fdbacb49df944051cfccf5116a1de30da9365747), this seemed to not cause any trouble, because `np.float64("inf").astype(int)` yielded -9223372036854775808 and the loop did not run. However, with the current `macos-14` GitHub runner, this yields 9223372036854775807, and accordingly, rather long computation times (this is what caused #1383).

I am not really sure what caused this change. macos-12 vs macos-14, arm64 vs x86, ...? all should be IEEE 754-compliant. Maybe some different integer type? :man_shrugging:

With this change `macos-14` runners yield the same results as the other runners. However, I am not familiar with computing power spectral density and I don't know if this is indeed what should happen. It would be great if somebody more familiar with that topic could double-check. (Also, can't we use [scipy.signal.welch](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.welch.html#scipy.signal.welch) here?)

### DIFF
--- a/pypesto/sample/geweke_test.py
+++ b/pypesto/sample/geweke_test.py
@@ -46,7 +46,11 @@ def spectrum(x: np.ndarray, nfft: int = None, nw: int = None) -> np.ndarray:
         n = nw
 
     # Number of windows
-    k = np.floor((n - n_overlap) / (nw - n_overlap)).astype(int)
+    k = (
+        np.floor((n - n_overlap) / (nw - n_overlap)).astype(int)
+        if nw != n_overlap
+        else 0
+    )
     index = np.arange(nw)
     # Normalizing scale factor
     kmu = k * np.linalg.norm(w) ** 2


pypesto/sample/geweke_test.py
```diff
@@ -46,7 +46,11 @@ def spectrum(x: np.ndarray, nfft: int = None, nw: int = None) -> np.ndarray:
         n = nw
 
     # Number of windows
-    k = np.floor((n - n_overlap) / (nw - n_overlap)).astype(int)
+    k = (
+        np.floor((n - n_overlap) / (nw - n_overlap)).astype(int)
+        if nw != n_overlap
```
dilpath, line 51: When is `nw == n_overlap`? Is the total sample size < 100? If so, then I guess this could be handled earlier, since it looks like a vector of zeroes will be returned if `k == 0`. Fine I guess but, if this is an estimate of the variance, then it probably breaks the z-score calculation at line 158.

dilpath (commented):
> Also, can't we use [scipy.signal.welch](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.welch.html#scipy.signal.welch) here?

I don't see an issue, the defaults there compare closely to what's implemented here, as far as I can tell.

pypesto/sample/geweke_test.py
dweindl: > When is `nw == n_overlap`?

* If `len(x) <= 3` and `nfft <= 3` and `nw == None`, for the current usages in pypesto, the first condition is sufficient
* If `len(x) <= 3` and `nfft == None` and `nw == None`
* If `nw < 2`

> this could be handled earlier

Yes.

I don't want to touch that more than necessary right now, but if somebody else feels like cleaning that up (or replacing it by the scipy implementation right away), there is certainly room for improvement.

I would merge this change rather soon, since it at least solves the ridiculously long looping and maintains the old behavior as experienced on most systems.

> Fine I guess but, if this is an estimate of the variance, then it probably breaks the z-score calculation at line 158.

That would have been the case since 2020 then. Would be great if somebody could take a deeper look.

dilpath (approved):
> I would merge this change rather soon, since it at least solves the ridiculously long looping and maintains the old behavior as experienced on most systems.

Sounds good, and thanks!

pypesto/sample/geweke_test.py
dweindl: > looks like a vector of zeroes will be returned if `k == 0`. Fine I guess but, if this is an estimate of the variance, then it probably breaks the z-score calculation at line 158.

From everything I've seen, it either returns `[]`, which is just skipped in `spectrum0`, or it returns `[nan]`, which will make `calculate_zscore` return `nan`.

PR #1390, opened 2024-05-05
Fix in doc of visualise.optimization_stats

PaulJonasJost:
Fixed doc example from generic call (that did not exists) to apropriate call


dweindl (approved):
Thx. 

Already better than before, but maybe you also want to mark the examples as python code?

stephanmg:
Thanks @PaulJonasJost - with code blocks looks now good to me.

stephanmg (approved):

PR #1391, opened 2024-05-05
remove random.seed in favor of np.random.seed

PaulJonasJost:
In some Jupyter notebooks, random.seed was used. This however guarantees no reproducibility as the start points are generated with numpy.


Doresic (approved):
Nice catch :D

PR #1393, opened 2024-05-05
More detailed defaults for `problem.get_full_vector`

PaulJonasJost:
As the doctoring already explained before, we usually want `problem.x_fixed_vals` as defaults if we update a vector x. I think this should be a default and leaving the nans should be left for gradients (hence another parameter to distinguish the two.

### DIFF
--- a/pypesto/problem/base.py
+++ b/pypesto/problem/base.py
@@ -345,11 +345,9 @@ class Problem:
         x: array_like, shape=(dim,)
             The vector in dimension dim.
         x_fixed_vals: array_like, ndim=1, optional
-            The values to be used for the fixed indices. If None, then nans are
-            inserted in case of grad and problem.x_fixed_vals is used in case
-            of x.
+            The values to be used for the fixed indices. If None and x_is_grad=False, problem.x_fixed_vals is used; for x_is_grad=True, nans are inserted.
         x_is_grad: bool
-            Whether the vector is a grad or x.
+            If true, x is treated as gradients.
         """
         if x is None:
             return None
--- a/pypesto/problem/base.py
+++ b/pypesto/problem/base.py
@@ -332,7 +332,10 @@ class Problem:
         self.normalize()
 
     def get_full_vector(
-        self, x: Union[np.ndarray, None], x_fixed_vals: Iterable[float] = None
+        self,
+        x: Union[np.ndarray, None],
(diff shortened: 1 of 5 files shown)


Doresic (approved):
Looks good.
Makes more sense for the default to grab problem.x_fixed_vals instead of not using them.

pypesto/problem/base.py
```diff
@@ -343,8 +346,10 @@ def get_full_vector(
             The vector in dimension dim.
         x_fixed_vals: array_like, ndim=1, optional
             The values to be used for the fixed indices. If None, then nans are
-            inserted. Usually, None will be used for grad and
-            problem.x_fixed_vals for x.
+            inserted in case of grad and problem.x_fixed_vals is used in case
+            of x.
+        x_is_grad: bool
+            Whether the vector is a grad or x.
```
m-philipps, line 352:
```suggestion
            If true, x is treated like gradients; if false (default) x is treated as parameters.
```

pypesto/problem/base.py
```diff
@@ -343,8 +346,10 @@ def get_full_vector(
             The vector in dimension dim.
         x_fixed_vals: array_like, ndim=1, optional
             The values to be used for the fixed indices. If None, then nans are
-            inserted. Usually, None will be used for grad and
-            problem.x_fixed_vals for x.
+            inserted in case of grad and problem.x_fixed_vals is used in case
+            of x.
```
m-philipps, line 350:
```suggestion
            The values to be used for the fixed indices. If None and x_is_grad=False, problem.x_fixed_vals is used; for x_is_grad=True, nans are inserted.
```

pypesto/problem/base.py
```diff
@@ -343,8 +346,10 @@ def get_full_vector(
             The vector in dimension dim.
         x_fixed_vals: array_like, ndim=1, optional
             The values to be used for the fixed indices. If None, then nans are
```
m-philipps, line 348:
```suggestion
```

m-philipps (approved):
Thank you! I left an optional suggestion to extend on the docstring.

pypesto/problem/base.py
PaulJonasJost:
```suggestion
            If true, x is treated as gradients.
```

PR #1395, opened 2024-05-06
Hdf5AmiciHistory with AggregatedObjective

plakrisenko:
In case AggregatedObjective contains AmiciObjective and prior, use Hdf5AmiciHistory type for histories.

### DIFF
--- a/pypesto/objective/aggregated.py
+++ b/pypesto/objective/aggregated.py
@@ -134,7 +134,7 @@ class AggregatedObjective(ObjectiveBase):
     def create_history(
         self, id: str, x_names: Sequence[str], options: HistoryOptions
     ):
-        amici_objectives = [obj for obj in self.objective._objectives if
+        amici_objectives = [obj for obj in self._objectives if
                             isinstance(obj, AmiciObjective)]
         if len(amici_objectives) == 1:
             # aggregate objective because of parameter priors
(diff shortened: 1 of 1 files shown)


pypesto/objective/aggregated.py
```diff
@@ -129,6 +131,18 @@ def get_config(self) -> dict:
             info[f"objective_{n_obj}"] = obj.get_config()
         return info
 
+    def create_history(
+        self, id: str, x_names: Sequence[str], options: HistoryOptions
+    ):
+        """See `history.generate.create_history` documentation."""
+        amici_objectives = [obj for obj in self._objectives if
+                            isinstance(obj, AmiciObjective)]
+        if len(amici_objectives) == 1:
```
PaulJonasJost, line 140:
```suggestion
        if not amici_objectives
```
Otherwise you don't cover the cases where we have multiple amici objectives?

PaulJonasJost: Additionally (just hypothetical, what if we have mixtures of both amici and non amici objectives, will there be any clashes?

pypesto/objective/aggregated.py
```diff
@@ -129,6 +131,18 @@ def get_config(self) -> dict:
             info[f"objective_{n_obj}"] = obj.get_config()
         return info
 
+    def create_history(
+        self, id: str, x_names: Sequence[str], options: HistoryOptions
+    ):
```
PaulJonasJost, line 136:
```suggestion
    ) -> HistoryBase:
```
+ import above

PaulJonasJost (approved):
Thanks a lot :)

dweindl (changes requested):
This will only create a history for a sub-objective and is probably not what a user will expect.

Issue #1397, opened 2024-05-07
`pyjulia` is no longer maintained


m-philipps:
The Julia objective makes use of the `pyjulia` package to call Julia code from Python. According to their website, "Ongoing development of the Python/Julia interface has transitioned to [PythonCall.jl/juliacall](https://github.com/JuliaPy/PythonCall.jl)". The workarounds that are required for `pyjulia` are not necessary with `juliacall`.

PR #1398, opened 2024-05-07
Update julia objective

m-philipps:
- Switch from no longer maintained `pyjulia` to `juliacall` package (#1397)
- Update example; switch to using `ForwardDiff`


sebapersson:
Does this PR also account for the interface to PEtab.jl?

m-philipps:
No, so far only the petab objective.

PR #1400, opened 2024-05-13
Fix JaxObjective

GiacomoFabrini:
- Add `pre_post_processor.setter`
- Add `base_objective` wrapper functions (value / value and grad) to `JaxObjective`


FFroehlich (approved):

pypesto/objective/jax/base.py
```diff
@@ -42,7 +54,9 @@ def _device_fun(base_objective: ObjectiveBase, x: jnp.array):
         jax computed input array.
     """
     return jax.pure_callback(
```
PaulJonasJost, line 56: Short documentation on the return structure would be helpful.

test/base/test_objective.py
```diff
@@ -281,8 +294,11 @@ def _fun(y, pypesto_fun, jax_fun_in, jax_fun_out):
         # can't use rtol = 1e-8 for 32bit
         rtol = 1e-16 if enable_x64 else 1e-4
         for x, rref, rj in zip(xx, rvals_ref, rvals_jax):
+            assert isinstance(rj, jnp.ndarray)
             if max_sensi_order == 0:
-                np.testing.assert_allclose(rref, rj, atol=atol, rtol=rtol)
+                np.testing.assert_allclose(
+                    rref, float(rj), atol=atol, rtol=rtol
+                )
             if max_sensi_order == 1:
```
PaulJonasJost, line 302: You define check_sensi_orders in the JaxObjective (line 146), where you also check that `max(sensi_orders)==0`. Should you change that to a 1?

PaulJonasJost (approved):

pypesto/objective/jax/base.py
FFroehlich: added

test/base/test_objective.py
FFroehlich: good catch, fixed and added test.

PR #1402, opened 2024-05-14
fix #1401

### LINKED ISSUE #1401: Roadrunner petab test case 0018 fails on Mac
Consistently fails. Checking out why in process

PaulJonasJost:
Works with MacOs-13. Could not reproduce with MacOs-14.2 on local computer. Bergen this for now before searching a complete day for a system specific bug for now.


Doresic (approved):
Thanks!

PR #1408, opened 2024-05-31
Added test for multiprocessing of roadrunner.

PaulJonasJost:
Roadrunner after 2.7.0 should be able to handle multiprocessing engine. added a small test. Debatable whether 30 starts is enough/too much, kinda hard to estimate.


Doresic (approved):
Looks good!
30 starts is probably too much. I can imagine that if there are any problems, they would come up already with a few of starts. But if it's fast it doesn't really matter. 

Btw, we have 15 procs (cores) available in testing?

Question more for my understanding: pyPESTO manages the multiprocessing, no? So what does Roadrunner have to support to be able to be ran with multiprocessing?

PR #1409, opened 2024-06-04
Visualize: visualization of estimated observable mapping

Doresic:
I implemented a visualization routine to visualize all estimated observable mappings: linear scaling+offset for relative data and splines for semiquantitative data. The visualization of estimated splines already existed, so I made the linear one and combined them into a single routine. However, I made it so that each can also be called separately.

To make the implementation easier, I had to expand some classes connected to the relative data: inner problem, inner parameter.

During the implementation, a few bugs popped up. I fixed those here as well. I'll comment on the PR so it's easier to distinguish.

The notebooks have also been updated accordingly.

### DIFF
--- a/pypesto/hierarchical/base_problem.py
+++ b/pypesto/hierarchical/base_problem.py
@@ -193,6 +193,11 @@ class AmiciInnerProblem(InnerProblem):
             amici.numpy.ExpDataView(edata)["observedData"] for edata in edatas
         ]
 
+        # Mask the data using the inner problem mask. This is necessary
+        # because the inner problem is aware of only the data it uses.
+        for i in range(len(data)):
+            data[i][~self.data_mask[i]] = np.nan
+
         if len(self.data) != len(data):
             return False
 
--- a/pypesto/hierarchical/inner_calculator_collector.py
+++ b/pypesto/hierarchical/inner_calculator_collector.py
@@ -35,7 +35,6 @@ from ..C import (
     SPLINE_KNOTS,
     SPLINE_RATIO,
     SRES,
-    InnerParameterType,
     ModeType,
 )
 from ..objective.amici.amici_calculator import AmiciCalculator
@@ -102,6 +101,10 @@ class InnerCalculatorCollector(AmiciCalculator):
         self.inner_calculators: list[
             AmiciCalculator
         ] = []  # TODO make into a dictionary (future PR, together with .hierarchical of Problem)
+
+        self.semiquant_observable_ids = None
+        self.relative_observable_ids = None
+
         self.construct_inner_calculators(
             petab_problem, model, edatas, inner_options
         )
@@ -109,7 +112,6 @@ class InnerCalculatorCollector(AmiciCalculator):
         self.quantitative_data_mask = self._get_quantitative_data_mask(edatas)
 
         self._known_least_squares_safe = False
-        self.semiquant_observable_ids = None
 
     def initialize(self):
         """Initialize."""
@@ -137,6 +139,9 @@ class InnerCalculatorCollector(AmiciCalculator):
                 inner_problem=relative_inner_problem
             )
             self.inner_calculators.append(relative_inner_solver)
+            self.relative_observable_ids = (
+                relative_inner_problem.get_relative_observable_ids()
+            )
 
         if ORDINAL in self.data_types or CENSORED in self.data_types:
             optimal_scaling_inner_options = {
@@ -178,12 +183,9 @@ class InnerCalculatorCollector(AmiciCalculator):
                 semiquant_problem.get_noise_dummy_values(scaled=True)
             )
             self.inner_calculators.append(semiquant_calculator)
-            self.semiquant_observable_ids = [
-                model.getObservableIds()[group - 1]
-                for group in semiquant_problem.get_groups_for_xs(
-                    InnerParameterType.SPLINE
-                )
-            ]
+            self.semiquant_observable_ids = (
+                semiquant_problem.get_semiquant_observable_ids()
+            )
 
         if self.data_types - {
             RELATIVE,
--- a/pypesto/hierarchical/relative/parameter.py
+++ b/pypesto/hierarchical/relative/parameter.py
@@ -1,0 +1,44 @@
+import logging
+
+from ...C import InnerParameterType
+from ..base_parameter import InnerParameter
+
+logger = logging.getLogger(__name__)
+
+
+class RelativeInnerParameter(InnerParameter):
+    """A relative (inner) parameter of the relative hierarchical optimization problem.
+
+    Attributes
+    ----------
+    observable_id:
+        The id of the observable the relative inner parameter is connected to.
+    group:
+        Group index. Corresponds to `amici_model.index(observable_id)` + 1.
+    """
+
+    def __init__(
+        self,
+        *args,
+        observable_id: str = None,
+        group: int = None,
+        **kwargs,
+    ):
+        """Construct.
+
+        Parameters
+        ----------
+        See class attributes.
+        """
+        super().__init__(*args, **kwargs)
+        if self.inner_parameter_type not in [
+            InnerParameterType.SCALING,
+            InnerParameterType.OFFSET,
+            InnerParameterType.SIGMA,
+        ]:
+            raise ValueError(
+                "For the RelativeParameter class, the parameter type has to be scaling, offset or sigma."
+            )
+
+        self.observable_id = observable_id
+        self.group = group
--- a/pypesto/hierarchical/relative/parameter.py
+++ b/pypesto/hierarchical/relative/parameter.py
@@ -11,17 +11,18 @@ class RelativeInnerParameter(InnerParameter):
 
     Attributes
     ----------
-    observable_id:
-        The id of the observable the relative inner parameter is connected to.
-    group:
-        Group index. Corresponds to `amici_model.index(observable_id)` + 1.
+    observable_ids:
+        A list of IDs of the observables that the relative inner parameter is connected to.
+        Can be a single or multiple observables.
+    observable_indices:
+        A list of indices of the observables that the relative inner parameter is connected to.
     """
 
     def __init__(
         self,
         *args,
-        observable_id: str = None,
-        group: int = None,
+        observable_ids: list[str] = None,
+        observable_indices: list[int] = None,
         **kwargs,
     ):
         """Construct.
@@ -40,5 +41,5 @@ class RelativeInnerParameter(InnerParameter):
                 "For the RelativeParameter class, the parameter type has to be scaling, offset or sigma."
             )
 
-        self.observable_id = observable_id
-        self.group = group
+        self.observable_ids = observable_ids
+        self.observable_indices = observable_indices
--- a/pypesto/hierarchical/relative/problem.py
+++ b/pypesto/hierarchical/relative/problem.py
@@ -2,7 +2,6 @@
 
 import logging
 
-import numpy as np
 import pandas as pd
 
 from ...C import (
@@ -11,12 +10,12 @@ from ...C import (
     SEMIQUANTITATIVE,
     InnerParameterType,
 )
-from ..base_parameter import InnerParameter
 from ..base_problem import (
     AmiciInnerProblem,
     _get_timepoints_with_replicates,
     ix_matrices_from_arrays,
 )
+from .parameter import RelativeInnerParameter
 
 try:
     import amici
@@ -66,43 +65,35 @@ class RelativeInnerProblem(AmiciInnerProblem):
             petab_problem, amici_model, edatas
         )
 
-    def check_edatas(self, edatas: list[amici.ExpData]) -> bool:
-        """Check for consistency in data.
+    def get_relative_observable_ids(self) -> list[str]:
+        """Get unique list of IDs of all relative observables with scaling and/or offset."""
+        return list(
+            {
+                x.observable_id
+                for x in self.xs.values()
+                if x.inner_parameter_type
+                in [
+                    InnerParameterType.SCALING,
+                    InnerParameterType.OFFSET,
+                ]
+            }
+        )
 
-        Currently only checks for the actual data values. e.g., timepoints are
-        not compared.
+    def get_groups_for_xs(self, inner_parameter_type: str) -> list[int]:
+        """Get unique list of ``RelativeParameter.group`` values."""
+        groups = [x.group for x in self.get_xs_for_type(inner_parameter_type)]
+        return list(set(groups))
 
-        Parameters
-        ----------
-        edatas:
-            A data set. Will be checked against the data set provided to the
-            constructor.
-
-        Returns
-        -------
-        Whether the data sets are consistent.
-        """
-        # TODO replace but edata1==edata2 once this makes it into amici
-        #  https://github.com/AMICI-dev/AMICI/issues/1880
-        data = [
-            amici.numpy.ExpDataView(edata)["observedData"] for edata in edatas
-        ]
-
-        if len(self.data) != len(data):
-            return False
-
-        for data0, data1 in zip(self.data, data):
-            if not np.array_equal(data0, data1, equal_nan=True):
-                return False
-
-        return True
+    def get_xs_for_group(self, group: int) -> list[RelativeInnerParameter]:
+        r"""Get ``RelativeParameter``\s that belong to the given group."""
+        return [x for x in self.xs.values() if x.group == group]
 
 
 def inner_problem_from_petab_problem(
     petab_problem: "petab.Problem",
     amici_model: "amici.Model",
     edatas: list["amici.ExpData"],
-) -> AmiciInnerProblem:
+) -> RelativeInnerProblem:
     """
     Create inner problem from PEtab problem.
 
@@ -128,9 +119,11 @@ def inner_problem_from_petab_problem(
     # matrixify
     ix_matrices = ix_matrices_from_arrays(ixs, data)
 
-    # assign matrices
+    # assign matrices, group and observable id to inner parameters
     for par in inner_parameters:
         par.ixs = ix_matrices[par.inner_parameter_id]
+        par.group = ixs[par.inner_parameter_id][0][2] + 1
+        par.observable_id = amici_model.getObservableIds()[par.group - 1]
 
     par_group_types = {
         tuple(obs_pars.split(";")): (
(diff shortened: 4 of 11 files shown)


pypesto/hierarchical/base_problem.py
```diff
@@ -193,6 +193,11 @@ def check_edatas(self, edatas: list[amici.ExpData]) -> bool:
             amici.numpy.ExpDataView(edata)["observedData"] for edata in edatas
         ]
 
+        # Mask the data using the inner problem mask. This is necessary
+        # because the inner problem is aware of only the data it uses.
+        for i in range(len(data)):
+            data[i][~self.data_mask[i]] = np.nan
+
```
Doresic, line 200: This is a bug fix. It caused an issue only if the model contained multiple non-quantitative data types (e.g. semiquantitative and relative). In that case, this check always failed, as we were not masking the data in the check in the same way it was masked in the inner problem initialization.

pypesto/hierarchical/inner_calculator_collector.py
```diff
@@ -102,14 +101,17 @@ def __init__(
...
+        self.relative_observable_ids = None
+
         self.construct_inner_calculators(
             petab_problem, model, edatas, inner_options
         )
 
         self.quantitative_data_mask = self._get_quantitative_data_mask(edatas)
 
         self._known_least_squares_safe = False
-        self.semiquant_observable_ids = None
```
Doresic, line 112: Bugfix.
`semiquant_observable_ids ` would be set to the right value in `construct_inner_calculators` and then was always overwritten to None in line 112.

pypesto/hierarchical/relative/parameter.py
Doresic, line 1: Extended the relative inner parameter to contain the `observable_id` and `group`. The `group` basically refers to observable_idx + 1. The naming comes from the ordinal implementation.

pypesto/hierarchical/relative/problem.py
```diff
@@ -66,43 +65,35 @@ def from_petab_amici(
             petab_problem, amici_model, edatas
         )
 
-    def check_edatas(self, edatas: list[amici.ExpData]) -> bool:
```
Doresic, line 69: Check edatas was re-defined completely the same for no reason. It is already defined in AmiciInnerProblem.

pypesto/hierarchical/relative/problem.py
```diff
@@ -66,43 +65,35 @@ def from_petab_amici(
...
-                return False
+    def get_groups_for_xs(self, inner_parameter_type: str) -> list[int]:
+        """Get unique list of ``RelativeParameter.group`` values."""
+        groups = [x.group for x in self.get_xs_for_type(inner_parameter_type)]
+        return list(set(groups))
 
-        return True
+    def get_xs_for_group(self, group: int) -> list[RelativeInnerParameter]:
+        r"""Get ``RelativeParameter``\s that belong to the given group."""
+        return [x for x in self.xs.values() if x.group == group]
```
Doresic, line 89: Ease of life class methods.

pypesto/hierarchical/semiquantitative/problem.py
```diff
@@ -147,6 +147,16 @@ def get_interpretable_x_ids(self) -> list[str]:
...
+    def get_semiquant_observable_ids(self) -> list[str]:
+        """Get the ids of semiquantitative observables."""
+        return list(
+            {
+                x.observable_id
+                for x in self.xs.values()
+                if x.inner_parameter_type == InnerParameterType.SPLINE
+            }
+        )
+
```
Doresic, line 159: Some more ease of life

pypesto/visualize/observable_mapping.py
```diff
@@ -37,10 +41,346 @@
...
+        )
+
+    # Plot the estimated spline approximations for semi-quantitative observables.
+    if n_semiquant_observables > 0:
+        axes = plot_splines_from_pypesto_result(
+            pypesto_result=pypesto_result,
+            start_index=start_index,
+            axes=axes,
+            rel_and_semiquant_obs_indices=rel_and_semiquant_obs_indices,
+        )
```
Doresic, line 138: The main routine `visualize_estimated_observable_mapping` calls the linear mapping and spline one if there are relative or semiquantitative observables, respectively. However, the two lower-level routines can also be called directly.

The same axes object is being passed around and plotted on.

Doresic:
Figure_1
Boehm Example with one semiquantitative and 2 relative observables

pypesto/hierarchical/relative/parameter.py
```diff
@@ -0,0 +1,44 @@
...
+logger = logging.getLogger(__name__)
+
+
+class RelativeInnerParameter(InnerParameter):
+    """A relative (inner) parameter of the relative hierarchical optimization problem.
+
+    Attributes
+    ----------
+    observable_id:
+        The id of the observable the relative inner parameter is connected to.
```
dilpath, line 15:
```suggestion
        The ID of the observable that the relative inner parameter is connected to.
```

pypesto/hierarchical/relative/problem.py
```diff
@@ -66,43 +65,35 @@ def from_petab_amici(
...
-        # TODO replace but edata1==edata2 once this makes it into amici
-        #  https://github.com/AMICI-dev/AMICI/issues/1880
-        data = [
-            amici.numpy.ExpDataView(edata)["observedData"] for edata in edatas
-        ]
-
-        if len(self.data) != len(data):
-            return False
+    def get_relative_observable_ids(self) -> list[str]:
+        """Get unique list of IDs of all relative observables with scaling and/or offset."""
```
dilpath, line 69:
```suggestion
        """Get IDs of all unique relative observables with scaling and/or offset."""
```

pypesto/hierarchical/relative/problem.py
```diff
@@ -66,43 +65,35 @@ def from_petab_amici(
...
-            if not np.array_equal(data0, data1, equal_nan=True):
-                return False
+    def get_groups_for_xs(self, inner_parameter_type: str) -> list[int]:
+        """Get unique list of ``RelativeParameter.group`` values."""
+        groups = [x.group for x in self.get_xs_for_type(inner_parameter_type)]
+        return list(set(groups))
 
-        return True
+    def get_xs_for_group(self, group: int) -> list[RelativeInnerParameter]:
+        r"""Get ``RelativeParameter``\s that belong to the given group."""
```
dilpath, line 88: Does this fail?

```suggestion
        """Get ``RelativeParameter``s that belong to the given group."""
```

pypesto/hierarchical/relative/parameter.py
dilpath: Is the naming offset worth keeping? Fine to leave as-is for this PR

pypesto/hierarchical/semiquantitative/problem.py
```diff
@@ -147,6 +147,16 @@ def get_interpretable_x_ids(self) -> list[str]:
             if x.inner_parameter_type == InnerParameterType.SIGMA
         ]
 
+    def get_semiquant_observable_ids(self) -> list[str]:
+        """Get the ids of semiquantitative observables."""
```
dilpath, line 151:
```suggestion
        """Get the IDs of semiquantitative observables."""
```

pypesto/visualize/observable_mapping.py
```diff
@@ -37,10 +41,346 @@
...
+
+    Parameters
+    ----------
+    pypesto_result:
+        The pyPESTO result object from optimization.
+    pypesto_problem:
+        The pyPESTO problem.
+    start_index:
+        The index of the optimization run in `result.optimize_result.list`
+        for which to visualize the estimated observable mapping.
```
dilpath, line 64:
```suggestion
        The observable mapping from this start's optimized vector will be plotted.
```

pypesto/visualize/observable_mapping.py
```diff
@@ -37,10 +41,346 @@
...
+
+def plot_linear_observable_mappings_from_pypesto_result(
+    pypesto_result: Result,
+    pypesto_problem: HierarchicalProblem,
+    start_index=0,
+    axes: Optional[plt.Axes] = None,
+    rel_and_semiquant_obs_indices: Optional[list[int]] = None,
+    **kwargs,
+):
+    """Plot the linear observable mappings from a pypesto result.
```
dilpath, line 158:
```suggestion
    """Plot the linear observable mappings from a pyPESTO result.
```

pypesto/visualize/observable_mapping.py
```diff
@@ -37,10 +41,346 @@
...
+    start_index:
+        The index of the `pypesto_result.optimize_result.list` to plot.
+    axes:
+        The axes to plot the linear observable mappings on.
+    rel_and_semiquant_obs_indices:
+        The indices of the relative and semi-quantitative observables in the
+        amici model. Important if both relative and semi-quantitative observables
+        will be plotted. If None, only the relative observables will be plotted.
+    kwargs:
+        Additional arguments to pass to the plotting function.
```
dilpath, line 175:
```suggestion
    **kwargs:
        Additional arguments to pass to the ``matplotlib.pyplot.subplots`` function.
```

pypesto/visualize/observable_mapping.py
```diff
@@ -37,10 +41,346 @@
...
+    inner_rdatas = amici.runAmiciSimulations(
+        amici_model,
+        amici_solver,
+        edatas,
+        num_threads=min(n_threads, len(edatas)),
+    )
+
+    # If any amici simulation failed, raise warning and return None.
+    if any(rdata.status != amici.AMICI_SUCCESS for rdata in inner_rdatas):
+        warnings.warn(
```
dilpath, line 275: Seems like an exception rather than a warning?

pypesto/visualize/observable_mapping.py
```diff
@@ -37,10 +41,346 @@
...
+        offset_value = (
+            inner_parameter_values[offset.inner_parameter_id]
+            if offset is not None
+            else 0
+        )
+
+        # Get the data mask for the current observable.
+        observable_data_mask = (
+            scaling_factor.ixs if scaling_factor is not None else offset.ixs
+        )
```
dilpath, line 335: Is the logic here that the scaling indices are a subset of the offset indices? Or are they equal or `None`, and you just need it from one of them?

pypesto/visualize/observable_mapping.py
```diff
@@ -37,10 +41,346 @@
...
+        simulation = extract_expdata_using_mask(
+            expdata=sim, mask=observable_data_mask
+        )
+
+        # min_simulation = np.min(simulation)
+        # min_measurement = np.min(measurements)
+
+        # Plot the simulations on the x-axis and the measurements on the y-axis.
+        # ax.plot(simulation, np.full(len(simulation), min_measurement), "go", label="Model output")
+        # ax.plot(np.full(len(measurements), min_simulation), measurements, "bs", label="Measurements")
```
dilpath, line 352:
```suggestion
```

pypesto/visualize/observable_mapping.py
```diff
@@ -37,10 +41,346 @@
...
+            label="Linear mapping",
+        )
+
+        ax.legend()
+        ax.set_title(f"Observable {observable_id}")
+        ax.set_xlabel("Model output")
+        ax.set_ylabel("Measurements")
+
+        # Store the ax in the axes array.
+        axes[ax_index] = ax
```
dilpath, line 371: Above, `ax = axes[ax_index]` -- no need to overwrite the `axes`, right?

pypesto/visualize/observable_mapping.py
```diff
@@ -184,15 +536,19 @@ def plot_splines_from_inner_result(
         The simulated model output.
     observable_ids:
         The ids of the observables.
+    axes:
+        The axes to plot the estimated spline approximations on.
+    rel_and_semiquant_obs_indices:
+        The indices of the relative and semi-quantitative observables in the
+        amici model. Important if both relative and semi-quantitative observables
+        will be plotted. If None, only the semi-quantitative observables will be plotted.
```
dilpath, line 544: Only semi-quantitative will be plotted in this method regardless, right?

dilpath (approved):
Looks great :+1:

pypesto/hierarchical/relative/parameter.py
```diff
@@ -0,0 +1,44 @@
...
+        if self.inner_parameter_type not in [
+            InnerParameterType.SCALING,
+            InnerParameterType.OFFSET,
+            InnerParameterType.SIGMA,
+        ]:
+            raise ValueError(
+                "For the RelativeParameter class, the parameter type has to be scaling, offset or sigma."
+            )
+
+        self.observable_id = observable_id
```
dweindl, line 43: Is an inner parameter always used for only a single observable? Not completely sure about the current implementation, but generally, they could be shared across observables.

pypesto/visualize/observable_mapping.py
```diff
@@ -66,6 +410,12 @@ def plot_splines_from_pypesto_result(
             "The calculator must be an instance of the InnerCalculatorCollector."
         )
 
+    # Check the result for start index contains the spline knots.
+    if SPLINE_KNOTS not in pypesto_result.optimize_result.list[start_index]:
+        raise ValueError(
+            "The result for the start index does not contain the spline knots."
```
dweindl, line 416:
```suggestion
            f"The result with index {start_index} does not contain the spline knots."
```

dweindl (commented):
Thanks, looks good to me.

The only concern is the one-parameter-to-one-observable mapping.

pypesto/hierarchical/relative/parameter.py
Doresic: Yeah, not sure...
The nicest thing would definitely be to rename it everywhere to observable index, but it is a thing for a separate PR. Didn't want to mix ordinal stuff here.

pypesto/visualize/observable_mapping.py
Doresic: I just need the data mask from one of them. Both of them have it. But also both of them could not exist (in the case of having only scaling or only offset in an observable)

pypesto/hierarchical/relative/problem.py
dilpath: No need, but I guess the `\s` is problematic?

pypesto/visualize/observable_mapping.py
dilpath: Can rewrite this as `scaling_factor.ixs or offset.ixs` then, should do the same thing, but fine as is

Doresic: Ohh true.
I forget about the `or` operator too often

pypesto/visualize/observable_mapping.py
stephanmg: Agreed

pypesto/visualize/observable_mapping.py
Doresic: True, it's ambigious.
Changed to:
`The indices of the relative and semi-quantitative observables in the
        amici model. Important if both relative and semi-quantitative observables
        will be plotted on the same axes.` in both places

pypesto/hierarchical/relative/parameter.py
Doresic: Nope, changed my mind. Changed to `observable_indices`.
Fixing the problem Daniel posed (with relative parameters possibly having multiple observable_ids) it felt more and more ridiculous to continue using groups instead of observable_indices. Especially since relative parameters can be in another group which is determined by the scaling/offset coupling.

pypesto/hierarchical/relative/problem.py
Doresic: Yeah, interestingly the docs test fails with
```
Warning, treated as error:
/home/runner/work/pyPESTO/pyPESTO/pypesto/hierarchical/relative/problem.py:docstring of pypesto.hierarchical.relative.problem.RelativeInnerProblem.get_xs_for_obs_idx:1:Inline literal start-string without end-string.
doc: exit 2 (611.35 seconds) /home/runner/work/pyPESTO/pyPESTO> sphinx-build -W -b html doc/ doc/_build/html pid=2938
  doc: FAIL code 2 (617.84=setup[6.49]+cmd[611.35] seconds)
  evaluation failed :( (618.03 seconds)
Error: Process completed with exit code 2.
```

pypesto/hierarchical/relative/parameter.py
Doresic: Changed to `self.observable_ids` and `self.observable_indices`

pypesto/hierarchical/relative/problem.py
dilpath: Ah OK. These docs suggest `\ s`, not `\s`, but if you found something that works then nevermind!

https://documatt.com/restructuredtext-reference/element/inline-literal.html#requires-whitespace-or-punctuation-around

Doresic: Ohh that's the issue?
Ok, will leave it as is for now, but thanks for checking. Good to have in mind.

dweindl (approved):
:+1:

pypesto/visualize/observable_mapping.py
```diff
@@ -37,27 +41,352 @@
...
+
+    Visualizes the estimated linear mapping for relative observables and the non-linear
+    spline approximation for semi-quantitative observables.
+
+    Parameters
+    ----------
+    pypesto_result:
+        The pyPESTO result object from optimization.
+    pypesto_problem:
+        The pyPESTO problem.
```
m-philipps, line 61: The pypesto problem objective is used downstream and I think that we can have problems without objective so it could be nice to specify here that the `pypesto_problem` is expected to have an objective

m-philipps (approved):
Thanks for the implementation, only two suggestions but no need to address them.

pypesto/visualize/__init__.py
```diff
@@ -19,6 +19,13 @@
 )
 from .ensemble import ensemble_identifiability
 from .misc import process_offset_y, process_result_list, process_y_limits
+from .observable_mapping import (
+    _add_spline_mapped_simulations_to_model_fit,
```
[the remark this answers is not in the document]
Doresic: True, you're right. Don't know why it was there.
Removing it.

pypesto/visualize/observable_mapping.py
Doresic: True.
Added it into the docstring and added some check to check that there is an objective and that it has the correct calculator.

Issue #1410, opened 2024-06-07
Aggregated Objective does not return expected outputs using `return_dict`


arrjon:
**Bug description**
The functions available in `pypesto.visualize.model_fit` cannot work with an `AggregatedObjective` and hence fail.
There are two reasons:
- the amici model is not available by objective.amici_model (I already fixed that issue in the branch `visualize_aggregated` by adding a new property to the `AggregatedObjective`)
- calling `pypesto_problem.objective(x, return_dict=True)` does return empty `rdatas`. (this I could not fix right away)

**Expected behavior**
The visuslization routines should still work properly. 
Also I would expect `pypesto_problem.objective(x, return_dict=True)` to return non empty `rdatas`.
Calling `pypesto_problem.objective._objectives[0](x, return_dict=True)` instead works and does return non empty `rdatas`. 

**To reproduce**
I attached a minimal example using the Blasi model and PeTab. 
[minimal example.zip](https://github.com/user-attachments/files/15741096/minimal.example.zip)

**Environment**
- Operating system: MacOs
- `pypesto` version: branch `visualize_aggregated`
- Python version: python 3.10

arrjon:
maybe @dweindl can help here?

dweindl:
> maybe @dweindl can help here?

I am not too familiar with either `AggregatedObjective` or `pypesto.visualize.model_fit`, but I agree that it should either work, or raise an informative error.

> the amici model is not available by objective.amici_model 

That makes sense, because often `AggregatedObjective` won't have only a single model.

> (I already fixed that issue in the branch `visualize_aggregated` by adding a new property to the `AggregatedObjective`)

I would rather handle `AggregatedObjective` separately in the visualization code instead of adding `amici_model` to `AggregatedObjective`, since there will probably be multiple models.

arrjon:
> I would rather handle `AggregatedObjective` separately in the visualization code instead of adding `amici_model` to `AggregatedObjective`, since there will probably be multiple models.

Okay, I thought that `AggregatedObjective` is usually an `AmiciObjective` and `NegLogPriors` and not multiple models. This makes it rather complicated to write a general visualization function.

dweindl:
I don't know how it's currently used, but one of the goals would be using it with PEtab problems comprising multiple models and I wouldn't add anything that will complicate this.

I think it's nevertheless possible to provide some meaningful visualizations. Just iterate over the objectives and visualize what's possible and give a warning if any of those cannot be handled.

[...]

PR #1411, opened 2024-06-07
Fixing Aggregated Objective Visualisations

arrjon:
Fixes the isses mentioned in #1410 so `AggregatedObjective` can be used in the `visualize.model_fit` function.

### DIFF
--- a/pypesto/visualize/model_fit.py
+++ b/pypesto/visualize/model_fit.py
@@ -18,7 +18,7 @@ from amici.petab.conditions import fill_in_parameters
 from amici.petab.simulations import rdatas_to_simulation_df
 from petab.visualize import plot_problem
 
-from ..C import CENSORED, ORDINAL, SEMIQUANTITATIVE
+from ..C import CENSORED, ORDINAL, RDATAS, SEMIQUANTITATIVE
 from ..petab.importer import get_petab_non_quantitative_data_types
 from ..problem import HierarchicalProblem, Problem
 from ..result import Result
@@ -99,7 +99,7 @@ def visualize_optimized_model_fit(
             )
 
     simulation_df = rdatas_to_simulation_df(
-        objective_result,
+        objective_result[RDATAS],
         amici_model,
         petab_problem.measurement_df,
     )
--- a/pypesto/visualize/model_fit.py
+++ b/pypesto/visualize/model_fit.py
@@ -89,7 +89,8 @@ def visualize_optimized_model_fit(
                 objective_result = objective(
                     pypesto_problem.get_full_vector(
                         x=x, x_fixed_vals=pypesto_problem.x_fixed_vals
-                    )
+                    ),
+                    return_dict=True,
                 )
                 break
         if amici_model is None:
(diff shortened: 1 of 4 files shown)


arrjon:
For the moment I iterate over the objectives in `AggregatedObjectives` to find all amici models and plot them individually. For the case with only one model and a `NegLogPrior` objective this works fine.

However, I did not test the function using a petab problem with multiple amici models. Potentially, one needs to extract information from the petab measurement file specific to the corresponding model.

pypesto/visualize/model_fit.py
```diff
@@ -72,62 +72,106 @@ def visualize_optimized_model_fit(
...
-        simulations_df=simulation_df,
-        **kwargs,
-    )
+    amici_models, objective_results = [], []
 
-    non_quantitative_data_types = get_petab_non_quantitative_data_types(
-        petab_problem
-    )
+    # get amici model
+    if hasattr(pypesto_problem.objective, AMICI_MODEL):
```
dweindl, line 79: What's the advantage of `hasattr` over `isinstance(. , AmiciObjective)`? The latter seems safer and will make it easier to find usages of `AmiciObjective.amici_model`.

pypesto/visualize/model_fit.py
```diff
@@ -72,62 +72,106 @@ def visualize_optimized_model_fit(
...
-        ):
-            axes = plot_categories_from_pypesto_result(
-                result,
-                start_index=start_index,
-                axes=axes,
+        objective_results.append(
+            pypesto_problem.objective(x, return_dict=True)
+        )
+        amici_models.append(pypesto_problem.objective.amici_model)
+    elif hasattr(pypesto_problem.objective, "_objectives"):
```
dweindl, line 88: As above - if there is no reason against it, I'd prefer `isinstance`.

pypesto/visualize/model_fit.py
```diff
@@ -72,62 +72,106 @@ def visualize_optimized_model_fit(
...
+        # iterate over objectives to find the amici models in the AggregatedObjective
+        x = result.optimize_result.list[start_index]["x"]
+
+        for objective in pypesto_problem.objective._objectives:
+            if hasattr(objective, AMICI_MODEL):
```
dweindl, line 93: `isinstance`?

dweindl, line 93: I'd emit a warning for any non-AmiciObjective encountered here. (Or less preferred add to the documentation that non-AmiciObjective are silently ignored.)

dweindl (commented):
Thanks.

pypesto/visualize/model_fit.py
arrjon: None, I will change that

Doresic (approved):
Looks good to me! Thanks
Might be good to add a test in `test/test_visualize.py` testing a case with an aggregated objective (similar to `test_visualize_optimized_model_fit()` or an extension of it).

PaulJonasJost (approved):

stephanmg (approved):

Issue #1412, opened 2024-06-10
NegLogParameterPriors evaluates prior density also on fixed parameters


arrjon:
**Bug description**
`NegLogParameterPriors` evaluates prior density also on fixed parameters. This should not be the case and might be important in the case of model selection. Currently, this objective is not aware of free and fixed parameters and just evals the prior density for all parameters. In the case, where the nominal value might lie outside of the prior range, this will return always `-inf`, otherwise this will not raise any errors or warnings.

**Expected behavior**
`NegLogParameterPriors` should only evaluate the prior on free parameters. 

**To reproduce**
Define a petab problem with fixed parameters, where these fixed parameters also have a prior defined.

**Possible Solutions**
- remove priors for fixed parameters in your petab file
- make the objective aware of fixed parameters

FFroehlich:
I have to disagree here. If you don't want a prior for a fixed parameter, don't specify one in the parameter table. In other words, in which situations would it be important to evaluate the prior for a fixed parameter?

In general, model selection metrics like AIC and BIC are only applied to the likelihood function, not the posterior and it feels more natural to use sampling based approaches with posteriors.

arrjon:
> I have to disagree here. If you don't want a prior for a fixed parameter, don't specify one in the parameter table. In other words, in which situations would it be important to evaluate the prior for a fixed parameter?


I agree that one should not specify priors for fixed parameters from the beginning. However, you can fix and unfix parameters later, and currently this will be ignored. For example, this can happen when you do model selection, where you fix parameters which were free in another model.



> 
> 
> In general, model selection metrics like AIC and BIC are only applied to the likelihood function, not the posterior and it feels more natural to use sampling based approaches with posteriors.

True, however, if you compute Bayes factors, then the prior matters and also the value of the prior density.

FFroehlich:
> > I have to disagree here. If you don't want a prior for a fixed parameter, don't specify one in the parameter table. In other words, in which situations would it be important to evaluate the prior for a fixed parameter?
> 
> I agree that one should not specify priors for fixed parameters from the beginning. However, you can fix and unfix parameters later, and currently this will be ignored. For example, this can happen when you do model selection, where you fix parameters which were free in another model.

But you could also argue that one can add and remove priors later. From my perspective the user can already specify whether there is a prior on a fixed parameter or not within the context of a petab problem.

[...]

PR #1413, opened 2024-06-10
NegLogParameterPriors does not evaluate prior density at fixed params

arrjon:
Based on #1412 

Making the `NegLogParameterPriors` aware of fixed parameters. The density is not evaluted for those. Moreover, the petab importet is updated to indicate the fixed parameters while creating the prior. 

Question: should this argument stay optional? If priors are not given for fixed parameters, no problem would occure. The problem only occurs when priors are given and parameters fixed.

### DIFF
--- a/pypesto/objective/priors.py
+++ b/pypesto/objective/priors.py
@@ -1,7 +1,7 @@
 import logging
 import math
 from collections.abc import Sequence
-from typing import Callable, Optional, Union
+from typing import Callable, Union
 
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


dilpath (approved):
Thanks, I think it looks fine, but since we have `prior_list`, we can overwrite immediately, instead of recomputing it every time it's accessed.

dilpath:
Thanks! Sorry for the mask suggestion, I wasn't familiar with the code

PaulJonasJost (approved):
Nice, a test for this would be good. Also not entirely sure whether it is relevant, but the problem has a function `fix_parameters`, for this it should be accounted as well? 🤔

arrjon:
After discussing this issue in #1412, we will not make the prior aware of fixed parameters. Instead, users should specifiy priors correctly in the PEtab problem. In this way, we will allow maximum flexibility.

FFroehlich (approved):

PR #1416, opened 2024-06-17
AmiciObjective: check that sensitivities wrt all relevant parameters …

### LINKED ISSUE #1414: AmiciObjective: For sensi_orders>0, check that sensitivities wrt all relevant parameters are computed
Related to https://github.com/AMICI-dev/AMICI/issues/2455

When we have an `AmiciObjective` and need to compute gradients, we should make sure that sensitivities can be computed w.r.t. all free parameters (this needs to be checked when constructing `AmiciObjective`, and whenever additional parameters are ixed). I.e. we need to check whether all non-fixed parameters exist in `amici_model.getParameterIds()`, taking into account `AmiciObjective.parameter_mapping`.

This is particularly important for PEtab import, where we aren't necessarily able to compute sensitivities w.r.t. to all parameters (https://github.com/AMICI-dev/AMICI/pull/2456).

dweindl:
…can be computed

Adds a check whether the amici model in an `AmiciObjective` is able to compute sensitivities w.r.t. to the objective parameters.

Closes #1414

This raised the question of whether an objective can have implicitly fixed parameters. I.e. whether it should be considered a use case, that some objective parameters that may change the objective value, have zero entries in the gradient. This is what's happening if one uses `PetabImporter.create_objective()` without passing it through `PetabImporter.create_problem()`. So far, pypesto is rather unclear on that. Whether this affects optimization will depend on into which `Problem` this objective will be put.

My assumption here is that this is undesirable and that changes in the objective value should be reflected in the gradient.

### DIFF
--- a/pypesto/objective/amici/amici.py
+++ b/pypesto/objective/amici/amici.py
@@ -175,21 +175,6 @@ class AmiciObjective(ObjectiveBase):
             )
         self.parameter_mapping = parameter_mapping
 
-        # Check that we can compute the max_sensi_order sensitivities
-        #  based on the supplied model. If not, raise an error.
-        # If max_sensi_order is not set, set it if necessary
-        if (self.max_sensi_order is None or self.max_sensi_order > 0) and (
-            missing_sensitivities := self._get_missing_sensitivities()
-        ):
-            if self.max_sensi_order is not None:
-                raise ValueError(
-                    f"`max_sensi_order` is {self.max_sensi_order}`, but "
-                    "the provided model does not support computing "
-                    "sensitivities w.r.t. the following parameters: "
-                    f"{missing_sensitivities}."
-                )
-            self.max_sensi_order = 0
-
         # If supported, enable `guess_steadystate` by default. If not
         #  supported, disable by default. If requested but unsupported, raise.
         if (
@@ -256,7 +241,7 @@ class AmiciObjective(ObjectiveBase):
         but aren't available missing.
         """
         required_sensitivities = set()
-        objective_parameters = set(self.x_ids)
+        objective_parameters = set(self.x_names)
         # resolve parameter mapping: collect all model parameters that depend
         #  on the optimization parameters
         for condition_mapping in self.parameter_mapping:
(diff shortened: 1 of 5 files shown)


pypesto/petab/importer.py
```diff
@@ -447,6 +447,9 @@ def create_objective(
         Returns
         -------
         A :class:`pypesto.objective.AmiciObjective` for the model and the data.
+        This object is expected to be passed to
+        :class:`PetabImporter.create_problem` to correctly handle fixed
+        parameters.
```
PaulJonasJost, line 452: Should we have a test to check the expected behavior if one does not pass it to `PetabImporter.create_problem`?

test/petab/test_petab_import.py
```diff
@@ -190,18 +193,24 @@ def test_max_sensi_order():
     )
 
     # fix max_sensi_order to 1
-    objective = problem.create_objective(max_sensi_order=1)
+    objective = importer.create_problem(
+        objective=importer.create_objective(max_sensi_order=1)
+    ).objective
```
PaulJonasJost, line 198: This might be a minor thing, but i find this a really unintuitive way of creating the objective function correctly.

PaulJonasJost (approved):
Thanks, the check looks sound to me, but probably good if @FFroehlich has another look at it.

Conceptual question: I get it right, that if currently (and henceforth) i create the objective **only** with `PetabImporter.create_objective()`, i get an objective back, which will artificially compute gradients/sensitivities for all parameters independent of whether they are fixed or not? artificially in the sense that for fixed parameters it returns 0? So the only difference between passing it/not passing it to .create_problem would be the dimensionality?

pypesto/petab/importer.py
dweindl: The problem is that we don't know the expected behaviour if one does not pass it to `PetabImporter.create_problem`

test/petab/test_petab_import.py
dweindl: The problem is, that petab fixed parameters are only handled in `create_problem`, but not in `create_objective`. Maybe this could/should be changed. Not completely sure what might rely on the current behaviour.

pypesto/objective/amici/amici.py
```diff
@@ -233,6 +234,33 @@ def __init__(
         # `set_custom_timepoints` method for more information.
         self.custom_timepoints = None
 
+    def _get_missing_sensitivities(self) -> set[str]:
+        """Get the model parameters for which sensitivities are missing.
+
+        Get the model parameters w.r.t. which sensitivities are required,
+        but aren't available missing.
```
FFroehlich, line 241:
```suggestion
        but aren't available.
```

test/petab/test_petab_import.py
FFroehlich: well both the importer and the problem should be aware of fixes parameters, so I agree that one should naturally expect `problem.create_objective`, `importer.create_objective` and `importer.create_problem().objective` to be equivalent and that passing `importer.create_objective(max_sensi_order=1)` to `importer.create_problem` isn't necessary.

Issue #1405, opened 2024-05-22
Lazy loading of optimization results from HDF5


dweindl:
Currently, when loading optimization results from an HDF5 file, the complete results are read into memory. For large result sets, this takes pretty long. It would be convenient if the data was only loaded as needed. E.g., I might only want to access objective function values, without loading some thousand Hessians.

PR #1421, opened 2024-06-25
Added a lazy result object.

PaulJonasJost:
Added a result object, that reads the results only if needed. Connected to #1405 

Things to consider still

- [x] Do we want a casing limit, if yes how would that look like?
- [x] Should the lazy result actually be a subclass of the optimizer result?
- [ ] Connected with the above, when printing the lazy result, all attributes are shown as `None`


test/base/test_lazy_result.py
```diff
@@ -0,0 +1,83 @@
...
+                        self.filename, f"optimization/results/{group_name}"
+                    )
+                )
+
+    def test_initialization(self):
+        # Test that the lazy results are initialized correctly
+        for lazy_result in self.lazy_results:
+            self.assertEqual(lazy_result.filename, self.filename)
+            self.assertEqual(lazy_result._data, {})
+            _ = lazy_result.id
```
dweindl, line 44: What is that for?

test/base/test_lazy_result.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def test_read_result_non_lazy(self):
+        # Test reading results without using the lazy loading option
+        result = read_result(self.filename, optimize=True, lazy=False)
+        self.assertIsInstance(result, Result)
+        self.assertTrue(result.optimize_result)
+
+        # Check if the optimize results are instances of OptimizerResult
+        for opt_result in result.optimize_result:
+            self.assertIsInstance(opt_result, OptimizerResult)
```
dweindl, line 83:
```suggestion
            self.assertIsInstance(opt_result, OptimizerResult)
            self.assertNotIsInstance(opt_result, LazyOptimizerResult)            
```
This will also be true with `lazy=True`

dweindl:
>     * [ ]  Do we want a casing limit, if yes how would that look like?

Not sure what you mean.

>     * [ ]  Should the lazy result actually be a subclass of the optimizer result?

I'd say so.

Something to be discussed: Should changes to the result just affect the OptimizerResult instance or should they be written to file?

>     * [ ]  Connected with the above, when printing the lazy result, all attributes are shown as `None`

Why aren't they loaded on demand then?

PaulJonasJost:
> > ```
> Not sure what you mean.

Sorry meant caching. If we load very big hessians, we might want to throw them out later again? 🤔 

> Something to be discussed: Should changes to the result just affect the OptimizerResult instance or should they be written to file?

That's an interesting thought. I would say per default no, as it would not be a result object created by optimization but changed thereafter. But a function would probably be good to propagate changes to the file?

> Why aren't they loaded on demand then?

Not entirely sure, it might be because in the OptimizerResult they are attributes while here they are properties? And so since LazyResult is a subclass, it will get initiated with those attributes as well 🤔 When calling `optimizer_result.id` it gives back the appropriate value, just not in the `__repr__`/`print`

dilpath:
I have been using hdfdict [1] with pydantic objects, no issues so far in my basic use cases (including numpy arrays), which has lazy loading on by default. Hopefully the lazy loading persists when a pydantic data object is reconstructed from HDF5 via hdfdict, but I haven't tested this yet.

So one alternative to this PR would be to refactor our results to be pydantic classes. I guess this might anyway occur when PEtab Result is supported in pyPESTO.

[1] https://github.com/SiggiGue/hdfdict/blob/master/hdfdict/hdfdict.py

test/base/test_lazy_result.py
```diff
@@ -0,0 +1,84 @@
...
+                        self.filename, f"optimization/results/{group_name}"
+                    )
+                )
+
+    def test_initialization(self):
+        # Test that the lazy results are initialized correctly
+        for lazy_result in self.lazy_results:
+            self.assertEqual(lazy_result.filename, self.filename)
+            self.assertEqual(lazy_result._data, {})
+            _ = lazy_result.id
```
PaulJonasJost, line 44:
```suggestion
```

test/base/test_lazy_result.py
PaulJonasJost: removed

PaulJonasJost:
@dilpath as we talked about, I think it makes sense to merge it in and when `petabResult` is out, we might anyways make it obsolete? In that case i think this as an intermediate step should be fine.

dilpath:
Yes, fine for me :+1:

PR #1424, opened 2024-07-02
Fix `return_dict` when an `AmiciObjective` is inside `AggregatedObjective`

dilpath:
`AmiciObjective` previously returned all AMICI RData information when `return_dict=True`. However, this was handled via `__call__`. If the PEtab problem has a prior, then an `AggregatedObjective` is created, with an `AmiciObjective` and the prior inside. The `AggregatedObjective` calls `call_unprocessed` of each internal objective, skipping `AmiciObjective.__call__` and hence one would see something like:
```python
# call AggregatedObjective
>>> result = objective(x, return_dict=True)
>>> result['rdatas'][0].y is None
True
>>> result = objective(x, return_dict=True, amici_reporting=amici.RDataReporting.full)
>>> result['rdatas'][0].y is None
False
```

This PR changes the above to
```python
>>> result = objective(x, return_dict=True)
>>> result['rdatas'][0].y is None
False
```
i.e., rdatas are now "full" and e.g. plotting is now possible.

An alternative fix would be to always pass `return_dict` to `call_unprocessed`. This would be fine for me, but might break someone's custom objective.


pypesto/objective/amici/amici.py
```diff
@@ -458,18 +439,23 @@ def call_unprocessed(
 
         x_dct = self.par_arr_to_dct(x)
 
-        # only ask amici to compute required quantities
         amici_reporting = (
             self.amici_reporting
             if amici_reporting is None
             else amici_reporting
         )
         if amici_reporting is None:
```
PaulJonasJost, line 447: It should be made clear in the documentation that amici_reporting takes precedence over return_dict imo.

pypesto/objective/base.py
```diff
@@ -178,6 +183,8 @@ def __call__(
         x_full = self.pre_post_processor.preprocess(x=x)
 
         # compute result
+        if self.share_return_dict:
+            kwargs["return_dict"] = return_dict
```
PaulJonasJost, line 187: What if  `return_dict` is in kwargs already?

pypesto/objective/aggregated.py
```diff
@@ -113,6 +116,9 @@ def call_unprocessed(
                 "The length of `kwargs_list` must match the number of "
                 "objectives you are aggregating."
             )
+        for objective_, objective_kwargs in zip(self._objectives, kwargs_list):
+            if objective_.share_return_dict:
+                objective_kwargs["return_dict"] = return_dict
```
PaulJonasJost, line 121: Same as in the base objective (see above or below comment)

PaulJonasJost (approved):
A test for this would be good as well ensuring it works as intended?

pypesto/objective/base.py
dilpath: It cannot be, since `return_dict` is it's own kwarg in this method's argument list.

pypesto/objective/base.py
```diff
@@ -178,6 +183,8 @@ def __call__(
         x_full = self.pre_post_processor.preprocess(x=x)
 
         # compute result
+        if self.share_return_dict:
```
FFroehlich, line 186: why is this necessary?

pypesto/objective/aggregated.py
FFroehlich: I would actually argue that the outer `return_dict` should take precedence over inner `return_dict` and that we don't need the `if objective_.share_return_dict:` check at all.

FFroehlich (approved):

dweindl:
> An alternative fix would be to always pass `return_dict` to `call_unprocessed`. This would be fine for me, but might break someone's custom objective.

I think this would make sense. Your current implementation would already enable issuing a DeprecationWarning and updating things accordingly.

dweindl (approved):
Thx

pypesto/objective/base.py
dilpath: Otherwise there might be some argument error, because `return_dict` is not in every custom user `objective.call_unprocessed` signature. But I'll switch to Daniel's suggestion now

stephanmg (approved):
Looks fine, many thanks for providing this fix.

dilpath:
> > An alternative fix would be to always pass `return_dict` to `call_unprocessed`. This would be fine for me, but might break someone's custom objective.
> 
> I think this would make sense. Your current implementation would already enable issuing a DeprecationWarning and updating things accordingly.

Thanks, I decided to simply switch to it by default now, with a `DeprecationWarning`.

PR #1426, opened 2024-07-03
Hierarchical: avoid recomputing inner parameters if simulation failed

dilpath:
`result.x = None` if simulation fails at the startpoint. `result.x` is used to compute hierarchical inner parameters after optimization. When `result.x = None`, an AMICI simulation to compute inner parameters is still run, unexpectedly... I didn't look deeper there, but this PR stops the computation of inner parameters if simulation fails at the startpoint (defined by `result.x = None`).

This PR means users do not see `NaN in p` AMICI  simulation errors due to `result.x = None`. This can be confusing for users, who then try (and fail) to reproduce those errors with `objective(result.x0)`.

### DIFF
--- a/pypesto/objective/amici/amici.py
+++ b/pypesto/objective/amici/amici.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import abc
 import copy
 import os
@@ -5,7 +7,7 @@ import tempfile
 from collections import OrderedDict
 from collections.abc import Sequence
 from pathlib import Path
-from typing import TYPE_CHECKING, Optional, Union
+from typing import TYPE_CHECKING, Union
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


pypesto/objective/amici/amici.py
```diff
@@ -83,7 +83,9 @@ def __init__(
         n_threads: Optional[int] = 1,
         fim_for_hess: Optional[bool] = True,
         amici_object_builder: Optional[AmiciObjectBuilder] = None,
-        calculator: Optional[AmiciCalculator] = None,
+        calculator: Optional[
+            Union[AmiciCalculator, "InnerCalculatorCollector"]  # noqa: F821
```
dweindl, line 87: Import InnerCalculatorCollector `if typing.TYPE_CHECKING`

dweindl (approved):
Makes sense.

Doresic (approved):
Good change. Thanks!

PR #1429, opened 2024-07-15
Disentangle amici and roadrunner

PaulJonasJost:
When installing only pypesto[roadrunner], there are additional amici dependencies, mainly occurring in hierarchical (basically all of them typings, thus fixed through string type annotation)


dweindl (commented):
Thanks. For all touched Python modules, I'd prefer adding `from __future__ import annotations` instead of the changes made here. This is will cover any future changes in type annotations there.

setup.cfg
```diff
@@ -97,6 +97,8 @@ amici =
     amici >= 0.21.0
 roadrunner =
     libroadrunner >= 2.7.0
+    petab >= 0.4.0
```
dweindl, line 100:
```suggestion
    %(petab)s
```

dweindl (approved):
:+1:

pypesto/hierarchical/semiquantitative/solver.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
FFroehlich, line 1: unused?

FFroehlich (approved):

pypesto/hierarchical/semiquantitative/solver.py
PaulJonasJost: needed for the ParameterMapping

m-philipps (approved):
Looks good as far as I can tell.

Issue #1117, opened 2023-09-12
Calculation of credibility intervalls currently not via joint posterior


shoepfl:
Hi there,

i recently figured out that the credibility intervals in pyPESTO are not calculated based on the x-percent HDI of the joint posterior but via percentiles for each parameter separately.
This is for example in the function `calculate_ci_mcmc_sample` the case.

Wouldn´t it be better to calculate a posterior threshold based on an x-percent HDI on the joint posterior and afterward restrict the individual parameter CIs to the overall posterior threshold value?
I think the other method could be problematic as individual calculation on each parameter does not necessarly have to match with the calculation on the joint posterior.

Best,

Sebastian

JanHasenauer:
Hey Sebastian,

yes, using the HDI computed from the full posterior would definitely be preferable. 
We can add this to the todo list.

If you happen to have a good code lying around, we would be happy if you could make a suggestion.

Best,
Jan

shoepfl:
Hi Jan,

my idea was to take the Highest Posterior Density (HPD), which could be calculated by sorting the posterior values and throwing away the lowest x-percent of these values.

A short example would be:
`import h5py`
`import numpy as np`

`file = h5py.File(example_pypesto_mcmc_chain.hdf5'), 'r')`
`neglogpost = np.sort(self.file['sampling']['results']['trace_neglogpost'][:][0])`
`n_samples = self.file['sampling']['results']['trace_neglogpost'].shape[1]  # total number of posterior samples`
`# lower 1-ci percentage samples of the posterior are removed to get the according HPD`
`self.lower_neglogpost = self.neglogpost[int(n_samples*(1-ci)):]`

of course, this comes with the effect that the interval might be split in multimodal cases.

shoepfl:
Adding to this, here would be a function that calculates HPD parameters based on the Pypesto sampling result:

`def calculate_hpd(pypesto_result, ci_level: float, stepsize: int = 1) -> pd.DataFrame:`
`    """Calculate Highest Posterior Density (HPD) samples of pypesto sampling result.`

`    Input:`
`        pypesto_result: The pyPESTO result object with filled sample result.`
`        ci_level: credibility level of HPD`
`        stepsize: Only one in `stepsize` values is plotted.`
`    """`
`    # get names of chain parameters`
`    param_names = pypesto_result.problem.get_reduced_vector(pypesto_result.problem.x_names)`

`    # first calculate burn in`
[...]

PR #1431, opened 2024-07-17
Added HPD calculation to ensemble

shoepfl:
The ensemble from_sample method now has an additional rel_cutoff argument. rel_cutoff allows the posterior to be cut to the alpha % highest density (here, non-normalized posterior probability).

I checked the code with my workflow, and it worked. Anyway, I would be happy to receive suggestions for improvement.

Adresses #1117

### DIFF
--- a/pypesto/ensemble/ensemble.py
+++ b/pypesto/ensemble/ensemble.py
@@ -555,6 +555,7 @@ class Ensemble:
     def from_sample(
         result: Result,
         remove_burn_in: bool = True,
+        rel_cutoff: float = None,
         chain_slice: slice = None,
         x_names: Sequence[str] = None,
         lower_bound: np.ndarray = None,
@@ -571,6 +572,11 @@ class Ensemble:
         remove_burn_in:
             Exclude parameter vectors from the ensemble if they are in the
             "burn-in".
+        rel_cutoff:
+            Relative cutoff. Exclude parameter vectors, for which the
+            (non-normalized) posterior value difference to the best vector is greater than
+            cutoff, i.e. include all vectors such that
+            `neglogpost(vector) <= neglogpost(max * rel-cutoff)`.
         chain_slice:
             Subset the chain with a slice. Any "burn-in" removal occurs first.
         x_names:
@@ -599,9 +605,19 @@ class Ensemble:
                 geweke_test(result)
             burn_in = result.sample_result.burn_in
             x_vectors = x_vectors[burn_in:]
+        else:
+            burn_in = 0
+
+        # added cutoff
+        if rel_cutoff is None:
+            pass
+        else:
+            x_vectors = calculate_hpd(result=result, burn_in=burn_in, ci_level=rel_cutoff)
+
         if chain_slice is not None:
             x_vectors = x_vectors[chain_slice]
         x_vectors = x_vectors.T
+
         return Ensemble(
             x_vectors=x_vectors,
             x_names=x_names,
@@ -1253,3 +1269,66 @@ def calculate_cutoff(
 
     range = chi2.ppf(q=percentile / 100, df=df)
     return fval_opt + range
+
+
+def calculate_hpd(
+    result: Result,
+    burn_in: int,
+    ci_level: float = 95,
+):
+    """
+    Calculate Highest Posterior Density (HPD) samples of pypesto sampling result.
+
+    The HPD is calculated for a user-defined credibility level (alpha). The HPD includes all
+    parameter vectors with a (non-normalized) posterior probability that is higher than the lowest 1-alpha %
+    posterior probability values.
+
+    Parameters
+    ----------
+    result:
+        The optimization result from which to create the ensemble.
+    burn_in:
+        Burn in index that is cut off before HPD is calculated.
+    ci_level:
+        Credibility level of the resulting HPD. 0.95 corresponds to the 95% CI.
+
+    Returns
+    -------
+    The HPD parameter vector.
+    """
+    # get names of chain parameters
+    param_names = result.problem.get_reduced_vector(result.problem.x_names)
+
+    # Get converged parameter samples as numpy arrays
+    chain = np.asarray(result.sample_result.trace_x[0, burn_in:, :])
+    neglogpost = result.sample_result.trace_neglogpost[0, burn_in:]
+    indices = np.arange(burn_in, len(result.sample_result.trace_neglogpost[0, :]))
+
+    # create df first, as we need to match neglogpost to the according parameter values
+    pd_params = pd.DataFrame(chain, columns=param_names)
+    pd_fval = pd.DataFrame(neglogpost, columns=['neglogPosterior'])
+    pd_iter = pd.DataFrame(indices, columns=['iteration'])
+
+    params_df = pd.concat(
+        [pd_params, pd_fval, pd_iter], axis=1, ignore_index=False
+    )
+
+    # get lower neglogpost bound for HPD
+    # sort neglogpost values of MCMC chain without burn in
+    neglogpost_sort = np.sort(neglogpost)
+
+    # Get converged chain length
+    chain_length = len(neglogpost)
+
+    # most negative ci percentage samples of the posterior are kept to get the according HPD
+    neglogpost_lower_bound = neglogpost_sort[int(chain_length*(ci_level))]
+
+    # cut posterior to hpd
+    hpd_params_df = params_df[params_df['neglogPosterior'] <= neglogpost_lower_bound]
+
+    # convert df to ensemble vector
+    hpd_params_df_vals_only = hpd_params_df.drop(columns=['iteration', 'neglogPosterior'])
+    hpd_ensemble_vector = hpd_params_df_vals_only.to_numpy()
+    print(hpd_ensemble_vector)
+
+    return hpd_ensemble_vector


dilpath:
@shoepfl Thanks very much!

@PaulJonasJost any thoughts on how to consolidate the two different cutoff methods? With this PR, using `rel_cutoff` with `Ensemble.from_optimization_*` will give chi2-based cutoffs, but `Ensemble.from_sample` will give hpd-based cutoffs. Ideally, the user would be able to choose. Should we create some `Cutoff` class that can be used like so?
```python
class Cutoff:
    ...
    def __call__(self, xs, fs): # or rather pass a list of vectors and their fvals?
        return self.apply_cutoff(xs=xs, fs=fs)
    @abstractmethod
    def apply_cutoff(self, xs=xs, fs=fs):
        ...  # specific rel/chi2/hpd cutoff method here
class RelCutoff(Cutoff): ...
class Chi2Cutoff(Cutoff): ...
class HpdCutoff(Cutoff): ...

cutoff = Chi2Cutoff(percentile=...)

ensemble = Ensemble.from_sample(result=result, cutoff=cutoff)
```
and inside `Ensemble.from_sample`:
```python
def from_sample(..., cutoff: Cutoff):
    ...
    x_vectors = cutoff(xs=result.sample_result.trace_x[0, burn_in:, :], fs=result.sample_result.trace_fval[0, burn_in:])
    ...
```

We could also integrate things like `remove_burn_in` and `chain_slice` (see `Ensemble.from_sample`) into `Cutoff`.

pypesto/ensemble/ensemble.py
```diff
@@ -571,6 +572,11 @@ def from_sample(
         remove_burn_in:
             Exclude parameter vectors from the ensemble if they are in the
             "burn-in".
+        rel_cutoff:
+            Relative cutoff. Exclude parameter vectors, for which the
+            (non-normalized) posterior value difference to the best vector is greater than
+            cutoff, i.e. include all vectors such that
+            `neglogpost(vector) <= neglogpost(max * rel-cutoff)`.
```
PaulJonasJost, line 579: If i understand the calculate_hpd correct, it is not the ones that are greater than a certain differnece, but the the `real_cutoff`-quantile (line 1324), in this case one should perhaps rename it more appropriately to `x_quantile` and also check (again line 1324) that it is <= 1?

pypesto/ensemble/ensemble.py
```diff
@@ -599,9 +605,19 @@ def from_sample(
...
+            burn_in = 0
+
+        # added cutoff
+        if rel_cutoff is None:
+            pass
+        else:
+            x_vectors = calculate_hpd(result=result, burn_in=burn_in, ci_level=rel_cutoff)
+
         if chain_slice is not None:
             x_vectors = x_vectors[chain_slice]
```
PaulJonasJost, line 618: A general remark, but it is not quite clear whether this is the desired order, it might be that we want to first slice and then calculate hpd. But that is only a minor thing here i guess.

pypesto/ensemble/ensemble.py
```diff
@@ -1253,3 +1269,65 @@ def calculate_cutoff(
 
     range = chi2.ppf(q=percentile / 100, df=df)
     return fval_opt + range
+
+
+def calculate_hpd(
+    result: Result,
+    burn_in: int,
```
PaulJonasJost, line 1276:
```suggestion
    burn_in: int = 0,
```
Just for the sake of having a default value here not requiring an input

pypesto/ensemble/ensemble.py
```diff
@@ -1253,3 +1269,65 @@ def calculate_cutoff(
...
+    ci_level: float = 95,
```
PaulJonasJost, line 1277:
```suggestion
    ci_level: float = .95,
```
I guess? otherwise 1324 would most certainly throw an error i think (out of bounds)

PaulJonasJost (changes requested):
Thanks for the addition. I think a test should be added as well, if you want we can help with that.

PaulJonasJost:
> @shoepfl Thanks very much!
> 
> @PaulJonasJost any thoughts on how to consolidate the two different cutoff methods? With this PR, using `rel_cutoff` with `Ensemble.from_optimization_*` will give chi2-based cutoffs, but `Ensemble.from_sample` will give hpd-based cutoffs. Ideally, the user would be able to choose. Should we create some `Cutoff` class that can be used like so?
> 
> ```python
> class Cutoff:
>     ...
>     def __call__(self, xs, fs): # or rather pass a list of vectors and their fvals?
>         return self.apply_cutoff(xs=xs, fs=fs)
>     @abstractmethod
>     def apply_cutoff(self, xs=xs, fs=fs):
>         ...  # specific rel/chi2/hpd cutoff method here
> class RelCutoff(Cutoff): ...
> class Chi2Cutoff(Cutoff): ...
> class HpdCutoff(Cutoff): ...
> 
> cutoff = Chi2Cutoff(percentile=...)
> 
> ensemble = Ensemble.from_sample(result=result, cutoff=cutoff)
> ```
> 
> and inside `Ensemble.from_sample`:
> 
> ```python
> def from_sample(..., cutoff: Cutoff):
>     ...
>     x_vectors = cutoff(xs=result.sample_result.trace_x[0, burn_in:, :], fs=result.sample_result.trace_fval[0, burn_in:])
>     ...
> ```
> 
> We could also integrate things like `remove_burn_in` and `chain_slice` (see `Ensemble.from_sample`) into `Cutoff`.

I think as it is currently used `rel_cutoff` here might be renamed. If i understand your suggestion correctly, the main advantage would be a reduction in direct parameters when calling `Ensemble.from_...`with a potential easier overview? Removing burn in and chain slice are quite sample specific, i am not entirely sure whether from optimization could straightforward be used as equivalents. But the idea itself sounds good.

pypesto/ensemble/ensemble.py
shoepfl: In principle, I would rather calculate the HPD first and then take a slice, as the HPD should become more precise that way; computationally, it's not an issue. But if there are other reasons to change the order, that would also be fine with me.

pypesto/ensemble/ensemble.py
shoepfl: So basically what I meant with "difference to best vector" is that we do not have symmetric percentiles but rather cut the parameters with the lowest posterior probability. Maybe the following description makes it clearer:

"Exclude parameter vectors, which have a (non-normalized) posterior value below the defined credibility level."

If we agree to use the naming "quantile" explicitly we have to make sure that there is no confusion on the quantile calculation on parameter values vs. the calculation on the posterior probability of the joint samples in this case.

shoepfl:
@PaulJonasJost, a test sounds good. Indeed, it would be nice if you could assist here, as there may already be similar test cases in the toolbox that we could add to.

PaulJonasJost:
> @PaulJonasJost, a test sounds good. Indeed, it would be nice if you could assist here, as there may already be similar test cases in the toolbox that we could add to.

I will look into it and write a test if you are fine with me editing the branch 👍🏼

shoepfl:
yes of course, and if I can help you can just assign me some tasks

pypesto/ensemble/ensemble.py
```diff
@@ -1253,3 +1269,65 @@ def calculate_cutoff(
...
+
+def calculate_hpd(
+    result: Result,
+    burn_in: int = 0,
+    ci_level: float = .95,
+):
+    """
+    Calculate Highest Posterior Density (HPD) samples of pypesto sampling result.
+
+    The HPD is calculated for a user-defined credibility level (alpha). The HPD includes all
```
dweindl, line 1282: Rephrase to match argument names? "alpha" is potentially confusing since the function doesn't take an alpha parameter.

PaulJonasJost:
added the test and renamed. Now i guess someone else has to review (@dilpath)?

pypesto/ensemble/ensemble.py
```diff
@@ -599,9 +604,21 @@ def from_sample(
                 geweke_test(result)
             burn_in = result.sample_result.burn_in
             x_vectors = x_vectors[burn_in:]
+        else:
+            burn_in = 0
```
dilpath, line 608: Usually just write `burn_in = 0` above the `if` statement.

pypesto/ensemble/ensemble.py
```diff
@@ -599,9 +604,21 @@ def from_sample(
...
+
+        # added cutoff
+        if f_quantile is None:
+            pass
+        else:
```
dilpath, line 613:
```suggestion
        # added cutoff
        if f_quantile is not None:
```

pypesto/ensemble/ensemble.py
```diff
@@ -599,9 +604,21 @@ def from_sample(
...
+            x_vectors = calculate_hpd(
+                result=result, burn_in=burn_in, ci_level=f_quantile
```
dilpath, line 615: Why `f_quantile` here but `ci_level` in `calculate_hpd`? Same terms everywhere would be preferable

pypesto/ensemble/ensemble.py
```diff
@@ -571,6 +572,10 @@ def from_sample(
         remove_burn_in:
             Exclude parameter vectors from the ensemble if they are in the
             "burn-in".
+        f_quantile:
+            A form of relative cutoff. Exclude parameter vectors, for which the
+            (non-normalized) posterior value is not within the f_quantile best
```
dilpath, line 577:
```suggestion
            (non-normalized) posterior value is not within the `f_quantile` best
```

pypesto/ensemble/ensemble.py
```diff
@@ -1253,3 +1270,77 @@ def calculate_cutoff(
...
+
+def calculate_hpd(
+    result: Result,
+    burn_in: int = 0,
+    ci_level: float = 0.95,
+):
+    """
+    Calculate Highest Posterior Density (HPD) samples of pypesto sampling result.
+
+    The HPD is calculated for a user-defined credibility level (ci_level). The
```
dilpath, line 1283:
```suggestion
    The HPD is calculated for a user-defined credibility level (`ci_level`). The
```

pypesto/ensemble/ensemble.py
```diff
@@ -1253,3 +1270,77 @@ def calculate_cutoff(
...
     return fval_opt + range
+
+
+def calculate_hpd(
+    result: Result,
+    burn_in: int = 0,
+    ci_level: float = 0.95,
+):
+    """
+    Calculate Highest Posterior Density (HPD) samples of pypesto sampling result.
```
dilpath, line 1281:
```suggestion
    Get Highest Posterior Density (HPD) samples.
```

pypesto/ensemble/ensemble.py
```diff
@@ -1253,3 +1270,77 @@ def calculate_cutoff(
...
+
+    The HPD is calculated for a user-defined credibility level (ci_level). The
+    HPD includes all parameter vectors with a (non-normalized) posterior
+    probability that is higher than the lowest 1-ci_level %
+    posterior probability values.
+
+    Parameters
+    ----------
+    result:
+        The optimization result from which to create the ensemble.
```
dilpath, line 1291:
```suggestion
        The sampling result from which to create the ensemble.
```

pypesto/ensemble/ensemble.py
```diff
@@ -1253,3 +1270,77 @@ def calculate_cutoff(
...
+    result: Result,
+    burn_in: int = 0,
+    ci_level: float = 0.95,
+):
+    """
+    Calculate Highest Posterior Density (HPD) samples of pypesto sampling result.
+
+    The HPD is calculated for a user-defined credibility level (ci_level). The
+    HPD includes all parameter vectors with a (non-normalized) posterior
+    probability that is higher than the lowest 1-ci_level %
```
dilpath, line 1285:
```suggestion
    probability that is higher than the lowest `1-ci_level` %
```

pypesto/ensemble/ensemble.py
```diff
@@ -1253,3 +1270,77 @@ def calculate_cutoff(
...
+
+    Parameters
+    ----------
+    result:
+        The optimization result from which to create the ensemble.
+    burn_in:
+        Burn in index that is cut off before HPD is calculated.
+    ci_level:
+        Credibility level of the resulting HPD. 0.95 corresponds to the 95% CI.
+        Values between 0 and 1 are allowed.
```
dilpath, line 1296:
```suggestion
        Only values between 0 and 1 are allowed.
```

pypesto/ensemble/ensemble.py
```diff
@@ -1253,3 +1270,77 @@ def calculate_cutoff(
...
+
+    Returns
+    -------
+    The HPD parameter vector.
```
dilpath, line 1300:
```suggestion
    The HPD parameter vectors.
```

pypesto/ensemble/ensemble.py
```diff
@@ -1253,3 +1270,77 @@ def calculate_cutoff(
...
+    """
+    if ci_level < 0 or ci_level > 1:
```
dilpath, line 1302: Matches the error message now.
```suggestion
    if not 0 <= ci_level <= 1:
```

dilpath (approved):
Thanks, looks good :+1:

Independently of a re-write into classes like `HpdCutoff` as discussed above, I think it would already be valuable to change `calculate_hpd` to instead take parameter vectors, and their objective function values (i.e., objects like `result.sample_result.trace_x[0]` and `result.sample_result.trace_fval` instead of `result`). Currently, `calculate_hpd` handles burn-in and is only applicable to `result.sample_result` objects, which seems suboptimal, since the logic should apply to any collection of vectors and their functional values.

Anyway, fine to merge as-is, since these `Ensemble.from_...` methods need some work anyway.

PaulJonasJost (approved):

PR #1435, opened 2024-07-30
Visualize: allow for log scale of hierarchical parameters for visualization

Doresic:
Allow for 'log' and 'log10' parameter scales of scaling and offset hierarchical parameters. This does not change anything in their inner optimizations, as that's done analytically, but it allows for better visualization of their values in parameter plots together with all other (mostly log-scaled) parameters. I added a Boehm example below.

Currently not supported for hierarchical sigma parameters as they cannot have different parameter bounds than (0, +inf). So log-scale of its lower bound is -inf causing issues in petab.lint checks. It would be possible to fix, but would require some larger restructuring of when the petab.lint is applied, removing hierarchical sigma parameters from the parameter table before the lint or changes in petab.


Doresic:
Boehm_log_scaling_offset

Doresic:
Added the scale description into the labels of parameters into the parameter plot 
Boehm_log_scaling_text

pypesto/hierarchical/base_parameter.py
```diff
@@ -68,6 +68,23 @@ def __init__(
...
+        ):
+            raise ValueError(
+                f"Inner parameter type `{inner_parameter_type}` "
+                f"cannot be log-scaled."
+            )
+
+        if scale in [LOG, LOG10] and lb <= 0:
+            raise ValueError(
+                f"Lower bound of inner parameter `{inner_parameter_id}` "
+                f"cannot be non-positive for log-scaled parameters."
```
dweindl, line 84:
```suggestion
                f"cannot be non-positive for log-scaled parameters. "
```

pypesto/hierarchical/base_problem.py
```diff
@@ -231,6 +237,28 @@ def scale_value(val: float | np.array, scale: str) -> float | np.array:
...
+    return scaled_dct
+
+
+def scale_back_value(val: float | np.array, scale: str) -> float | np.array:
+    """Scale back a single value."""
+    if scale == "lin":
+        return val
+    if scale == "log":
+        return np.exp(val)
+    if scale == "log10":
```
dweindl, line 257: Use `LIN, LOG, LOG10` as above

pypesto/problem/hierarchical.py
```diff
@@ -35,6 +35,11 @@ class HierarchicalProblem(Problem):
         Only relevant if hierarchical is True. Contains the bounds of easily
         interpretable inner parameters only, e.g. noise parameters, scaling
         factors, offsets.
+    inner_scales:
+        The scales for the inner optimization parameters. Only relevant if
+        hierarchical is True. Contains the scales of easily interpretable inner
+        parameters only, e.g. noise parameters, scaling factors, offsets. Can
+        be LIN, LOG or LOG10. Used only for visualization purposes.
```
dweindl, line 42: pypesto.C.{LIN,LOG,LOG10}

pypesto/visualize/parameters.py
```diff
@@ -404,8 +415,13 @@ def handle_inputs(
     if ub is None:
         ub = result.problem.ub_full
 
-    # get labels
-    x_labels = result.problem.x_names
+    # get labels as x_names and scales
+    x_labels = [
+        f"{x_name} ({x_scale})"
```
dweindl, line 420: If all parameters use the same scale, the scale could be included once in the x axis label instead of repeatedly in the tick labels.

dweindl (approved):
Thanks, looks good to me.

pypesto/hierarchical/semiquantitative/problem.py
```diff
@@ -420,7 +433,7 @@ def noise_inner_parameters_from_parameter_df(
             SplineInnerParameter(
                 inner_parameter_id=row[PARAMETER_ID],
                 inner_parameter_type=InnerParameterType.SIGMA,
-                scale=LIN,
+                scale=row[PARAMETER_SCALE],
```
PaulJonasJost, line 436: If sigma parameters cannot handle the log values, why would we change that here from the LIN?

PaulJonasJost (approved):
Looks good to me.

pypesto/hierarchical/semiquantitative/problem.py
Doresic: True... This is left from a version in which I tried doing it for sigma as well.
But even tho it doesn't work for sigma, it might be good to keep row[PARAMETER_SCALE] here. There are petab checks that will fail before you come to this spot if you have any LOG scale for the sigma parameters. So it feels cleaner to have it taken from petab parameter df instead of hard-coding LIN scale here.

pypesto/visualize/parameters.py
Doresic: Hmmm true, will change

Doresic: I changed it, but it is quite a bit of changes to make it work, worth it?
Look at `x_axis_label` choice after `# If all the scales are the same, put it in the x_axis_label` comment at line 465, and then it has to be passed back and into `parameters_lowlevel`. Works tho, and the plot is nicer.

Doresic:
@PaulJonasJost I did add the option we talked about in the meeting briefly, that sigma hierarchical parameters are plotted in LOG10 scale by default. But it takes some manual finding which parameters are sigma parameters in the inner subproblems which can contain noise parameters, then rescaling all the estimated values and its bounds to LOG10. Not the nicest, but it works.

For reference, lines 536 to 558 in `_handle_inner_inputs` and the `log10_scale_hier_sigma` argument being passed around everywhere until it comes to that point.

Can leave and can remove, what do you think?

pypesto/visualize/parameters.py
dweindl: Works for me, but it was an optional suggestion for now.



> and then it has to be passed back and into `parameters_lowlevel`.

Alternatively, you could override labels outside parameters_lowlevel, but this doesn't simplify things that much, right?

Doresic: Yeah not much, think it would be even uglier if it was outside of there

Doresic: Will see what the visualization ppl (@m-philipps and @stephanmg) think.

stephanmg: Hi, looks fine to me, don't think the repetition of scale label is too bad. Certainly optional to think and implement a more visually pleasing appearance.

stephanmg (approved):

PR #1442, opened 2024-08-03
Petab Importer reforge

### LINKED ISSUE #1301: Decouple PEtab and AMICI
Currently, the PEtab importer depends on AMICI, so I have to install AMICI to use it. However, I use PEtab+pyPESTO in a few applications that do not involve AMICI. Using the PEtab importer to create the pyPESTO problem is convenient, and I supply a custom (non-AMICI) objective to `PetabImporter.create_problem(objective=custom_objective)`. Having it default to using an AMICI objective is fine, but then the imports etc need to handled properly when AMICI is not available.

PaulJonasJost:
CHanged the way the petab importer works. It used to be tailored to amici models and contain a lot of functionality that were only necessary for objective function creation of amici. As we now support other simulators, changed it to a `Importer-Factory` design.

The importer only creates the problem now. You can however supply a `simulator_type` to it, which then creates the corresponding factory, creating the objective. All Functions that are simulator specific are in the factory.

This should make the Code more clear in my opinion, we are getting rid of the PetabImporter depending on the AmiciObjectBuilder and make it much more versatile to support a new simulator.

Also added a basic support for any PEtab simulator. This allows every simulator who is supporting PEtab to directly use pyPESTO. This is not an efficient implementation, but it is also not meant to be one, rather lowering hurdles to use pyPESTO.

- [x] Check that all tests work fine and adapt to new code structure


dweindl (commented):
Thanks. Some first feedback:

Overall, it makes sense to split up the PEtab-related functionality. 

I am not sure how much demand there really is for the petab.Simulator support, but fine to have it. (It should be relatively easy to include a new simulator, but for really using it during parameter estimation, one would probably want a more efficient implementation).

I am not so happy with the naming yet, see specific comments.

pypesto/petab/importer.py
```diff
@@ -809,39 +408,74 @@ def create_problem(
...
-        if model is None:
-            model = self.create_model(verbose=verbose)
-
-        measurement_df = self.petab_problem.measurement_df
-
-        return amici.petab.simulations.rdatas_to_measurement_df(
-            rdatas, model, measurement_df
+        """See :meth:`AmiciFactory.rdatas_to_measurement_df`."""
+        raise NotImplementedError(
+            "This function has been moved to `AmiciFactory`."
```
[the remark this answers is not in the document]
PaulJonasJost: i'll be honest, i dont think they even belong to pyPESTO, as it is a really light wrapper around it. Can create a deprecation warning as well.

Why would it require a 1.0? we could go with a 0.6.0 as we have done before on breaking changes?

pypesto/petab/factory.py
```diff
@@ -0,0 +1,854 @@
...
+                    elif isinstance(parameter_id_or_value, numbers.Number):
+                        mapping_per_condition[3][override] = LIN
+                    else:
+                        raise ValueError(
+                            "The parameter value in the condition table "
+                            "is not a number or a parameter ID."
+                        )
+        return mapping
+
+    def create_objective(
```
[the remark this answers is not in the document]
PaulJonasJost: Could if we view the sole purpose of the factory as the objective generation. If we view the other objects like datas and/or model part of its tasks as well, they should not be private. The petabImporter had them previously also as not private. If we want them to be private, there would be no need to have a lot of arguments in them allowing to separately call them as they anways were not meant to be used that way?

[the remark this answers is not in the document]
dweindl: I will have to have a closer look what might be useful elsewhere. Not everything that was public before should have been public before. If the other functions are to be used separately, then they should not be in this class, or this class should not be called ObjectiveFactory or similar. A XFactory has (or should have) a single purpose, which is to produce an instance of X.

pypesto/objective/petab.py
```diff
@@ -0,0 +1,108 @@
...
+        x_dict = OrderedDict(zip(self._x_names, x))
+        x_unscaled = self.petab_problem.unscale_parameters(x_dict)
+        par_df = self.petab_problem.parameter_df
+        par_df["nominalValue"] = par_df.index.map(x_unscaled)
+        # add here try except whether simulator has its own method for this
+        try:
+            self.simulator.set_parameters(x_unscaled)
+        except AttributeError:
+            self.simulator = self.simulator.__class__(
+                petab_problem=self.petab_problem
```
fbergmann, line 64: @PaulJonasJost  The issue with the reinitialization here, is that any parameters set before using custom constructors will be lost. I'm not sure how bad that is, but at the very lest, the working directory should be passed along as well. 

just something to think about

PaulJonasJost: I think it is not feasible to accurately reinitialise, thus i removed the tryExcept and we expect the Simulator to have such a functionality.

pypesto/petab/importer.py
```diff
@@ -171,11 +166,21 @@ def __init__(
             model_name = _find_model_name(self.output_folder)
         self.model_name = model_name
 
+        self.simulator_type = simulator_type
+        self.simulator = simulator
```
dilpath, line 170: As the importer only needs to use this in `create_objective`, why not remove these attributes and implement them as arguments of `create_objective` instead?

e.g. if I want to compare the libroadrunner and AMICI objectives, why do I need two importer objects? I could use the same importer like `importer.create_objective_creator(..., simulator_id=AMICI)` and `importer.create_objective_creator(..., simulator_id=ROADRUNNER)`

No strong opinion on this though, fine as is

dilpath (approved):
Overall, looks nice, thanks :slightly_smiling_face:

pypesto/petab/importer.py
Doresic: +1 
Think this makes sense. Also not strong opinion.

pypesto/petab/importer.py
```diff
@@ -739,7 +342,8 @@ def create_problem(
         A :class:`pypesto.problem.Problem` for the objective.
         """
         if objective is None:
-            objective = self.create_objective(**kwargs)
+            self.factory = self.create_objective_creator()
```
Doresic, line 345: Why are the names of the objects different?
Importer has them as `self.factory`, and they are in the `factory.py` file, but all of the classes are called `Creator`.
Wouldn't it be simpler if it's either only one or the other?

i.e. `self.factory = self.create_objective_factory()` and those are, for instance, `PetabSimulatorObjectiveFactory`

Doresic: Oh, I just saw the discussion you had with Daniel about the naming (about XFactory should only make X). 
Should in that case nothing be called factory, and all should be creators, including here?

Doresic (approved):
Looks great. Good changes.

pypesto/petab/importer.py
PaulJonasJost: having this would make it imo much easier to 
1. init the importer -> 2. importer.create_problem()
without having to pass these arguments constantly to create_problem. Could add it as additional possibility to `create_objective` such that people can do either? Argument in `create_objective` would then take precedence

dilpath: IMO the cleanest implementation would be that default behavior (AMICI simulator) is usable as you said like
```python
importer = PetabImporter(petab_problem)
problem = importer.create_problem()
```

but custom behavior is like
```python
importer = PetabImporter(petab_problem)
obj = importer.create_objective(AMICI)
obj.amici_solver.setAbsoluteTolerance(1e-8)
problem = importer.create_problem(obj)
```
Since it's quite common for me to customize things like simulator tolerances, passing `obj` to `create_problem` is no change for me.

Just to explain my (weak) preference.

PaulJonasJost: passing the objective to problem is still possible. And the other behavior would look like (+- details)
```python
importer = PetabImporter(petab_problem, AMICI)
obj = importer.create_objective()
obj.amici_solver.setAbsoluteTolerance(1e-8)
problem = importer.create_problem(obj)
```
I think going with allowing both would be best solution to be able to adjust to individual preferences?

dweindl: By now, `PetabImporter` doesn't have much functionality or internal state left. Maybe we could directly go for something like

```python
importer = AmiciPetabImporter(petab_problem, ...)
problem = importer.create_problem()
...
```

PaulJonasJost: > By now, `PetabImporter` doesn't have much functionality or internal state left. Maybe we could directly go for something like
> 
> ```python
> importer = AmiciPetabImporter(petab_problem, ...)
> problem = importer.create_problem()
> ...
> ```

I think currently the main advantage is that we can easily switch by just changing an argument in the Importer to switch simulators if needed/wanted which we would lose with the above. additionally the `create_problem` function is for all of them kind of the same which just in my mind creates this natural hierarchy of "Importer -> Inner thing (=Objective creator)"

FFroehlich (approved):

PaulJonasJost:
As no one had a strong opinion, I would put it up to vote how to organize the code:

1. Keep the *PetabImporter* and the *objective creator* separate but don't enforce setting the simulator_type in the init of the importer, rather also allowing it to be put in the `create_objective_creator` as well. (Vote with 🎉)
    **Mentioned Up-/Downsides**:
        - (+) Clear separation of tasks, one creates the objective and anything auxiliary, the other only has to create the problem from the petab problem
        - (-) More overloading of arguments in the PetabImporter
        - (+) Easy switching of underlying objective and/or potential to easily create two different problems to compare
2. For each simulator type create its own ***Simulator**PetabImporter* (Vote with 🚀 )
    **Mentioned Up-/Downsides**:
        - (+) Loose some arguments and no overload of arguments
        - (+/-) Clear functionality tailored to **exactly** one simulator type
        - (-) **Strong** break with current codepipelines. Any current code having previously used the importer will have to be revised

I would keep the poll open till end of tomorrow :)

PS: sorry for the emojis of vote, these were the two most neutral ones I saw one can react with under a comment.

Issue #1445, opened 2024-08-09
objective.FD does not inherit the PrePostProcessor


PaulJonasJost:
If we dont inherit the PrePostProcessor from the objective, potentially fixed parameters will have the system fail

PR #1446, opened 2024-08-09
FD objective with fixed parameters fix

PaulJonasJost:
Fixed an issue when the objective used for the FD had a PrePostProcessor, i.e. fixed values. Also added it to the fd test


PaulJonasJost:
Closes #1445

PR #1447, opened 2024-08-14
Profile: improve profiling code

Doresic:
I implemented improvements and changes to the profiling algorithm and visualization. The following was tested on 5 models (Boehm, Rosenbrock, Randall, Yuhong, Wuyan) and has seen improvements in robustness and stability of profiles.

Major changes:
- Changed the default profiling `next_guess_method` from `adaptive_step_regression` to `adaptive_step_order_1`. The former had a problem with poor extrapolation -- polynomials of high degree are known to diverge outside of the fitting region. In the testing I've done, the latter was more robust and faster.
- The adaptive method chooses a heuristic `next_obj_target` and changes the step length until the proposed objective function value is stepped over. Until now, the adaptive method considered only a `high_next_obj_target` value that is higher than the current objective function value. This caused it to make large steps if the profiler was in a region of decreasing objective function values. Thus, I added the `low_next_obj_target` as well. If any of the two is reached, the corresponding step is returned.
- In many cases, the local optimization in `walk_along_profile` failed. In most cases, this was due to a poor initial proposal of `next_x` from which the optimization was started. In cases where this happened, the only fail-safe was to try to resample the `next_x` uniformly. This made the profile very discontinuous as it jumped randomly through the parameter space. Thus, in case of optimization failure before resorting to resampling of `next_x` I made the algorithm re-try the `next_guess` generation, but this time with different `max_step_size` and `min_step_size` hyperparameters. First, the `next_guess` function is called with an increasingly smaller `max_step_size` to try to stay closer to the last point which had a successful optimization. This is done until optimization is successful or `max_step_size` is reduced below `min_step_size`. Then, the idea is to try to jump over the "region of poor simulation/optimization" by iteratively calling `next_guess` with an increasingly bigger `min_step_size`.  If these two do not succeed the `next_x` is randomly re-sampled as a last resort.
- I changed some things about the `profiles()` visualization:
  - The visualization now plots dots to show the density of the profile. Simple, but nice to have.
  - Added a `plot_objective_values` option. This makes the plotting routine plot the objective function value across the parameter values instead of the default likelihood ratio. It was interesting to inspect here and there, so I left it there as I think it might be nice for others.
  - Added y-ticks to all subplots, as the scales of different subplots can be drastically different.
  - Added a `quality_colors` option. As the profile now has adaptive `max_step_size` and `min_step_size` values, and sometimes even re-samples the points, it's nice to know when what happens. Thus, red points indicate a step with a reduced `max_step_size`, blue points indicate a step with an increased `min_step_size` and green points indicate that the next guess `x_next` was resampled. If none of these happened, the color is black. These colors are saved throughout profiling in `ProfilerResult.x_color_path`.

Minor changes:
- Added constant logging.INFO information about the progress of the profiling. There was nothing of this sort there so it was impossible to know if the profiling entered some buggy infinite loop.
- Changed the form of the `next_obj_target` to depend on the objective function distance to the last profile point (`last_delta_fval`) rather than the distance to the global optimum objective function value (`delta_obj_value`). The latter made it make larger steps in low-likelihood regions. However, I'm not sure whether it is desirable to that degree. 
- Added a trust region of extrapolation in all parameters other than the one that's being profiled. This restricts the extrapolations from making too large steps.
- Had to increase step checks in the profile tests -- The adapted algorithm makes more steps in regions of decreasing objective function values. 

Fixed some bugs:
- If the profiling was started at a lower parameter boundary, the order one extrapolation would explode on the first step to the right as the x difference was 0 -> division by 0.
- There was a bug with stopping criteria when reaching the left boundary. The step returned would be on the other side, and the profiling would bounce off the edge, and continue in the other direction when it shouldn't -- ending up in an infinite loop without exit condition. Absolute values at the right spots fixed it.


pypesto/profile/options.py
```diff
@@ -112,5 +112,5 @@ def validate(self):
         if self.default_step_size < self.min_step_size:
             raise ValueError("default_step_size must be >= min_step_size.")
 
-        if self.magic_factor_obj_value < 0 or self.magic_factor_obj_value >= 1:
-            raise ValueError("magic_factor_obj_value must be >= 0 and < 1.")
+        if self.magic_factor_obj_value < 1:
+            raise ValueError("magic_factor_obj_value must be > 1.")
```
Doresic, line 116: The magic factor now has to be larger than 1 rather than between 0 and 1. The next_obj_target now depends on the objective function value difference of the last two profile points, rather than the difference of the last profile point to the global minimum. The former is in general much smaller than the latter, so we want to increase it rather than decrease it using a factor.

pypesto/profile/profile_next_guess.py
```diff
@@ -206,15 +233,18 @@ def clip_to_bounds(step_proposal):
...
-        step_length = problem.lb_full[par_index] - x[par_index]
-        return x + step_length * delta_x_dir
+        step_length = abs(problem.lb_full[par_index] - x[par_index])
+        return clip_to_bounds(x + step_length * delta_x_dir)
 
     if par_direction == 1 and (min_delta_x > problem.ub_full[par_index]):
-        step_length = problem.ub_full[par_index] - x[par_index]
-        return x + step_length * delta_x_dir
+        step_length = abs(problem.ub_full[par_index] - x[par_index])
+        return clip_to_bounds(x + step_length * delta_x_dir)
```
Doresic, line 247: This is where the bug with the exit condition was. Absolute value fixes it.

pypesto/profile/profile_next_guess.py
```diff
@@ -429,14 +525,36 @@ def do_line_search(
...
+        else:
+            direction = "increase"
+
+    elif next_obj < current_obj:
+        next_obj_target = low_next_obj_target
+        if next_obj <= low_next_obj_target:
+            direction = "decrease"
+            decreasing_to_low_target = True
+        else:
+            direction = "increase"
```
Doresic, line 556: If the initial guess is between the lower and higher target, we increase step size till we hit any of them. (exit conditions below)

If the initial guess is below lower target or above higher target, we decrease the step size till we hit the lower or higher target, respectively.

Doresic:
Randall_example_profile
Example showcasing the quality coloring

Doresic:
example_randall_obj_fun
This is the sample profile, but the plotted value is the objective function rather than the log-posterior ratio. 
Here it's easy to see how the steps at which `next_x` was resampled (green points) correspond to jumps in the objective function.

pypesto/profile/walk_along_profile.py
```diff
@@ -78,26 +79,168 @@ def walk_along_profile(
...
+                    n_sres=0,
+                    x0=np.array([]),
+                    fval0=fval,
+                    time=0,
+                )
+                optimizer_result.update_to_full(problem=problem)
+                optimization_successful = True
+                color_next = np.concatenate((color_now[:3], [0.3]))
+
+        if not optimization_successful:
```
PaulJonasJost, line 158: this is a lot of intendation. It might be savable with a `if optimization_sucessful: current_profile.append... return current profile`?

pypesto/profile/walk_along_profile.py
```diff
@@ -78,26 +79,168 @@ def walk_along_profile(
...
+            # Cannot optimize successfully by reducing max_step_size
+            # Let's try to optimize by increasing min_step_size
+            logger.warning(
+                f"Failing to optimize at {problem.x_names[i_par]}={x_next[i_par]} after reducing max_step_size."
+                f"Trying to increase min_step_size."
+            )
+
+            min_step_increase_factor = 1.25
+
+            while not optimization_successful:
```
PaulJonasJost, line 168: a while loop in python is not entered initially if the condition is directly wrong. Thus this whole part can be put with one less intendation

pypesto/profile/walk_along_profile.py
PaulJonasJost: with below comment this might not be necessary

PaulJonasJost (approved):
Looks mostly good to me. Would try to decrease intendation wherever possible for easier understanding

pypesto/profile/options.py
Doresic: True...
It is kinda the scaling factor of the adaptive target value.
Changed to `adaptive_target_scaling_factor`.

pypesto/profile/profile.py
```diff
@@ -104,6 +104,8 @@ def create_next_guess(
             current_profile_,
             problem_,
             global_opt_,
+            min_step_increase_factor_,
+            max_step_reduce_factor_,
```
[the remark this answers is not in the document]
Doresic: None of them are really. All "hyperparameters" a user can provide are in `ProfileOptions` or`next_guess_method`. Nothing outside of them should be touched. That's why they're not even mentioned in `parameter_profile`.

Specifically, these two are used to control the reduction of max_step_size and increase of min_step_size in case optimization fails. They are specified by the algorithm in `walk_along_profile.py`, never by the user.

pypesto/profile/profile.py
```diff
@@ -24,7 +24,7 @@ def parameter_profile(
     profile_index: Iterable[int] = None,
     profile_list: int = None,
     result_index: int = 0,
-    next_guess_method: Union[Callable, str] = "adaptive_step_regression",
+    next_guess_method: Union[Callable, str] = "adaptive_step_order_1",
```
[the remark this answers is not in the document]
Doresic: Hmm not sure. In all testing I did, it did perform worse than order 1. But it might be still that I just didn't find an example in which it just might be better. So I wouldn't remove it just to be able to test further, as I will be working on some further improvements of profiling soon :)

pypesto/profile/profile_next_guess.py
```diff
@@ -395,12 +482,16 @@ def do_line_search(
     step_size_guess: float,
     par_extrapol: Callable,
     next_obj: float,
-    next_obj_target: float,
+    current_obj: float,
```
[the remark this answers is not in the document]
Doresic: Yeah, basically.
There're two targets now, so need to extend it in that way -- exit conditions, interpolation etc.
And there are the min/max increase/reduction factors that change the "hyperparameters".

pypesto/profile/walk_along_profile.py
Doresic: Yeepp true...

pypesto/profile/options.py
Doresic: Take a look at the documentation of it in ProfileOptions. Not sure if informative enough.

pypesto/visualize/profiles.py
```diff
@@ -98,17 +119,30 @@ def profiles(
                 # multiple results per axes object
                 color_ind = i_result
 
+            # If debugging_colors is set to True, we use the colors provided
```
m-philipps, line 122:
```suggestion
            # If quality_colors is set to True, we use the colors provided
```

pypesto/visualize/profiles.py
```diff
@@ -215,6 +253,12 @@ def profiles_lowlevel(
         # if we have empty profiles and more axes than profiles: skip
         if n_plots != n_fvals and fval is None:
             continue
+        # If we use colors from profiler_result.color_path,
+        # we need to take the color path of each profile
+        if isinstance(color, list) and isinstance(color[i_plot], np.ndarray):
```
m-philipps, line 258: Why a `np.array` here? I no other data structure works, that should be reflected in the type hints of `profiles` and `profiles_lowlevel`, otherwise it could be a more generic sequence type

pypesto/visualize/profiles.py
```diff
@@ -215,6 +253,12 @@ def profiles_lowlevel(
...
+            i_color = color[i_plot]
+        else:
+            i_color = color
```
m-philipps, line 261: Just a personal preference but overall we tend to use `i_something` as an index so this could instead be `color_i`

pypesto/visualize/profiles.py
m-philipps: Is the second condition indirectly checking if the colors are from `profiler_result.color_path`?

m-philipps (changes requested):
Thanks for the improvement! I checked the `visualize` file.

vwiela (approved):
Looks good to me, thanks for improving.

pypesto/visualize/profiles.py
Doresic: > Is the second condition indirectly checking if the colors are from `profiler_result.color_path`?

Yeah, basically.
There are 3 cases:
- `color = None`, then `color_i` will be set to `None` for some default later in `profile_lowlevel`. This is the case if `quality_colors=False` and there's only one result and one profile_list_id (small digression here: one result can have multiple profile results, for instance, in a case when you ran profiling in all local minima you wanted. Then each of those profile results corresponds to one profile_list_id).
- `color = np.ndarray` of length 4 -- one RGBA color. This is the case if there's more than one result or more than one profile_list_id. Then the color cannot be the default one but are specified beforehand by  `process_result_list_profiles`.
- `color = list[np.ndarray[np.ndarray]]` -- this is the case if we use color_paths (quality_colors). The list iterates over profiled parameters, and each entry is the numpy array of RGBAs of length 4. 

In the first two cases, we need to pass the color directly as the color_i (color of the specific i-th profile) -- either None or just one RGBA.
In the last case, we need to select the color_path of the corresponding parameter profile we're plotting now.

And it seemed the easiest to check whether it's a list and its entry is a numpy array.
I could use `if color is not None and not isinstance(color, np.ndarray)` as well, but it's kinda the same (and I think I might've tried that first, and had a bug with a weird edge case).

pypesto/visualize/profiles.py
Doresic: Like it, it's nicer

pypesto/visualize/profiles.py
Doresic: I probably should change the docstring tho... As currently it just says that it can be RGBA or None

Doresic: Suggestion:
```
color: RGBA, list[np.ndarray[RGBA]], optional
        Color for profiles in plot. In case of quality_colors=True, this is a list of
        np.ndarray[RGBA] for each profile -- one color per profile point for each profile.
```

m-philipps: My issue is rather that there are two different options for coloring: the `quality_colors` and the `colors` argument of `profiles`.
1. If a user supplies a list of `np.ndarray`, that should not be interpreted as `quality_color` here
2. If the `profile_lowlevel` only takes a list of `np.ndarray`, that should be reflected in the type hint in lines 47-48 (old).

Doresic: Yes, there are now two different coloring options. But I do check and raise a ValueError if both of them are given.

`profiles_lowlevel` could previously (old code) take only a single np.ndarray (one RGBA color). It was never given a list of np.ndarray. When a user provides a list of `np.ndarray` (a list of multiple colors) then one color represents profiles of either one result or one profile_list_id -- `profiles_lowlevel` still gets only one color.

And now, the only case at which the `profiles_lowlevel` takes a list of np.ndarrays is when `quality_colors` is used.

So it should never clash. But I can put in the docstring of `colors` (47-48 old) that it cannot be used together with `quality_colors`.

I hope I haven't misunderstood you.

stephanmg: After this corrections, `visualize` looks fine to me. Thanks for the contribution.

stephanmg (approved):

m-philipps (approved):

pypesto/visualize/profiles.py
```diff
@@ -45,7 +47,7 @@ def profiles(
         List of reference points for optimization results, containing at
         least a function value fval.
     colors:
-        List of colors, or single color.
+        List of colors, or single color. Cannot be provided if quality_colors is set to True.
```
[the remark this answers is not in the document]
Doresic: Will edit it and push what it should be. As it can be one result with multiple colors (one per profile_list_id)

PR #1453, opened 2024-09-04
Removing Aesara from readthedocs

PaulJonasJost:
tried installing pypesto[aesara] in a complete new venv, which already threw errors. Removing it from the documentation for now (potentially completely from the package later depending on usage)


Doresic (approved):
LGTM!

PR #1457, opened 2024-09-13
SacessOptimizer: Fix acceptance threshold for objective improvement

dweindl:
There seems to be a difference in the acceptance threshold for objective improvement that is reported in the sacess paper and the value that is used in the implementation. Changing to the value used in the original implementation.

For the manager:
[Paper](https://doi.org/10.1186/s12859-016-1452-4): 10%
Implementation accompanying the paper: [0.5%](https://bitbucket.org/DavidPenas/sacess-library/src/18b2027af5b97158e5d5303487455e5e529bcf5a/src/method_module_fortran/eSS/parallelscattersearchfunctions.f90#lines-388) 
More [recent](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1011151) implementation: [0.01%](https://bitbucket.org/DavidPenas/sacess-library/src/508e7ac15579104731cf1f8c3969960c6e72b872/src/method_module_fortran/eSS/parallelscattersearchfunctions.f90#lines-396)
Also added a check for the minimum acceptance threshold as in the original implementation. 

For the workers: 
should be 0.01% instead 0.01.


stephanmg:
Might it be good to have this as a user-controllable parameter then?

stephanmg (approved):
Fine.

dweindl:
> Might it be good to have this as a user-controllable parameter then?

Agreed. Will do that along with a couple of other hyperparameters in a separate PR (#1458).

Issue #1463, opened 2024-09-16
`petab.create_objective` with custom edatas is broken


FFroehlich:
`petab.create_objective` allows the user to pass custom `edatas` (which could have potentially different length/order than the original ones) but then proceeds generates the default parameter mapping https://github.com/ICB-DCM/pyPESTO/blob/ead29b36db80d2eeee91b5d29bfbf4430137f8aa/pypesto/petab/importer.py#L470

Might be fixed by simply passing an updated `simulation_conditions` object based on the provided `edatas`.

PR #1464, opened 2024-09-16
attempt fix + add test

FFroehlich:
see #1463


FFroehlich:
Looks like `amici.petab.parameter_mapping.create_parameter_mapping` is not generating the expected mapping. Needs further investigation.

Issue #1396, opened 2024-05-07
Frequent readthedocs timeout


PaulJonasJost:
At least in my recent experience, a lot of times readthedocs seams to time out during building process. Something to keep track about.

dweindl:
This is mostly about the increasing number of notebooks. They should be as simple and orthogonal as possible. And use cheap models as far as possible.

PaulJonasJost:
again facing frequent timeout. Will go through the notebooks and look into it, also looking into #1460

PaulJonasJost:
closing this for now until the timeout problem arises again

PR #1470, opened 2024-09-18
Brushing over notebook

### LINKED ISSUE #1460: Frequent failures in model_evidence_and_bayes_factors.ipynb
Frequent failures in model_evidence_and_bayes_factors.ipynb, e.g. in https://readthedocs.org/projects/pypesto/builds/25637907/:
```
Notebook error:
CellExecutionError in example/model_evidence_and_bayes_factors.ipynb:
------------------
labels = [
        '-1/2 BIC',
        'Arithmetic Mean',
        'Laplace', 
        'Harmonic Mean', 
        'Stabilized\nHarmonic Mean',
        'Bridge Sampling',
        'Nested Sampling',
        'Thermodynamic\nIntegration',
        'Steppingstone\nSampling',
        'Variational Inference\nHarmonic Mean'
]

bayes_factors = [
    -1/2*models['mixture_model1']['BIC']+1/2*models['mixture_model2']['BIC'],
    models['mixture_model1']['arithmetic_log_evidence']-models['mixture_model2']['arithmetic_log_evidence'],
    models['mixture_model1']['laplace_evidences'][0]-models['mixture_model2']['laplace_evidences'][0],
    models['mixture_model1']['harmonic_log_evidence']-models['mixture_model2']['harmonic_log_evidence'],
    models['mixture_model1']['harmonic_stabilized_log_evidence']-models['mixture_model2']['harmonic_stabilized_log_evidence'],
    models['mixture_model1']['bridge_log_evidence']-models['mixture_model2']['bridge_log_evidence'],
    models['mixture_model1']['nested_log_evidence']-models['mixture_model2']['nested_log_evidence'],
    models['mixture_model1']['thermodynamic_log_evidence']-models['mixture_model2']['thermodynamic_log_evidence'],
    models['mixture_model1']['steppingstone_log_evidence']-models['mixture_model2']['steppingstone_log_evidence'],
    models['mixture_model1']['vi_harmonic_log_evidences']-models['mixture_model2']['vi_harmonic_log_evidences']
]

true_bf = models['mixture_model1']['true_log_evidence'] - models['mixture_model2']['true_log_evidence']
------------------

[0;31mKeyError[0m: 'thermodynamic_log_evidence'
```

The KeyError occurs after sampling fails with 
```
ValueError: Not all chains have converged yet. You may want to use a larger number of samples, or try ´use_all_chains=False´, which might increase the integration error.
```

See also https://github.com/ICB-DCM/pyPESTO/actions/runs/10882550532/job/30193829942?pr=1459

PaulJonasJost:
Mainly:
- shortening many optimizations
- Replacing Boehm with conversion reaction where possible
- Thermodynamic more robust.

Hopefully closes #1460 and #1396


doc/example/amici.ipynb
```diff
@@ -623,21 +713,25 @@
...
     "\n",
-    "For more details on PEtab, the interested reader is referred to [PEtab's format definition](https://petab.readthedocs.io/en/latest/documentation_data_format.html), for examples the reader is referred to the [PEtab benchmark collection](https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab). The Model from _[Böhm et al. JProteomRes 2014](https://pubs.acs.org/doi/abs/10.1021/pr5006923)_ is part of the benchmark collection and will be used as the running example throughout this notebook.\n"
+    "For more details on PEtab, the interested reader is referred to [PEtab's format definition](https://petab.readthedocs.io/en/latest/documentation_data_format.html), for examples the reader is referred to the [PEtab benchmark collection](https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab). For demonstration purposes, a simple model of conversion-reaction will be used as the running example throughout this notebook.\n"
```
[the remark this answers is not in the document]
Doresic: Although, I'm not sure if it's a good idea to change the model in this notebook as the Boehm model is used in the first part of the notebook. Swapping the model in the middle of the notebook is unnecessarily confusing.

doc/example/amici.ipynb
```diff
@@ -806,34 +930,9 @@
...
-   },
-   "outputs": [],
-   "source": [
-    "startpoint_method = pypesto.startpoint.uniform"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {
-    "collapsed": false,
```
Doresic, line 836: I think info should be given somewhere at least. However, I'm not sure where it's best to put it. 
This notebook already gives more-detailed information about optimization, history, etc... so it doesn't feel out of place in here.
I would not put it in getting-started as I don't see it as essential enough.

doc/example/amici.ipynb
```diff
@@ -907,13 +1018,16 @@
...
+     "outputs_hidden": false
+    },
     "pycharm": {
      "name": "#%%\n"
     }
    },
    "outputs": [],
    "source": [
-    "n_starts = 20  # usually a value >= 100 should be used\n",
+    "n_starts = 10  # usually a value >= 100 should be used\n",
```
Doresic, line 1030: I would just check that these 10 starts with the set random seed give a good model fit. It's kinda important to have fast and good sampling and profiling.

Also, is 10 enough to have a parameter histogram and scatter plot that looks like something nice?

doc/example/amici.ipynb
Doresic: Did we deprecate this way of specifying the standpoint method? If so, I would update this then to what should be used.

doc/example/conversion_reaction.ipynb
```diff
@@ -157,7 +158,7 @@
     "\n",
     "# do the optimization\n",
     "result = optimize.minimize(\n",
-    "    problem=problem, optimizer=optimizer, n_starts=100, filename=None\n",
+    "    problem=problem, optimizer=optimizer, n_starts=10, filename=None\n",
```
Doresic, line 161: is this sufficient to get nice results always? There's no seed in this notebook. So it should be nice for any startpoints.

doc/example/conversion_reaction.ipynb
```diff
@@ -270,7 +271,7 @@
     ")\n",
     "\n",
     "result = sample.sample(\n",
-    "    problem, n_samples=10000, sampler=sampler, result=result, filename=None\n",
+    "    problem, n_samples=1000, sampler=sampler, result=result, filename=None\n",
```
Doresic, line 274: Similarly as above, is it sufficent? Don't need to answer, but just check the figures and what they look like.
Recently, when I was updating notebooks, I was choosing the lowest numbers that didn't alter the figures too much -- such that it still looks presentable. So I would just make sure it stays that way if we're reducing it further.

doc/example/custom_objective_function.ipynb
Doresic, line 1: I just noticed that we mention custom objective functions in the getting-started notebook, but we do not link or mention that there is another notebook going into detail on this. It would be nice to do so, no?

doc/example/getting_started.ipynb
```diff
@@ -157,27 +166,79 @@
...
+       " 3.76158192263132e-36,\n",
+       " 6.018531076210112e-36,\n",
+       " 9.780112998841433e-36]"
+      ]
+     },
+     "execution_count": 42,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
```
Doresic, line 241: Remove all outputs from the notebook

Doresic (changes requested):
Mostly looks good! Thanks.

I added a few comments of the same type about the reduction of starts of optimization and samples of sampling. I think it's good to reduce as much as we can so it takes slower. But I would not reduce so low that the figures look less presentable or nice. If the histograms start being just random bars it's not much use showing them. So if you didn't already, check whether the figures stay the same in character -- especially if there's no random seed that controls startpoints and how well the model is fit.

doc/example/amici.ipynb
PaulJonasJost: the startpoint method is now defined as a property of the problem/objective, it is not used as an argument to optimization anymore

Doresic: Could we include this change here? I don't think we show it anywhere else.

doc/example/conversion_reaction.ipynb
PaulJonasJost: conversion reaction looks good imo 👍🏼

doc/example/amici.ipynb
PaulJonasJost: think problem would here be more that we do find optima veeery well and hence the correlation plot looks hideous. You think we should switch back to boehm? Histogram looks soso

Doresic: Yeah, if it makes the notebook, the figures and conclusions taken from them make more sense, I would go for that rather than it being a bit faster

doc/example/amici.ipynb
dilpath, line 1: Change PEtab image to use the new `v1` path https://github.com/PEtab-dev/PEtab/blob/main/doc/v1/gfx/petab_files.png

doc/example/getting_started.ipynb
```diff
@@ -54,11 +55,11 @@
    "source": [
     "## 1. Objective Definition\n",
     "\n",
-    "PyPESTO allows the definition of a custom objectives, as well as offers support for objectives defined in the [PEtab](https://github.com/PEtab-dev/PEtab) format.\n",
+    "PyPESTO allows the definition of custom objectives and offers support for objectives defined in the [PEtab](https://github.com/PEtab-dev/PEtab) format.\n",
```
dilpath, line 58:
```suggestion
    "pyPESTO allows the definition of custom objectives and offers support for objectives defined in the [PEtab](https://github.com/PEtab-dev/PEtab) format.\n",
```

dilpath (approved):
Fine for me, thanks!

> I would not reduce so low that the figures look less presentable or nice.

re: this balance between computation time of CI vs. sufficient computation for nice figures, we could use some environment variable (e.g. `CI` and `GITHUB_ACTIONS` are always set to true during GitHub actions CI) to determine the computation effort in each notebook. This could look like this at the top of the notebooks.

```python
import os
n_starts = 100
n_samples = 10000
if os.environ.get("CI", None) is not None:
    n_starts = 10
    n_samples = 1000

Doresic:
Ohh I just remembered.
At the retreat, we did talk about visualization again, and Jan again mentioned how it's always good to know in which order one should look at visualizations: Firstly, it's important to see that the model actually fits the data. If that's not true, waterfall plot and parameter plot are not important. Only when the data is somewhat fit, it's nice to look at the waterfall plot, to see convergence and parameter plot to see a fast-less-informative indication of identifiability and whether the bounds are good enough (actually changing the parameter bounds might be necessary to fit the data in the first place). Only then it makes sense to go into profiling and sampling -- to get some more accurate insight into parameter identifiability.

It would be nice if this was implemented in the getting_started notebook as a somewhat guide to estimating models. This might take some restructuring of the notebook, so I'm also ok with doing it myself in a separate PR. Wanted to mention it here so it's not forgotten :D

PR #1477, opened 2024-10-02
ESSOptimizer: Fix recombination bug

dweindl:
This fixes a bug in the recombination step of the `ESSOptimizer`, also used by `SacessOptimizer`.

This will in particular improve performance when using scatter search without local optimizers.

### DIFF
--- a/pypesto/optimize/ess/ess.py
+++ b/pypesto/optimize/ess/ess.py
@@ -443,11 +443,15 @@ class ESSOptimizer:
             raise ValueError("i == j")
         x = self.refset.x
 
-        d = x[j] - x[i]
-        alpha = np.sign(j - i)
+        d = (x[j] - x[i]) / 2
+        # i < j implies f(x_i) < f(x_j)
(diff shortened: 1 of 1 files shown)


pypesto/optimize/ess/ess.py
```diff
@@ -443,11 +443,15 @@ def _combine(self, i, j) -> np.array:
             raise ValueError("i == j")
         x = self.refset.x
 
-        d = x[j] - x[i]
-        alpha = np.sign(j - i)
+        d = (x[j] - x[i]) / 2
```
PaulJonasJost, line 446: why was this changed to /2?

PaulJonasJost (approved):
i am guessing that line 450 created the bug?

pypesto/optimize/ess/ess.py
dweindl: Because it was wrong :)

dweindl:
> i am guessing that line 450 created the bug?

L446 + L450

PR #1480, opened 2024-10-02
ESSOptimizer: Fix bug in go-beyond

dweindl:
Fixes some issues related to copies vs views that stopped the go-beyond search prematurely after the first successful step.

### DIFF
--- a/pypesto/optimize/ess/ess.py
+++ b/pypesto/optimize/ess/ess.py
@@ -560,7 +560,7 @@ class ESSOptimizer:
     def _maybe_update_global_best(self, x, fx):
         """Update the global best value if the provided value is better."""
         if fx < self.fx_best:
-            self.x_best = x[:]
+            self.x_best[:] = x
             self.fx_best = fx
             self.x_best_has_changed = True
             self.history.update(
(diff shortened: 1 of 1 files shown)


pypesto/optimize/ess/ess.py
```diff
@@ -583,9 +583,9 @@ def _go_beyond(self, x_best_children, fx_best_children):
                 continue
 
             # offspring is better than parent
-            x_parent = self.refset.x[i]
+            x_parent = self.refset.x[i].copy()
```
PaulJonasJost, line 586: Do i see correctly that the problem was that `x_parent/child` are edited further below and thus edit the reset itself?

pypesto/optimize/ess/ess.py
```diff
@@ -560,7 +560,7 @@ def _do_local_search(
     def _maybe_update_global_best(self, x, fx):
         """Update the global best value if the provided value is better."""
         if fx < self.fx_best:
-            self.x_best = x[:]
+            self.x_best[:] = x
```
PaulJonasJost, line 563: how is this related to the go-beyond algorithm?

PaulJonasJost (approved):
Looks good, thanks. I added some questions for my own clarification 🙏🏼

pypesto/optimize/ess/ess.py
dweindl: It's not. But it's another potential copying issue.

pypesto/optimize/ess/ess.py
dweindl: The problem was not so much editing the refset per se, but that in the second inner iteration, `x_parent ` would become identical to `x_child`.

Issue #1475, opened 2024-09-30
Reduce computation and non-deterministic behaviour in evidence and sampling tests


dilpath:
Continuing on #1461 / #1473
It looks like there is unnecessary computation in the evidence tests. There are `optimize.minimize` and `sample.sample` calls there. These could be done once locally and the samples stored, to speed up the tests and remove most sources of non-deterministic behavior in the tests. The stored samples could also be reused in `test_samples_ci`, `test_ground_truth`, `test_autocorrelation_pipeline` too, since sample generation is already tested in `test_pipeline`.

The remaining non-deterministic thing would be in the bridge sampling itself, e.g.
https://github.com/ICB-DCM/pyPESTO/blob/61c912d56b4e435105da9e95dbdb31ae1303a807/pypesto/sample/evidence.py#L348-L352

This could be resolved with a random seed, ideally like
```python
def bridge_sampling_log_evidence(..., rng: np.random.Generator):
    ...
    proposal_samples = rng.normal(
    ...
```
so that users can specify a random seed via numpy's newer/recommended rng.

There are some other things that aren't clear to me:
- the same problem isn't used in each evidence computation test -- why not?
- why is the "real" evidence only used in `test_harmonic_mean_log_evidence`, but not in `test_bridge_sampling`?

PR #1486, opened 2024-10-10
Clean up sampling tests

PaulJonasJost:
- Moved problem definitions into a utility file
- Sped up Dynasty sampling in pipeline (kudos to Jonas)
- Removed the separated modes test, as it was less of a test, more of a showcase - covered in sampler_study.ipyn
- Moved harmonic mean to other evidence calculations

Should speed up most of the very slow tests in `base`, clean up the code and thereby adress #1475

### DIFF
--- a/test/sample/test_sample.py
+++ b/test/sample/test_sample.py
@@ -85,7 +85,7 @@ def sampler(request):
     elif request.param == "Pymc":
         from pypesto.sample.pymc import PymcSampler
 
-        return PymcSampler(tune=5, progressbar=False)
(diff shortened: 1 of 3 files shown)


arrjon (approved):
LGTM

dilpath (approved):
Thanks for the many improvements, including speeding things up! I don't think this addresses some of the unnecessary computation and non-deterministic behavior described in #1475 , but that can be revisited if failing tests become an issue again. Fine to merge and close/leave open that issue as you like.

PR #1489, opened 2024-10-17
Updated PEtab.jl to work with version 3.x

PaulJonasJost:


sebapersson:
Thanks for updating, everything looks good for me!

vwiela (approved):

PaulJonasJost:
removed the julia specific caching. We faced a lot of problems repeatedly due to that. Installing dependencies went from 9:45 (in [gradient check example](https://github.com/ICB-DCM/pyPESTO/actions/runs/11459496220/job/31884116566) to 11:15. In my opinion the 2min extra (plus fluctuations obviously) are worth the less hassle (90% sure a lot of julia fails stemmed from the julia caching). What do you guys think @dweindl @Doresic @vwiela @dilpath?

Issue #1448, opened 2024-08-15
How to set `fill_fixed_parameters`=True for petab parameter mappings?


FFroehlich:
So there is https://github.com/ICB-DCM/pyPESTO/blob/ead29b36db80d2eeee91b5d29bfbf4430137f8aa/pypesto/petab/importer.py#L476, which I don't like as this means that edata.plist will always be all parameters instead of subsetting to the ones that actually require sensitivities.

Now the best way that I found to fix this is
```
pypesto_problem = importer.create_problem()
simulation_conditions = petab.get_simulation_conditions(
    importer.petab_problem.measurement_df
)
pypesto_problem.objective.parameter_mapping = amici.petab_objective.create_parameter_mapping(
    petab_problem=importer.petab_problem,
    simulation_conditions=simulation_conditions,
    scaled_parameters=True,
    amici_model=pypesto_problem.objective.amici_model,
    fill_fixed_parameters=True,
)
```
which is not great, because it requires me to have pretty deep knowledge of the code and requires a couple more lines of code than what I had hoped for.

Is there any specific reason why we are setting `fill_fixed_parameters=True`? If there is could we either expose this option to the user or set smarter defaults when this is really necessary?

Looks like this was introduced in https://github.com/ICB-DCM/pyPESTO/pull/912, but seems unrelated to the linked issue https://github.com/ICB-DCM/pyPESTO/issues/882. Unclear what "fix warning from fixed overrides" refers to.

dweindl:
Sorry for the late reply.

> Looks like this was introduced in #912, but seems unrelated to the linked issue #882. 

The linked issue is unrelated to `fill_fixed_parameters`, correct. It was linked because that other change in this PR (https://github.com/ICB-DCM/pyPESTO/issues/882#issuecomment-1204463465).

> Unclear what "fix warning from fixed overrides" refers to.

I don't recall that either. My guess would be: Before the change, pypesto was calling the amici/pypesto functions for filling parameter values according to the parameter mapping. The parameter mapping already set the values for the fixed parameters (`fill_fixed_parameters=True`). Thus, their symbols didn't show up in the parameter mapping anymore. Yet, pypesto still provided values for those parameters which triggered a warning that values were passed for unknown parameters. The PR was to prevent this warnings.


For anything based on the PEtab importer, changing `fill_fixed_parameters` should not be necessary.

> this means that edata.plist will always be all parameters instead of subsetting to the ones that actually require sensitivities

[...]

PR #1493, opened 2024-10-18
AmiciObjective/PEtab import: Fix plist

dweindl:
For further background, see #1448.

We don't have to compute the full gradient in every model simulation. Some entries might not be required because of fixed parameter or some condition-specific objective parameter mapping. The former was supported (however, not tested), the latter was not supported. Now both are tested an supported.

There was no good way to communicate the fixed parameters to AmiciCalculator where ExpData.plist gets set. Passing this information is currently only possible through the parameter mapping, based on which the required sensitivities are determined. Therefore, in addition to the general parameter mapping, there is now a parameter mapping that accounts for the fixed parameters. Not sure what could be a more elegant way to handle that.

### DIFF
--- a/pypesto/objective/amici/amici.py
+++ b/pypesto/objective/amici/amici.py
@@ -178,8 +178,12 @@ class AmiciObjective(ObjectiveBase):
             parameter_mapping = create_identity_parameter_mapping(
                 amici_model, len(edatas)
             )
+        # parameter mapping where IDs of the currently fixed parameters
+        #  have been replaced by their respective values
+        #  (relevant for setting ``plist`` in ExpData later on)
         self.parameter_mapping = parameter_mapping
-
+        # parameter mapping independent of fixed parameters
+        self._parameter_mapping_full = copy.deepcopy(parameter_mapping)
         # If supported, enable `guess_steadystate` by default. If not
         #  supported, disable by default. If requested but unsupported, raise.
         if (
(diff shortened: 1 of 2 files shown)


test/petab/test_petab_import.py
```diff
@@ -137,35 +136,76 @@ def test_check_gradients(self):
...
+    objective.amici_solver.setRelativeTolerance(1e-15)
+
+    # slightly perturb the parameters to avoid vanishing gradients
+    par = np.asarray(petab_importer.petab_problem.x_nominal_free_scaled) * 1.01
+
+    # call once to make sure ExpDatas are populated
+    fx1, grad1 = objective(par, sensi_orders=(0, 1))
+
+    plists1 = {edata.plist for edata in objective.edatas}
+    assert len(plists1) == 6, plists1
```
PaulJonasJost, line 172: why asser plists1 as well? isnt `len(plists1) == 6` enough?

pypesto/objective/amici/amici.py
```diff
@@ -178,8 +178,12 @@ def __init__(
             parameter_mapping = create_identity_parameter_mapping(
                 amici_model, len(edatas)
             )
+        # parameter mapping where IDs of the currently fixed parameters
+        #  have been replaced by their respective values
+        #  (relevant for setting ``plist`` in ExpData later on)
         self.parameter_mapping = parameter_mapping
-
+        # parameter mapping independent of fixed parameters
+        self._parameter_mapping_full = copy.deepcopy(parameter_mapping)
```
PaulJonasJost, line 186: Curious, but is that always the full one? What if I now start by creating the amici objective with fixed parameters and then unfix a parameter from the problem?

PaulJonasJost (approved):
Looks good to me. Also would not be aware of a better suitable/more elegant approach

test/petab/test_petab_import.py
dweindl: The second expression [is just the error message](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).

pypesto/objective/amici/amici.py
dweindl: > What if I now start by creating the amici objective with fixed parameters and then unfix a parameter from the problem?

How would you do that? In terms of `Problem.fix_parameters`, I am not sure how you could start with fixed parameters before having had an objective without fixed parameters.

> Curious, but is that always the full one?

I guess, that depends on the definition of "full". Currently, it's "full" in the sense of "all parameters that were considered to be estimated at the time the AmiciObjective was created". Ever trying to change that to something "fuller" afterwards is probably not a good idea.

PaulJonasJost: mhh you are right i guess. I just checked the petabImporter. Was concerned how the mappings is done if petab has fixed parameters.

dweindl:
ping @FFroehlich

FFroehlich (approved):
thanks 🙏

Issue #1487, opened 2024-10-13
Support `sensi_orders` in gradient checks


dilpath:
While debugging another model, the issue arose again that sometimes `Objective(x, sensi_orders=(0,))` provides a different `fval` than `Objective(x, sensi_orders=(0,1))`. Tolerances etc. can be adjusted until these two are similar; however, pyPESTO's `check_grad` might incorrectly report that gradients are not close, because it computes `fval` with sensis
https://github.com/ICB-DCM/pyPESTO/blob/ead29b36db80d2eeee91b5d29bfbf4430137f8aa/pypesto/objective/base.py#L535

but `fval_p` and `fval_m` (for the plus/minus steps used in finite difference approximations) without sensis
https://github.com/ICB-DCM/pyPESTO/blob/ead29b36db80d2eeee91b5d29bfbf4430137f8aa/pypesto/objective/base.py#L556
https://github.com/ICB-DCM/pyPESTO/blob/ead29b36db80d2eeee91b5d29bfbf4430137f8aa/pypesto/objective/base.py#L561

This could be resolved by using the same `sensi_orders` in all function calls, or otherwise providing some way for the user to specify this.

dweindl:
> This could be resolved by using the same `sensi_orders` in all function calls, or otherwise providing some way for the user to specify this.

It might make sense to address that independently of gradient checks, since this will also be a problem for any optimizer that calls the objective with varying `sensi_orders`. Maybe a simple wrapper around Objective that always passes a certain `sensi_orders` internally and then filters the result to match the externally requested  `sensi_orders`.

PR #1497, opened 2024-10-20
Gradient check example

### LINKED ISSUE #1481: Add gradient check methods to notebooks
Since it's generally recommended, it would be good to have it in the examples.

e.g. https://github.com/ICB-DCM/pyPESTO/blob/ead29b36db80d2eeee91b5d29bfbf4430137f8aa/pypesto/objective/base.py#L431

### LINKED ISSUE #1494: Conditional variable initialisation in `AmiciObjective.check_gradients_match_finite_differences`
Gradient check for an `AmiciObjective`
```python
pypesto_problem.objective.check_gradients_match_finite_differences(
        x=startpoints[0],
    )
```
invokes an error:
```
  File "<stdin>", line 1, in <module>
  File "/home/maren/pyPESTO/pypesto/objective/amici/amici.py", line 665, in check_gradients_match_finite_differences
    *args, x=x, x_free=x_free, **kwargs
                       ^^^^^^
UnboundLocalError: cannot access local variable 'x_free' where it is not associated with a value
```

This is because `x_free` is only created if the condition in [line 651](https://github.com/ICB-DCM/pyPESTO/blob/ead29b36db80d2eeee91b5d29bfbf4430137f8aa/pypesto/objective/amici/amici.py#L651) is `True`. 

Version `0.5.1`

m-philipps:
- add a documentation example for performing gradient checks (closes #1481)
- fix #1494

I'd be happy about suggestions on the "Best practices" and "How to fix my gradients".


doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,896 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### How to \"fix\" my gradients?\n",
+    "\n",
+    "- Find suitable simulation tolerances.\n",
+    "- Consider switching from adjoint to forward sensitivities, which tend to be more robust.\n",
```
dweindl, line 862: Most importantly, test your gradients using the settings you will use later on.

Also a comment related to https://github.com/ICB-DCM/pyPESTO/issues/1487 might be helpful.

dweindl (commented):
Thanks. I think I would just skip over `check_grad` and directly introduce `check_grad_multi_eps`. The latter performs much better.

m-philipps:
> I think I would just skip over `check_grad` and directly introduce `check_grad_multi_eps`. The latter performs much better.

I agree that `check_grad_multi_eps` is convenient to use in practice; I like showcasing `check_grad` first to build it up like a tutorial, such that it is easier to understand what `check_grad_multi_eps` is doing for people who aren't so familiar with gradient checks. I'm also fine with changing it though.

doc/example/gradient_check.ipynb
```diff
@@ -861,7 +858,8 @@
     "- Find suitable simulation tolerances.\n",
     "- Consider switching from adjoint to forward sensitivities, which tend to be more robust.\n",
     "- Check the simulation logs for Warnings and Errors.\n",
-    "- Ensure that the model is correct."
+    "- Ensure that the model is correct.\n",
+    "- Be aware that, depending on the tolerances, the gradient might differ when it is calculated with sensitivities, because `Objective(x, sensi_orders=(0,))` provides a different value than `Objective(x, sensi_orders=(0,1))`"
```
vwiela, line 862: Personal preference, but I would formulate it a bit less strict. Currently, it sounds for me as if the two calls always provide different values.

Maybe you use can in the second part of the sentence.

Be aware that, depending on the tolerances, the gradients might differ when the objective function value is calculated with sensitivities, because `Objective(x, sensi_orders=(0,))`  can provide a different value than `Objective(x, sensi_orders=(0,1))`.

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,894 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### How to \"fix\" my gradients?\n",
+    "\n",
+    "- Find suitable simulation tolerances.\n",
+    "- Consider switching from adjoint to forward sensitivities, which tend to be more robust.\n",
+    "- Check the simulation logs for Warnings and Errors.\n",
+    "- Ensure that the model is correct.\n",
+    "- Be aware that, depending on the tolerances, the gradient might differ when it is calculated with sensitivities, because `Objective(x, sensi_orders=(0,))` provides a different value than `Objective(x, sensi_orders=(0,1))`"
```
m-philipps, line 862:
```suggestion
    "- Be aware that, depending on the tolerances, the gradient might differ when it is calculated with sensitivities, because `Objective(x, sensi_orders=(0,))` might provide a different value than `Objective(x, sensi_orders=(0,1))`"
```

pypesto/objective/amici/amici.py
```diff
@@ -650,7 +650,9 @@ def check_gradients_match_finite_differences(
         """
         if x is None and "petab_problem" in dir(self.amici_object_builder):
             x = self.amici_object_builder.petab_problem.x_nominal_scaled
-            x_free = self.amici_object_builder.petab_problem.x_free_indices
+            kwargs = {
+                "x_free": self.amici_object_builder.petab_problem.x_free_indices
+            }
         return super().check_gradients_match_finite_differences(
-            *args, x=x, x_free=x_free, **kwargs
+            *args, x=x, **kwargs
```
PaulJonasJost, line 657: Is there a specific reason for those changes? I would feel like keeping an argument explicit always gives more info.

PaulJonasJost (approved):
Looks good to me, thanks for this.

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,894 @@
...
+   "cell_type": "markdown",
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "# Gradient checks\n",
+    "\n",
+    "It is best practice to do gradient checks before and after gradient-based optimization.\n",
```
dweindl, line 13: I think it would be good to include some rationale for why to check it afterwards and what to look for. I.e. except for parameters with active bounds, the values should be close to 0. At the same time, this might make it difficult to get good FD approximations.

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,894 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "#### Set up an example problem\n",
```
dweindl, line 48: I think the heading levels aren't consistent yet (https://pypesto--1497.org.readthedocs.build/en/1497/example/gradient_check.html)

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,894 @@
...
+    "\n",
+    "Create the pypesto problem and a random vector of parameter values."
```
dweindl, line 50: The latter only happens in the next section. Keep text/code coherent.

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,894 @@
...
+    "    verbosity = 0,\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Explanation of the gradient check result columns:\n",
+    "- `grad`: Objective gradient\n",
```
dweindl, line 268: Add blank line before listing for proper list rendering (see https://pypesto--1497.org.readthedocs.build/en/1497/example/gradient_check.html)

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,894 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### How to \"fix\" my gradients?\n",
+    "\n",
+    "- Find suitable simulation tolerances.\n",
+    "- Consider switching from adjoint to forward sensitivities, which tend to be more robust.\n",
```
dweindl, line 859: I am not sure whether to include that ASA/FSA point. Generally, ASA should be able to provide accurate gradients too.

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,894 @@
...
+    ").background_gradient(\n",
+    "    cmap=sns.light_palette(\"red\", as_cmap=True), subset=[\"rel_err\"],\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### How to \"fix\" my gradients?\n",
```
dweindl, line 856: Maybe it should be distinguished here whether one brought one's own gradient function or whether one is using the petab-amici-pipeline. Most points only apply to the latter.

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,894 @@
...
+    "\n",
+    "- Find suitable simulation tolerances.\n",
+    "- Consider switching from adjoint to forward sensitivities, which tend to be more robust.\n",
+    "- Check the simulation logs for Warnings and Errors.\n",
+    "- Ensure that the model is correct.\n",
```
dweindl, line 861: I guess, the question is not so much about the model. With a wrong model, we should still get accurate gradients. It's more about the correctness of the gradient computation.

pypesto/objective/amici/amici.py
m-philipps: This was necessary to fix the error described in #1494

doc/example/gradient_check.ipynb
m-philipps: Alternative suggestion:
`Ensure that the model can produce the expected gradients.`

doc/example/gradient_check.ipynb
m-philipps: You mean, a 0 gradient makes the FD approximation difficult?

doc/example/gradient_check.ipynb
m-philipps: I got the input that switching to FSA made optimisation possible on some occasions. 
Is it improper to say, FSA "fixes gradients"? Should it be narrowed down, why FSA is more robust in practice?

doc/example/gradient_check.ipynb
dweindl: Yes, that, and one might miss gradient entries that are (incorrectly) always zero.

doc/example/gradient_check.ipynb
dweindl: I think something like "try the other method" is fine, but I am not sure about the general statement on increased robustness of FSA. Sometimes ASA may perform better, sometimes FSA. If there is a good rationale for why FSA would be more robust then please include that.

doc/example/gradient_check.ipynb
dweindl: I am not sure how helpful that is. It boils down to "if your gradients are wrong, fix your gradients", doesn't it?

doc/example/gradient_check.ipynb
m-philipps: Shouldn't these already show up before optimisation?

doc/example/gradient_check.ipynb
m-philipps: The point is that the structural model might be the reason for issues in simulation. 
I'm not aware of a specific example though, maybe when events are involved?

doc/example/gradient_check.ipynb
m-philipps: No, I'm not aware of a rationale, the hint above was my takeaway from the discussion we had in the subgroup meeting.

doc/example/gradient_check.ipynb
dweindl: Yes, they should.

doc/example/gradient_check.ipynb
dweindl: Then I'd just drop the second half.

doc/example/gradient_check.ipynb
dweindl: Okay, I see what you mean. This would have to be explained in more detail and might get out of scope here.  In this short form, I don't find it too helpful and would rather drop it.

vwiela (approved):

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,903 @@
...
+       "k_exp_hetero          2.010635e+08  1.805003e+07  9.998990e-01  \n",
+       "k_exp_homo            3.383970e+07  7.345596e+06  7.910607e-01  \n",
+       "k_imp_hetero          4.382216e+08  1.415004e+08  1.196396e-01  \n",
+       "k_imp_homo            1.211021e+08  9.630702e+06  3.501887e-03  \n",
+       "k_phos                8.509938e+07  1.611509e+07  5.058878e-04  \n",
+       "sd_pSTAT5A_rel        9.372550e+07  6.663599e+02  1.622721e-10  \n",
+       "sd_pSTAT5B_rel        8.401874e+07  3.362799e+01  1.362745e-10  \n",
+       "sd_rSTAT5A_rel        9.538284e+07  3.640203e+01  4.970090e-01  "
+      ]
+     },
```
Doresic, line 253: Do you think we should add some conclusion on this or one of the gradient checks? In the sense what would we think if we saw this gradient check: the absolute errors are rather high but can be expected with finite differences on random points. We see this also with fd_err, so changing the eps might make sense.
But what's reassuring is that the relative error is not too high in most cases.

Would redoing the gradient check with another eps make sense to show how it can affect the check?

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,903 @@
...
+   "source": [
+    "### How to \"fix\" my gradients?\n",
+    "\n",
+    "- Find suitable simulation tolerances.\n",
+    "\n",
+    "Specific to the petab-amici-pipeline:\n",
+    "\n",
+    "- Check the simulation logs for Warnings and Errors.\n",
+    "- Consider switching between forward and adjoint sensitivity algorithms.\n",
+    "- Be aware that, depending on the tolerances, the gradient might differ when it is calculated with sensitivities, because `Objective(x, sensi_orders=(0,))` might provide a different value than `Objective(x, sensi_orders=(0,1))`"
```
Doresic, line 871: When you say the gradient might differ when it's calculated with sensitivities you mean that about the finite differences gradient? The simulations can differ when calculated with sensitivities (sensi_orders=(0,1,)) or without (sensi_orders=(0,)). Then the FD can differ as well. 
But the pypesto gradient we get only with sensi_orders=(0,1,)

Doresic (approved):
Added some comments, more in the direction of how would one interpret the results of gradient checks. Because sometimes we get high absolute errors, but relative are still ok, so is this an issue, or not? Some guidance for users.

Not sure if we want to add that type of content to this notebook tho, or just want to leave it only as -- "This can be done and it's done using these functions."

doc/example/gradient_check.ipynb
m-philipps: Removed this here and opened an issue #1547

pypesto/objective/amici/amici.py
```diff
@@ -667,9 +667,11 @@ def check_gradients_match_finite_differences(
         """
         if x is None and "petab_problem" in dir(self.amici_object_builder):
             x = self.amici_object_builder.petab_problem.x_nominal_scaled
-            x_free = self.amici_object_builder.petab_problem.x_free_indices
+            kwargs = {
+                "x_free": self.amici_object_builder.petab_problem.x_free_indices
+            }
```
dweindl, line 672:
```suggestion
            kwargs["x_free"] = self.amici_object_builder.petab_problem.x_free_indices
```
The current implementation would just ignore any other kwargs.

doc/example/gradient_check.ipynb
```diff
@@ -0,0 +1,1201 @@
...
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
+   "source": [
+    "# Gradient checks\n",
+    "\n",
+    "It is best practice to do gradient checks before and after gradient-based optimization.\n",
+    "\n",
+    "1. Find suitable tolerances to use during optimisation. Importantly, test your gradients using the settings you will use later on.\n",
```
dweindl, line 15:
```suggestion
    "1. Find suitable tolerances to use during optimization. Importantly, test your gradients using the settings you will use later on.\n",
```

dweindl (approved):
Looks good to me after fixing the kwargs issue and clearing the notebook outputs.

pypesto/objective/amici/amici.py
m-philipps: Thanks!

PR #1503, opened 2024-11-04
ESSOptimizer: Fix priority for local search startpoints

dweindl:
Fixes a bug in ESSOptimizer that lead to incorrect priorities for the selection of candidate solutions for local searches. A numpy array was incorrectly initialized from a generator, which led to the `balance` hyperparameter being practically ignored, and candidate solutions being ranked only by objective function value.


dweindl:
> So it's better to use `from_iter` in cases when we want to use the output as a numpy array rather than a generator? Because generators sometimes mess up as they don't behave like numpy arrays in all cases?

Not just better, but necessary. `numpy.array(some_generator)` will not execute the generator, but store it as such. Not only generators, but all iterators.

PR #1504, opened 2024-11-11
Hierarchical: fix no error if inner observable parameter in noise formula & viceversa

Doresic:
If an inner observable parameter was present in the noise formula, there was no error, even though that's not compatible with hierarchical optimization. This was caused by not searching for observable placeholders, overrides, and inner parameters in noise formulas and vice versa.


dweindl (approved):
Looks good to me.

PR #1505, opened 2024-11-12
Hierarchical: remove inner datas from relative calculator

Doresic:
Don't think there's any need for the `inner_rdatas` as well as `rdatas` in the output of the calculator. Looks like an artifact of something from before.


dweindl (approved):
I don't see any need for it either.

PaulJonasJost (approved):
also dont see a need.

PR #1510, opened 2024-11-18
Fix ESSOptimizer min of empty sequence

dweindl:
Fixes `ValueError: min() arg is an empty sequence)` that may cause deadlocks.


vwiela (approved):
Good fix for now, thanks.

PR #1515, opened 2024-11-20
Visualize: fix flatten of observable mapping with one observable

Doresic:
Observable mapping visualization failed in the case of only one observable. Fixed to account for the case of one observable.


stephanmg:
Looks good to me.

vwiela (approved):
Looks good, thanks for the fix.

PaulJonasJost (approved):
Should add a test for future cases probably

dweindl:
I think there are two more such cases in `pypesto/visualize/observable_mapping.py` and one in `pypesto/visualize/ordinal_categories.py`.

test/visualize/test_visualize.py
```diff
@@ -1240,3 +1241,54 @@ def test_parameters_correlation_matrix(result_creation):
...
+        / "doc"
+        / "example"
+        / "example_ordinal"
+        / "example_ordinal.yaml"
+    )
+    petab_problem = petab.Problem.from_yaml(example_ordinal_yaml)
+    # Set seed for reproducibility.
+    np.random.seed(0)
+    optimizer = pypesto.optimize.ScipyOptimizer(
+        method="L-BFGS-B", options={"maxiter": 10}
```
dweindl, line 1261: For just testing that visualization does not produce an error, a single iteration would suffice, right?

dweindl (approved):

stephanmg (approved):
Thanks for the contribution. From my side this can be merged.

PR #1517, opened 2024-11-25
Improve exception-handling in SacessOptimizer

### LINKED ISSUE #1512: Improve exception-handling in SacessOptimizer
Currently, uncaught exceptions in `SacessWorker`s will result in deadlocks in `SacessOptimizer.minimize`. This needs to change.

dweindl:
Previously, uncaught exceptions in SacessOptimizer worker processes would have resulted in deadlocks in `SacessOptimizer.minimize`.

These changes will (in most cases) prevent deadlocks and other errors due to missing results from individual workers. Furthermore, this will lead to termination of minimize() relatively soon after an error occurred on some worker - not only after some other exit criterion is met.

Closes #1512.

### DIFF
--- a/pypesto/optimize/ess/sacess.py
+++ b/pypesto/optimize/ess/sacess.py
@@ -7,6 +7,7 @@ import logging.handlers
 import multiprocessing
 import os
 import time
+from contextlib import suppress
 from dataclasses import dataclass
 from math import ceil, sqrt
 from multiprocessing import get_context
@@ -20,6 +21,7 @@ import numpy as np
 
 import pypesto
 
+from ... import MemoryHistory
 from ...startpoint import StartpointMethod
 from ...store.read_from_hdf5 import read_result
 from ...store.save_to_hdf5 import write_result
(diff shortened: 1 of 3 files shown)


pypesto/optimize/ess/sacess.py
```diff
@@ -853,15 +926,24 @@ def _run_worker(
...
+        h = logging.handlers.QueueHandler(log_process_queue)
+        worker._logger = logging.getLogger(
+            multiprocessing.current_process().name
+        )
+        worker._logger.addHandler(h)
 
-    return worker.run(problem=problem, startpoint_method=startpoint_method)
+        return worker.run(problem=problem, startpoint_method=startpoint_method)
+    except Exception as e:
+        with suppress(Exception):
```
PaulJonasJost, line 942: why do we need to suppress the exception here since we are already in an except? Or do you want to make sure it is definitely aborting?

PaulJonasJost (approved):
Looks good to me, thank you 🙏🏼

pypesto/optimize/ess/sacess.py
dweindl: It the logging fails for what ever reason, we don't care, but we need to ensure that the line below that block gets executed.

PR #1518, opened 2024-11-25
Fix `NLoptOptimizer.__repr__`

dweindl:
An incorrect value was used for `local_options`.


PaulJonasJost (approved):
Thanks for seeing this 👍🏼

Issue #1520, opened 2024-11-26
test_deepcopy_objective fails


dweindl:
https://github.com/ICB-DCM/pyPESTO/actions/runs/12026858059/job/33526668472:

```
___________________________ test_deepcopy_objective ____________________________

    def test_deepcopy_objective():
        """Test copying objectives (needed for MultiProcessEngine)."""
        petab_importer = pypesto.petab.PetabImporter.from_yaml(
            os.path.join(
                models.MODELS_DIR,
                "Boehm_JProteomeRes2014",
                "Boehm_JProteomeRes2014.yaml",
            )
        )
        factory = petab_importer.create_objective_creator()
        objective = factory.create_objective()
    
        objective.amici_solver.setSensitivityMethod(
            amici.SensitivityMethod_adjoint
        )
    
        objective2 = copy.deepcopy(objective)
    
        # test some properties
        assert (
            objective.amici_model.getParameterIds()
            == objective2.amici_model.getParameterIds()
        )
        assert (
            objective.amici_solver.getSensitivityOrder()
            == objective2.amici_solver.getSensitivityOrder()
        )
        assert (
            objective.amici_solver.getSensitivityMethod()
            == objective2.amici_solver.getSensitivityMethod()
        )
        assert len(objective.edatas) == len(objective2.edatas)
    
        assert objective.amici_model is not objective2.amici_model
        assert objective.amici_solver is not objective2.amici_solver
>       assert objective.steadystate_guesses is not objective2.steadystate_guesses
E       AttributeError: 'AmiciObjective' object has no attribute 'steadystate_guesses'. Did you mean: 'reset_steadystate_guesses'?

test/base/test_engine.py:114: AttributeError
```

PR #1523, opened 2024-11-27
CI: don't cache amici; pytest ignore amici_models; passenv BNGPATH

dweindl:
* Don't cache the amici installation. We are using amici/develop on purpose to detect problems early on. E.g., #1520 should have been detected much earlier.
* Ignore `amici_models` when collecting tests
* Use external `BNGPATH` if set

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,7 +157,7 @@ jobs:
 
     - name: Run tests
       timeout-minutes: 35
-      run: tox -e petab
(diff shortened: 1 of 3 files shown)


dweindl:
> What do we need a BNGPATH for? It is for bionetgen but I am not entirely aware here we use it.

Yes, for BioNetGen, which is used by pysb.

Also used here:
https://github.com/ICB-DCM/pyPESTO/blob/3b97e345b63b33297a0dbf7844f68328cc9a378b/test/petab/test_petab_import.py#L23-L29

PR #1528, opened 2024-11-27
More flexible `OptimizationResultHDF5Writer`

### LINKED ISSUE #1526: `OptimizationResultHDF5Writer` should accept `OptimizeResult`
`OptimizationResultHDF5Writer` should accept `OptimizeResult`, because that is what it is supposed to store. Currently, it requires a `Result`.

### LINKED ISSUE #1527: `OptimizationResultHDF5Writer` should allow writing individual `OptimzerResults`
Currently it's not possible to save several `OptimizerResults` incrementally using `OptimizationResultHDF5Writer`. The full set has to be written at once. (`overwrite=False` would result in an error, `overwrite=True` would delete the previously stored ones).
Ideally, `OptimizationResultHDF5Writer` would allow storing individual `OptimizerResults`.

dweindl:
Previously, only writing `Result` was supported.

Changes:
*  `OptimizeResult`, `OptimizerResult`(s) can be written by `OptimizationResultHDF5Writer.write()` (Closes #1526)
*  `OptimizerResult`s can be written incrementally by `OptimizationResultHDF5Writer.write_optimizer_result()` (Closes #1527)
* Fixed an `AttributeError` in `pypesto.store.save_to_hdf5.check_overwrite` with `h5py.Group`s

### DIFF
--- a/pypesto/store/save_to_hdf5.py
+++ b/pypesto/store/save_to_hdf5.py
@@ -1,23 +1,22 @@
 """Include functions for saving various results to hdf5."""
+from __future__ import annotations
 
 import logging
 import os
 from numbers import Integral
 from pathlib import Path
-from typing import Union
 
 import h5py
 import numpy as np
 
+from .. import OptimizeResult, OptimizerResult
 from ..result import ProfilerResult, Result, SampleResult
 from .hdf5 import write_array, write_float_array
 
 logger = logging.getLogger(__name__)
 
 
-def check_overwrite(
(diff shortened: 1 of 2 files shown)


PaulJonasJost (approved):
Thanks, that indeed makes it more flexible. Should we introduce these changes to profiler and sampler writer as well?

dweindl:
> Should we introduce these changes to profiler and sampler writer as well?

If there is demand. I wouldn't do it just for the sake of it.

PR #1529, opened 2024-11-27
SacessOptimizer: More efficient saving of intermediate results

dweindl:
Don't write old local optima again and again in every iteration to save time and to avoid huge fragmented files.

Some other smaller modifications/fixes:
* make sacess_history plotting more robust. 
* Set SacessOptimizer.exit_flag

### DIFF
--- a/pypesto/optimize/ess/sacess.py
+++ b/pypesto/optimize/ess/sacess.py
@@ -24,7 +24,10 @@ import pypesto
 from ... import MemoryHistory
 from ...startpoint import StartpointMethod
 from ...store.read_from_hdf5 import read_result
-from ...store.save_to_hdf5 import write_result
+from ...store.save_to_hdf5 import (
+    OptimizationResultHDF5Writer,
+    ProblemHDF5Writer,
+)
 from ..optimize import Problem
 from .ess import ESSExitFlag, ESSOptimizer
 from .function_evaluator import create_function_evaluator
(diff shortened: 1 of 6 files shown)


dweindl:
Hi, could somebody please review?

vwiela (approved):

pypesto/optimize/ess/sacess.py
```diff
@@ -683,20 +692,14 @@ def run(
             # perform one ESS iteration
             ess._do_iteration()
 
-            if self._tmp_result_file:
```
stephanmg, line 686: Another hyper parameter like `trace_save_iter` for ESS?

stephanmg (commented):
Remaining looks good. Thanks!

pypesto/optimize/ess/sacess.py
dweindl: Not really. This class is not intended to be used directly, only through `SacessOptimizer`. In that case, this will always evaluate to True. It's rather for unit testing purposes where one doesn't want/need any intermediary result files.

dweindl: Or was that a request to make saving intermediate result optional in `SacessOptimizer`?

stephanmg: Yes @dweindl sorry, shaky mouse. Don't see it be super urgent to add this, but agreed, add separately.

PR #1530, opened 2024-11-29
Select: use the new "standardized" `petab_select` classes

dilpath:
- [x] fix notebook: https://pypesto--1530.org.readthedocs.build/en/1530/example/model_selection.html
- [x] use PEtab Select vis in notebook

Will not merge until the next PEtab Select release
- [x] update petab-select version in setup.cfg


doc/example/model_selection.ipynb
```diff
@@ -375,6 +349,7 @@
    "outputs": [],
    "source": [
     "petab_select_problem.model_space.reset_exclusions()\n",
+    "petab_select_problem.state.reset()\n",
```
Doresic, line 352: Could you comment shortly why you're resetting and what exactly?
I can guess what it is, but just for a new user to not be confused or thinks that it's necessary to do always.

doc/example/model_selection.ipynb
```diff
@@ -331,18 +311,12 @@
...
-    "        ),\n",
-    "        **initial_model_label,\n",
-    "    },\n",
-    ");"
+    "plot_data = plot.PlotData(\n",
+    "    models=petab_select_problem.state.models,\n",
+    "    criterion=Criterion.AIC,\n",
+    ")\n",
+    "plot_data.augment_labels(criterion=True)\n",
+    "plot.graph_history(plot_data=plot_data);"
```
Doresic, line 319: I like this a lot, feels easier to pass arguments to plotting

doc/example/model_selection.ipynb
```diff
@@ -425,6 +420,7 @@
    "source": [
     "# Repeat with AICc and criterion_threshold == 10\n",
     "petab_select_problem.model_space.reset_exclusions()\n",
+    "petab_select_problem.state.reset()\n",
```
Doresic, line 423: same as above

pypesto/select/postprocessors.py
```diff
@@ -109,10 +110,15 @@ def model_id_binary_postprocessor(problem: ModelProblem):
...
-    problem.model.model_id = model_id
+    warnings.warn(
+        (
+            "This is obsolete. Model IDs are by default the model hash, which "
+            "is now similar to the binary string."
+        ),
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    problem.model.model_id = str(problem.model.hash)
```
Doresic, line 121: This is very nice.

pypesto/select/problem.py
```diff
@@ -21,11 +20,10 @@ class Problem:
     Attributes
     ----------
     calibrated_models:
-        Storage for all calibrated models. A dictionary, where keys are
-        model hashes, and values are :class:`petab_select.Model` objects.
+        All calibrated models.
```
Doresic, line 23: would it make sense to link the `Models` class here?

pypesto/select/problem.py
```diff
@@ -142,78 +140,60 @@ def select(
 
         Returns
         -------
-        A 3-tuple, with the following values:
+        A 3-tuple, with the following values from this iteration:
```
Doresic, line 143: is it a 2-tuple then? If only best model and all models are returned?

pypesto/visualize/select.py
Doresic, line 1: This is nice, it's nice to have it all in petab_select rather then in pypesto as well.

Doresic (approved):
Like the changes a lot. Much nicer to use

stephanmg (approved):
The changes wrt visualization look fine to me. Thanks for the contribution.

doc/example/model_selection.ipynb
dilpath: Good point. It's not a great interface currently because there is an ongoing refactor to move all "state" information to `problem.state`. The state information is currently distributed among several PEtab Select classes, and is unlikely to be resolved before publication.
So to make things simpler for the user here, I changed this code to simply reload the PEtab Select problem, instead of resetting it.

doc/example/model_selection.ipynb
dilpath: I'm glad, another benefit is that it simplified the plotting code in the library quite a bit too :grinning:

pypesto/select/problem.py
dilpath: Yes, done now, thanks

PR #1533, opened 2024-12-02
Add Ipopt factory for use with SacessOptimizer

dweindl:
... to forward the overall wall time limit to Ipopt.

**Note: the Ipopt walltime limit is not supported by the current (outdated) Ubuntu Ipopt packages** 

This should work with the ubuntu 25.04 ipopt version.

### DIFF
--- a/pypesto/optimize/ess/sacess.py
+++ b/pypesto/optimize/ess/sacess.py
@@ -1295,6 +1295,51 @@ class SacessCmaFactory:
         return f"{self.__class__.__name__}(options={self._options})"
 
 
+class SacessIpoptFactory:
+    """Factory for :class:`IpoptOptimizer` instances for use with :class:`SacessOptimizer`.
+
+    :meth:`__call__` will forward the walltime limit and function evaluation
+    limit imposed on :class:`SacessOptimizer` to :class:`IpoptOptimizer`.
+    Besides that, default options are used.
+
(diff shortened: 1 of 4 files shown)


pypesto/optimize/ess/sacess.py
```diff
@@ -1248,6 +1248,52 @@ def __repr__(self):
...
+        self, max_walltime_s: int, max_eval: int
+    ) -> pypesto.optimize.IpoptOptimizer:
+        """Create a :class:`IpoptOptimizer` instance."""
+
+        options = self._ipopt_options.copy()
+        if np.isfinite(max_eval):
+            options["max_wall_time"] = max_walltime_s
+
+        # only accepts int
+        if np.isfinite(max_eval):
```
PaulJonasJost, line 1285:
```suggestion
        if np.isfinite(max_walltime):
            options["max_wall_time"] = max_walltime_s

        # only accepts int
        if np.isfinite(max_eval):
```

PaulJonasJost (changes requested):
Thanks, needs a small adjustment. Requesting changes to prevent pushing into queue.

pypesto/optimize/ess/sacess.py
dweindl: Oops, thanks.

PR #1534, opened 2024-12-03
Relative: Fix calculate directly not scaling inner pars when applying to rdatas

Doresic:
At the end of the `calculate_directly` method of the `RelativeAmiciCalculator` the inner parameters that are optimized need to be applied to the rdatas (compared to the re-simulation of `call_amici_twice` with non-dummy inner parameter values).

However, the inner parameters were not scaled accordingly in this step. E.g. if the scale of a scaling inner parameter (value = 10^-2) was set to LOG10 (log10 value -2), it would stay in LOG10 scale and be applied like that (multiplying its rdata by -2) instead of being scaled to LIN scale and then being applied to the rdatas.


dweindl (approved):
:+1:

PR #1538, opened 2024-12-11
Update publish_dockerhub.yml

stephanmg:
Enable on develop branch instead of main.


dweindl:
I think `ICB_DCM/pypesto` needs to be changed to `stephanmg/pypesto` in the workflow.

stephanmg:
> I think `ICB_DCM/pypesto` needs to be changed to `stephanmg/pypesto` in the workflow.

True. Changed.

stephanmg:
@vwiela or @PaulJonasJost ?

vwiela (approved):

Issue #1535, opened 2024-12-04
visualize.parameters produces error when working with hdf5-stored result.problem.x_scales


C-Peiter:
Seems like, when I load a result using `result=pypesto.store.read_result("path/to/result.hdf5")` the `result.problem.x_scales` attribute gets loaded as an array of byte-strings. This produces a type error at `visualize/parameters.py:128` when all scales are the same. The error is `TypeError: can only concatenate str (not "bytes") to str` and otherwise prints the scale as, for example, `b'log10'`.

The result file was saved with
```python
pypesto.store.write_result(
    result=result,
    filename="path/to/result.hdf5",
    problem=True,
    optimize=True,
    profile=True,
    sample=True,
)
```

OS: Ubuntu 22.04
HDF5 lib on ubuntu: libhdf5-103-1 1.10.7+repack-4ubuntu2
Python version: 3.10.12
pyPESTO version: 0.5.4
h5py version: 3.12.1

stephanmg:
@C-Peiter Could you provide the snippet which produces the error with the .hdf5 file? Then I'll try to reproduce this on my Ubuntu and have a look.

C-Peiter:
I attached a result stored in hdf5 and the code to load the result and plot the parameters. I also have a whole PEtab problem for that, if you want.
[mwe.zip](https://github.com/user-attachments/files/18095304/mwe.zip)

stephanmg:
Think a simple `decode('utf-8')` should be enough, realized in the commit of PR #1542

PR #1542, opened 2024-12-18
Update parameters.py

stephanmg:
Fix issue #1535

### DIFF
--- a/pypesto/visualize/parameters.py
+++ b/pypesto/visualize/parameters.py
@@ -464,7 +464,7 @@ def handle_inputs(
 
     # If all the scales are the same, put it in the x_axis_label
     if len({x_scale for _, x_scale in x_labels}) == 1:
-        x_axis_label = "Parameter value (" + x_labels[0][1] + ")"
+        x_axis_label = "Parameter value (" + x_labels[0][1].decode('utf-8') + ")"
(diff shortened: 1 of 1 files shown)


m-philipps:
Thanks. Is there a reason for using `bytes` instead of `str` in `pypesto.store.read_result`? If not, I would be in favour of changing that.

stephanmg:
Problem class defines this:
`  x_scales: Optional[Iterable[str]] = None,`

So the issue can't be in the read-in, as one can't store a byte string in this.

stephanmg:
`pypesto.store.read_result` reads into the`pypesto.result`, `problem` has the attribute `x_scales` with type hint defined as `Optional[Iterable[str]]` still when I print directly `result.problem.x_scales` I get a byte str as reported by Clemens too.

the issue originates from `pypesto.store.read_from_hdf5.py` - ProblemHDF5Reader class.

data is read from the h5 file, which provides byte strings, but is never decoded (x_names is done correctly and decoded)

stephanmg:
Now should be fixed at the point or origin. Please have a look.

PaulJonasJost (approved):
Thanks for noticing 👍🏼

stephanmg:
> Thanks for noticing 👍🏼

All credits to @C-Peiter for reporting the issue ;-)

PR #1549, opened 2025-01-31
make print parameter summary nicer

arrjon:
This pull request includes a small update to the `print_parameter_summary` method in the `pypesto/problem/base.py` file to add more detailed information about the parameters. I added the fields `scale` and `fixedVal (scaled)`. This can be used to verify that the problem is loaded as expected. Any other wishes what to include here?


arrjon:
Now bounds are only shwon for free parameters, good point Paul!

Issue #1553, opened 2025-03-04
Common interface for common optimizer options


dweindl:
**Feature description**
It would be convenient if one could set common optimizer options through a common interface. For example, a time limit.

**Motivation/Application**
Having `pypesto.optimize.optimizer.Optimizer.set_maxtime()` (and `.supports_maxtime`) would be easier than looking up the specific option for each optimizer. Currently, for example, the SacessOptimizer, needs to handle every optimizer individually to pass down the time limit. This would also be helpful any multi-start optimization with a fixed time budget.

PR #1555, opened 2025-03-04
Make CmaOptimizer respect wall time limits and max evals under SacessOptimizer

dweindl:
Adds `SacessCmaFactory` so that the wall time and function evaluation budget of `SacessOptimizer`/`ESSOptimizer` can be passed on to `CmaOptimizer`.


stephanmg:
Looks good to me.

stephanmg (approved):

PaulJonasJost (approved):
Looks good. Do there a way to unify the definition of the factories? Not sure about this, just a thought.

dweindl:
> Looks good. Do there a way to unify the definition of the factories? Not sure about this, just a thought.

With #1553, yes.

PR #1556, opened 2025-03-05
Faster reading of HDF5 optimization results

dweindl:
Some changes to read HDF5 optimization results more efficiently.

For some test file provided by @m-philipps with 41,387 OptimizerResults, this reduces loading time by 95% (682s -> 33s).


PaulJonasJost (approved):
Thanks 🙏🏼 Do i see correctly that the main problem were the recursive calls?

dweindl:
> Thanks 🙏🏼 Do i see correctly that the main problem were the recursive calls?

Not recursion per se, but the associated repeated checking of all IDs for each single OptimizerResult was one of the main issues.

m-philipps:
Thank you!

PR #1558, opened 2025-03-07
Fix SacessOptimizer relative fval improvement check

dweindl:
Fix calculation of the relative objective improvement. Not expecting a substantial impact in practice.

Avoids division potential division-by-0 warnings.

Also, add a test for the SacessOptimizer adaptation step.

### DIFF
--- a/test/optimize/test_optimize.py
+++ b/test/optimize/test_optimize.py
@@ -574,6 +574,33 @@ def test_ess_multiprocess(problem, request):
     print("ESS result: ", res.summary())
 
 
+def test_sacess_adaptation(capsys):
+    """Test that adaptation step of the SACESS optimizer succeeds."""
+    obj = rosen_for_sensi(max_sensi_order=2, integrated=False)["obj"]
+    lb = 0 * np.ones((1, 10))
+    ub = 1 * np.ones((1, 10))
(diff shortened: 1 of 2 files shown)


test/optimize/test_optimize.py
```diff
@@ -574,6 +574,33 @@ def test_ess_multiprocess(problem, request):
...
+        options=SacessOptions(
+            # trigger frequent adaptation
+            # - don't do that in production
+            adaptation_min_evals=0,
+            adaptation_sent_offset=0,
+            adaptation_sent_coeff=0,
+        ),
+    )
+    ess.minimize(problem)
+    assert "Updated settings on worker" in capsys.readouterr().err
```
PaulJonasJost, line 601: Just for me to understand the code:

We set the options such that It adapts every evaluation? If you say don't actually do this, wouldn't it make more sense to check whether settings similar to this were used and throw a warning?

PaulJonasJost (approved):

test/optimize/test_optimize.py
dweindl: Not during every iteration, but very often.
It's not that clear what good values are. If I only issue a warning at `0`, this suggests `1` is good, so I'd rather avoid it altogether.

PR #1559, opened 2025-03-10
Remove obsolete aesara requirement

dweindl:
Aesara is not used anywhere in pypesto.


dweindl:
Removing aesara requires sorting out a number of third-party dependency issues... :hankey:

PR #1561, opened 2025-03-13
Add examples to pypesto module

PaulJonasJost:
I added the option to use examples directly in the Pypesto module.

Some thoughts for open discussion:
- Currently, we load them from GitHub, thus needing an internet connection. Another option would be to include the files, though this might add unnecessary files to the package
- Some examples describe a pipeline; for those, I have not added the usage of examples.
- @dilpath there is the "testing" module that only contains two hierarchical examples. According to my gitblame these were created by you. Would it be ok to move this to the examples module as its own examples?

The examples for now have a petab_problem, an objective and a problem along with a small description. Happy for feedback, whether I missed something.


dilpath (approved):

dilpath:
>     * Currently, we load them from GitHub, thus needing an internet connection. Another option would be to include the files, though this might add unnecessary files to the package

Another option (unless this is what you meant) is it conditionally add it to the pacakge, like `pip install pypesto[examples]`

>     * @dilpath there is the "testing" module that only contains two hierarchical examples. According to my gitblame these were created by you. Would it be ok to move this to the examples module as its own examples?

Fine for me :+1:

PR #1564, opened 2025-03-31
SacessOptimizer: Minor updates

dweindl:
* debug logs include time for saving intermediary results
* split file creation and tmp file deletion
* add number of local searches to worker results


PaulJonasJost (approved):
Looks good to me 👍🏼

PR #1566, opened 2025-04-14
fix cutoff based on chi2 distribution

plakrisenko:
Division by 2 was missing in the calculation of the cutoff, based on the [Wilk's theorem](https://en.wikipedia.org/wiki/Wilks%27_theorem) which states that $-2\log(\text{likelihood ratio})$ asymptotically approaches the chi-squared distribution. The ensemble is created based on 
$-2\log(\mathcal{L}(\theta)/\mathcal{L}(\hat{\theta})) = -2\log(\mathcal{L}(\theta)) - (-2\log(\mathcal{L}(\hat{\theta}))) = 2(J(\theta) - J(\hat{\theta}))) \leq \Delta_{\alpha}$, where $\Delta_{\alpha}$ is a percentile of the $\chi^2$ distribution and $J(\hat{\theta})$ is the smallest objective function value found during optimization. The ensemble contains all the parameter vectors $\theta$ that satisfy $J(\theta)\leq J(\hat{\theta}) + \Delta_{\alpha}/2$.
 
Co-authored-by: Dilan Pathirana <59329744+dilpath@users.noreply.github.com>


dilpath (approved):
:+1: 

A test to ensure that we compute the expected confidence intervals for a simple problem would be great. No need for this PR, but then we should open an issue for that.

dweindl:
> Division by 2 was missing in the calculation of the cutoff, based on the [Wilk's theorem](https://en.wikipedia.org/wiki/Wilks%27_theorem) which states that − 2 log ⁡ ( likelihood ratio ) asymptotically approaches the chi-squared distribution. The ensemble is created based on − 2 log ⁡ ( L ( θ ) / L ( θ ^ ) ) = − 2 log ⁡ ( L ( θ ) ) − ( − 2 log ⁡ ( L ( θ ^ ) ) ) = 2 ( J ( θ ) − J ( θ ^ ) ) ) ≤ Δ α , where Δ α is a percentile of the χ 2 distribution and J ( θ ^ ) is the smallest objective function value found during optimization. The ensemble contains all the parameter vectors θ that satisfy J ( θ ) ≤ J ( θ ^ ) + Δ α / 2 .

Thanks. It would be great if you could add this description to that function, since apparently this was non-obvious.

PaulJonasJost (approved):
Thanks for that 🙏🏼

pypesto/ensemble/ensemble.py
```diff
@@ -1222,13 +1222,22 @@ def calculate_cutoff(
...
     parameter is chi^2 distributed with n_theta degrees of freedom.
 
+    The ensemble is created based on
+    :math: `-2\log(\mathcal{L}(\theta)/\mathcal{L}(\hat{\theta})) =
+    -2\log(\mathcal{L}(\theta)) - (-2\log(\mathcal{L}(\hat{\theta}))) =
+    2(J(\theta) - J(\hat{\theta}))) \leq \Delta_{\alpha}`, where :math: `mathcal{L}` is the likelihood,
+    :math: `J` is the negative log-likelihood,:math: `\Delta_{\alpha}` is a percentile of the
+    :math: `\chi^2` distribution and :math: `J(\hat{\theta})` is the smallest objective function value
+    found during optimization. The ensemble contains all the parameter vectors :math: `\theta` that satisfy
+    :math: `J(\theta)\leq J(\hat{\theta}) + \Delta_{\alpha}/2`.
```
dilpath, line 1239: This doesn't appear to render correctly
https://pypesto--1566.org.readthedocs.build/en/1566/api/pypesto.ensemble.html#pypesto.ensemble.calculate_cutoff

There is an example here, might just need some newlines
https://amici.readthedocs.io/en/latest/_modules/amici/import_utils.html#noise_distribution_to_cost_function

dilpath (approved):
Thanks! Looks great

PR #1567, opened 2025-04-15
Update getting_started.ipynb

Doresic:
The comments on the profiling results in the getting started notebook were outdated. They referred to the results of the old version with the Boehm model rather than the new version with the conversion reaction model.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,8 +20,8 @@ project_urls =
 # Author information
 author = The pyPESTO developers
 author_email = yannik.schaelte@gmail.com
-maintainer = Paul Jonas Jost, Maren Philipps, Domagoj Dorešić, Fabian Fröhlich
-maintainer_email = paul.jost@uni-bonn.de, maren.philipps@uni-bonn.de, domagoj.doresic@uni-bonn.de, fabian.frohlich@crick.ac.uk
+maintainer = Paul Jonas Jost, Domagoj Dorešić, Vincent Wieland, Fabian Fröhlich
+maintainer_email = paul.jost@uni-bonn.de, domagoj.doresic@uni-bonn.de, vwieland@uni-bonn.de, fabian.frohlich@crick.ac.uk
 
 # License information
 license = BSD-3-Clause


doc/example/getting_started.ipynb
```diff
@@ -639,9 +647,7 @@
     }
    },
    "source": [
-    "The plot shows that seven parameters are identifiable, since the likelihood is tightly centered around the optimal parameter. Two parameters (`k_exp_hetero` and `k_imp_homo`) cannot be constrained by the data.\n",
-    "\n",
-    "Furthermore pyPESTO allows to visualize confidence intervals directly via"
+    "Furthermore, pyPESTO allows to visualize approximate confidence intervals directly via `profile_cis`. Confidence intervals are computed by finding parameter values for which the log posterior ratio is above the approximate threshold assuming a $\\chi^2$ distribution of the likelihood test statistic. The plot shows that both model parameters are identifiable, since the confidence regions are finite."
```
vwiela, line 650: Isn't it the likelihood ratio test statistic?

dweindl, line 650: > The plot shows that both model parameters are identifiable, since the confidence regions are finite.

For parameters with finite bounds, the computed confidence region would always be finite, independent of identifiability, wouldn't it?

dweindl: > I'm unsure whether that's important to add here or even how I would state it. 

I'd find it important to clarify it, to not encourage users to do their "identifiability analysis" via `is_identifiable = ~np.any(np.isinf(compute_ci(...)))` which would claim identifiability for every constrained parameter.

Maybe something like that the profile crosses the threshold on both sides of the MLE, and therefore, the parameter is considered identifiable? (ignoring potential multiple optima)

> But the statement that it's identifiable because the confidence intervals are finite is definitely true, no?

I'd say it depends on how the confidence intervals are computed. If we just stop at the bounds, then no, not necessarily. If we extend the bounds as you pointed out, then yes.

PaulJonasJost (approved):
Thanks. Why remove the profile index, though?

PR #1570, opened 2025-05-05
Speed up `read_hdf5_optimization`

dweindl:
Avoid redundant HDF5 file reading. Reduces the time for loading optimization results from HDF5 by roughly 50% for a test file with 71560 OptimizerResults.


PaulJonasJost (approved):
Looks good, thanks 👍🏼

vwiela (approved):
Did improve loading times for me as well, thanks for the improvement.

Issue #1568, opened 2025-04-24
test/base/test_objective.py::test_jax fails


dweindl:
E.g., https://github.com/ICB-DCM/pyPESTO/actions/runs/14634447500/job/41062583013:

```
________________________ test_jax[0-False-False-False] _________________________
jax.errors.SimplifiedTraceback: For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.

The above exception was the direct cause of the following exception:

max_sensi_order = 0, integrated = False, enable_x64 = False
fix_parameters = False

    @pytest.mark.parametrize("enable_x64", [True, False])
    @pytest.mark.parametrize("fix_parameters", [True, False])
    def test_jax(max_sensi_order, integrated, enable_x64, fix_parameters):
        """Test function composition and gradient computation via jax"""
        import jax
        import jax.numpy as jnp
    
        if max_sensi_order == 2:
        unbatched_args, batched_args = util.partition_list(is_batched, new_args)
        def _batch_fun(batched_args):
          merged_args = util.merge_lists(is_batched, unbatched_args, batched_args)
          return prim.bind(
              *merged_args,
              result_avals=result_avals,
              vmap_method=vmap_method,
              **kwargs,
          )
        unroll = vmap_method == "sequential_unrolled"
        g = lambda _, x: ((), _batch_fun(x))
        _, outvals = jax.lax.scan(g, (), batched_args, unroll=unroll)
      else:
>       raise NotImplementedError(
            f"vmap is only supported for the {prim.name} primitive when vmap_method "
            "is one of 'sequential', 'sequential_unrolled', 'expand_dims', "
            f"'broadcast_all', or 'legacy_vectorized'. Got {vmap_method=}.")
E       NotImplementedError: vmap is only supported for the pure_callback primitive when vmap_method is one of 'sequential', 'sequential_unrolled', 'expand_dims', 'broadcast_all', or 'legacy_vectorized'. Got vmap_method=None.

.tox/base/lib/python3.10/site-packages/jax/_src/ffi.py:724: NotImplementedError
```

CC @FFroehlich

dweindl:
I think this is the relevant DeprecationWarning with jax<0.6.0:

[...]

PR #1572, opened 2025-05-06
CI: require jax < 0.6.0

dweindl:
Workaround for https://github.com/ICB-DCM/pyPESTO/issues/1568.


vwiela (approved):
I think good as a workaround for the time being, but we should keep it on the ToDo list.

PR #1574, opened 2025-05-06
Parameter ensemble plot

plakrisenko:
Visualization of parameter values included in the ensemble.


pypesto/visualize/ensemble.py
```diff
@@ -413,3 +414,59 @@ def _create_patches(
...
+    Returns
+    -------
+    ax: matplotlib.Axes
+        The plot axes.
+    """
+
+    if ax is None:
+        fig, ax = plt.subplots(figsize=size)
+
+    x = -0.4
```
stephanmg, line 446: What is x?

pypesto/visualize/ensemble.py
```diff
@@ -413,3 +414,59 @@ def _create_patches(
...
+        rectangles.append(
+            Rectangle((x, np.min(par_values)), w, h))
+        x += w + 0.2
+    ax.add_collection(PatchCollection(rectangles, facecolors=[1, 1, 1, 1], edgecolors='dimgrey'))
+
+    for i, v in enumerate(ensemble.x_vectors):
+        ax.scatter(x=[i]*ensemble.n_vectors, y=v, s=40, color=colors, alpha=0.6)
+    # plot the best parameter values
+    ax.scatter(np.arange(ensemble.n_x), ensemble.x_vectors[:, 0], s=40,
+               color=[1.        , 0.        , 0.        , 1.        ])
```
stephanmg, line 464: Same here.

stephanmg (approved):
See single line comments.

pypesto/visualize/ensemble.py
plakrisenko: the x coordinate for the rectangle vertex, changed now

test/visualize/test_visualize.py
```diff
@@ -636,6 +636,22 @@ def test_ensemble_identifiability():
     # test plotting from a collection object
     visualize.ensemble_identifiability(my_ensemble)
 
+@close_fig
+def test_ensemble_parameters_plot():
+    # creates a test problem
+    problem = create_problem(n_parameters=100)
```
PaulJonasJost, line 642: such a big one necessary for the test? 🙈

PaulJonasJost (approved):
Thanks for this :) looks good to me

pypesto/visualize/ensemble.py
```diff
@@ -413,3 +414,70 @@ def _create_patches(
...
+    parameter_ids: Optional[list[int]] = None,
+    size: Optional[tuple[float]] = (6, 12)
+):
+    """
+    Visualize parameter ensemble.
+
+    Parameters
+    ----------
+    ensemble:
+        ensemble of parameter vectors (from pypesto.ensemble).
```
Doresic: So this is a method to generally just visualize the ensemble, so we cannot really say that the rectangles represent, for example, 95% confidence intervals. But one would need to create an ensemble in some way, and then subset it using a xi^2 threshold to get confidence intervals, right?
Would it be good to have this showcased somewhere? How to use this method correctly? Or is it ok to just expand the description of the visualization in the docstring to describe that it will just visualize all members of the ensemble.

Doresic (approved):
Looks really nice! Will use myself :) I have a couple of comments
- can we have a general legend on all the elements of the figure (the rectangles, red point, dashed line)
- can we have a bar on the side or at the top of the plot that shows the grey and mentions what the grey gradient actually represents? Btw, what does it actually represent? The obj. func. value of the point?

PR #1577, opened 2025-05-16
Fix import in doc/example/synthetic_data.ipynb

dweindl:
The previous import was long deprecated and the respective module has been moved in amici 0.32.0.

Also build the documentation on a more recent Ubuntu with a more recent version of SWIG.


dilpath (approved):
Thanks!

PR #1589, opened 2025-07-28
remove unnecessary title and axes labels

plakrisenko:
I think duplication of the title and axes labels is not needed in the zoom in subfigure. What do you think?

Before:
waterfall1

After:
waterfall2

### DIFF
--- a/doc/example/getting_started.ipynb
+++ b/doc/example/getting_started.ipynb
@@ -594,12 +594,19 @@
     "%%time\n",
     "%%capture\n",
     "\n",
+    "import scipy as sp\n",
+    "\n",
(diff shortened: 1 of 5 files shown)


dweindl:
> I think duplication of the title and axes labels is not needed in the zoom in subfigure. What do you think?

Could you add a before/after figure in a comment?

stephanmg (approved):
Looks fine to me.

stephanmg (approved):
Potentially could also do (not tested):
`inset_axes.set(title=None, xlabel=None, ylabel=None)`

But does not matter. Approved.

PR #1598, opened 2025-08-25
split ProfilerTask into descending and ascending parts

plakrisenko:


Doresic (approved):
Looks good to me.
Nicely written, easy to understand.
Putting some comments in some places might make it easier to understand what's being done, like
- pair the descending and ascending profile halves for each parameter using paired_profiles
- combine the two profile halves before assigning them to the profile result list

Just a suggestion, good as is.

PR #1620, opened 2025-10-13
Test with Python 3.14

dweindl:
Test with Python 3.14, except for Julia tests.

Julia tests `test_petabJL_from_module` and `test_pyjulia_pipeline` still fail with 3.14.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -88,7 +88,7 @@ deps =
 #    python3 -m pip install git+https://github.com/AMICI-dev/amici.git@main\#egg=amici&subdirectory=python/sdist
 commands_pre =
    python3 -m pip uninstall -y amici
-   python3 -m pip install git+https://github.com/AMICI-dev/amici.git@v0.34.1\#egg=amici&subdirectory=python/sdist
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -88,7 +88,7 @@ deps =
 #    python3 -m pip install git+https://github.com/AMICI-dev/amici.git@main\#egg=amici&subdirectory=python/sdist
 commands_pre =
    python3 -m pip uninstall -y amici
-   python3 -m pip install git+https://github.com/AMICI-dev/amici.git@v0.34.1\#egg=amici&subdirectory=python/sdist
+   python3 -m pip install git+https://github.com/AMICI-dev/amici.git@v0.34.2\#egg=amici&subdirectory=python/sdist
```
PaulJonasJost, line 91: Should we add a TODO here to fix that when we merged amici 1.0.0?

PaulJonasJost (approved):
Thanks. Don't we also need to change to 3.12 then?

tox.ini
dweindl: Not necessary. Tests will just fail if that isn't removed.

PaulJonasJost: True

PR #1622, opened 2025-10-15
Prior sample

arrjon:
This pull request introduces comprehensive support for sampling parameter values from prior distributions throughout the `pyPESTO` codebase. The main changes include integrating this as startpoint methods in optimization and sampling.

In a Bayesian workflow, one would like to do multistart with initial points sampled from the prior. Furthermore, starting sampling in the optimal point might be harmful as chains can get stuck in the minimum and not explore the parameter space. Combining random values from the prior and optimal points can mitigate this problem.

**Prior sampling support:**

* Added a `sample` method to the `NegLogPriors` class, allowing users to draw samples from defined prior distributions. 
* Extended the `get_parameter_prior_dict` and related functions to include `type` and `parameters` fields, ensuring all necessary information for sampling is present in the prior dictionaries. 

**Startpoint initialization integration:**

* Added the `PriorStartpoints` method and updated the startpoint interface to accept priors, enabling startpoint sampling from prior distributions when available. When you load a petab problem, then the petab starting points are still used following the petab standard (so uniform on scale of the parameters).
 
**Sampling and parallel tempering workflow updates:**

* Modified the main `sample` function and parallel tempering sampler to support warm starting from previous optimization results or sampling from priors, controlled by a new `warm_start` parameter and a `same_start_point` option for parallel tempering. 

**Logging and usability improvements:**

* Added informative logging when aggregated objectives are created with priors and when different initialization strategies are used in sampling. Before, people might have thought they are doing maximum likelihood because they did not carefully check their petab problem.


pypesto/petab/importer.py
```diff
@@ -408,6 +408,7 @@ def create_problem(
                     "is not yet supported."
                 )
             objective = AggregatedObjective([objective, prior])
+            logger.info("Created an aggregated objective using priors.")
```
vwiela: This is always creating a AggregatedObjective if priors are specified in the petab problem, right?

Should we either make it optional?
Or better explain in the docs how to subset the objective to only the likelihood, such that users who want just likelihood optimization, can do that easily without wrapping their head around the code too much.

I mean I can think of use cases where one just want to optimize the likelihood although the PEtab problem contains priors and then it should be easily possible to do so.

pypesto/objective/priors.py
```diff
@@ -225,6 +228,50 @@ def residual_jacobian(self, x):
...
+            Array of shape (n_samples, n_parameters) containing samples from the prior.
+        """
+        rng = np.random.default_rng(seed)
+
+        # Determine the number of parameters
+        if self.x_names is not None:
+            n_parameters = len(self.x_names)
+        else:
+            # Find the maximum index in prior_list
+            n_parameters = max(prior["index"] for prior in self.prior_list) + 1
```
vwiela: For `self.x_names = None` and only one parameter having a prior specified (say the parameter with index 10), this line leads then to priors for the first 11 parameters.
And the first 10, since they have no specified, will be filled with zeros.

vwiela (commented):
Thanks for the nice enhancement. 

I left some comments to discuss.

Personally, I like your warm-start idea. And would be good to also have it for the startpoints of the parallel tempering chains, I guess.

Maybe, we should also include this in the example notebook then.

pypesto/petab/importer.py
arrjon: This was always the case, I only added the logging. If priors are given, we load the posterior. The standard way for maximum likelihood estimation is specifying a corresponding petab problem without priors. If we want an option for users to turn off priors even if they are specified in the petab problem, we can add this is another pull request.

pypesto/objective/priors.py
arrjon: I changed the line of code so now `nan` are returned, just to make sure that the zeros are not used anywhere. The test is still working, so please let me know where exactly it fails for you (so I can also improve my testing).

arrjon:
> Personally, I like your warm-start idea. And would be good to also have it for the startpoints of the parallel tempering chains, I guess.

I added this now!

PaulJonasJost (approved):
Thanks, looks good to me. Did not specifically check the notebooks though

doc/example/model_evidence_and_bayes_factors.ipynb
dilpath: delete commented out parts

pypesto/objective/priors.py
```diff
@@ -225,6 +228,51 @@ def residual_jacobian(self, x):
...
+                raise ValueError(
+                    "Prior type and parameters must be specified for sampling. "
+                    "Use 'get_parameter_prior_dict' to create proper prior dictionaries."
+                )
+
+            prior_samples_dict[index] = _sample_from_prior(
+                prior_type, prior_parameters, n_samples, rng
+            )
+
+        return prior_samples_dict
```
dilpath: Is this the most useful format to return the prior samples in? Elsewhere in pyPESTO, samples are arrays of e.g. n_parameters x n_samples (or transpose...). The dictionary means that users will need to construct the vectors themselves, right?

pypesto/objective/priors.py
```diff
@@ -560,3 +614,62 @@ def function(x):
...
+    """
+    if prior_type in [C.UNIFORM, C.PARAMETER_SCALE_UNIFORM]:
+        low = prior_parameters[0]
+        high = prior_parameters[1]
+        return rng.uniform(low, high, n_samples)
+
+    elif prior_type in [C.NORMAL, C.PARAMETER_SCALE_NORMAL]:
+        mean = prior_parameters[0]
+        std = prior_parameters[1]
+        return rng.normal(mean, std, n_samples)
```
dilpath: Where is this transformed into a sample from the log-normal distribution, if parameter scale is log?

pypesto/sample/sample.py
```diff
@@ -67,14 +75,41 @@ def sample(
...
+                get_start_params = PriorStartpoints(check_fval=True)
+                x0_prior = get_start_params.sample(
+                    n_starts=1,
+                    lb=problem.lb,
+                    ub=problem.ub,
+                    priors=problem.x_priors,
+                )[0]
+                x0 = warm_start * x0 + (1 - warm_start) * x0_prior
+        else:
+            logger.info("No initial point provided.")
```
dilpath: Should be an exception with message e.g. "set warm_start to 0"? Or how can the "warm start" setting be satisfied?

pypesto/startpoint/prior.py
```diff
@@ -0,0 +1,173 @@
...
+        xs: Startpoints, shape (n_starts, n_par).
+        """
+        # If priors are available, use prior.sample()
+        if priors is not None:
+            # Initialize full array with correct number of parameters
+            n_par = len(lb)
+            samples = np.zeros((n_starts, n_par))
+
+            # Identify which parameters have priors
+            # prior_samples_dict = {prior["index"]: np.zeros(n_starts) for prior in priors.prior_list}
```
dilpath:
```suggestion
```

test/base/test_prior.py
```diff
@@ -173,6 +173,35 @@ def test_derivatives(prior_type_list, scale):
...
+    Tests the sampling from the prior.
+    """
+
+    prior_list = [
+        get_parameter_prior_dict(
+            iprior,
+            prior_type,
+            (
+                [-1, 1]
+                if prior_type in ["uniform", "parameterScaleUniform"]
```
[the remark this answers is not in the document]
dilpath: Maybe a better test would be to generate many samples and check that the empirical probability is within some tolerance of the prior density (or cumulative function might make more sense). Since the prior density is tested elsewhere, it can be assumed true. Then sampling from all prior types can be tested easily (currently I think parameterScale is probably not implemented correctly?).

test/base/test_startpoint.py
```diff
@@ -137,3 +137,109 @@ def test_startpoints_from_problem():
...
+    # Create PriorStartpoints instance
+    startpoint_method = PriorStartpoints(use_guesses=False)
+
+    # Generate startpoints
+    xs = startpoint_method(n_starts=50, problem=problem)
+
+    # Check shape and bounds
+    assert xs.shape == (50, dim)
+    assert np.all(xs >= lb)
+    assert np.all(xs <= ub)
```
dilpath: Possible to test that it looks like a truncated distribution?

dilpath (changes requested):
Thanks, looks very useful

pypesto/objective/priors.py
```diff
@@ -225,6 +228,51 @@ def residual_jacobian(self, x):
...
+
+        for prior in self.prior_list:
+            index = prior["index"]
+            prior_type = prior.get("type", None)
+            prior_parameters = prior.get("parameters", None)
+
+            if prior_type is None or prior_parameters is None:
+                raise ValueError(
+                    "Prior type and parameters must be specified for sampling. "
+                    "Use 'get_parameter_prior_dict' to create proper prior dictionaries."
```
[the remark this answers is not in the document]
arrjon: What do you mean with 

> and provide them to `__init__`.

?

test/base/test_prior.py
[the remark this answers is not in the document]
arrjon: I included a new test now

dilpath (approved):
Thanks!

pypesto/objective/priors.py
```diff
@@ -225,6 +229,68 @@ def residual_jacobian(self, x):
...
+        seed:
+            Random seed for reproducibility.
+
+        Returns
+        -------
+        samples:
+            Dictionary of samples from the prior distributions for each
+            parameter index.
+        """
+        rng = np.random.default_rng(seed)
```
dweindl:
```suggestion
    def sample(
        self, n_samples: int = 1, rng: int = None
    ) -> dict[str, np.ndarray]:
        """
        Sample from the prior distribution. Parameters are returned on defined parameter scale.

        Parameters
        ----------
        n_samples:
            Number of samples to draw.
        rng:
            Random number generator or seed for reproducibility.

        Returns
        -------
        samples:
            Dictionary of samples from the prior distributions for each
            parameter index.
        """
        rng = np.random.default_rng(rng)
```

Let's stick to `rng`, this becoming kind of standard across various libraries.

dweindl (approved):
Thanks, looks good to me.

pypesto/objective/priors.py
dilpath: Then the typehint `rng: int` should also be adapted.

Doresic (approved):
Looks good! Useful stuff, thanks!
Took a somewhat detailed look, couldn't find anything looking off.

PR #1627, opened 2025-10-21
Allow to define colors more flexibly

plakrisenko:


stephanmg:
Looks fine, thanks for adding.

PR #1637, opened 2025-11-17
Make DlibOptimizer solver_epsilon configurable

### LINKED ISSUE #1492: DlibOptimizer -- don't hardcode solver_epsilon
`DlibOptimizer` currently has a hardcoded `solver_epsilon` (the magic `0.002`):

https://github.com/ICB-DCM/pyPESTO/blob/ead29b36db80d2eeee91b5d29bfbf4430137f8aa/pypesto/optimize/optimizer.py#L672-L678

My current interpretation of that parameter (documentation is a bit sparse) is that it's the minimum absolute objective function improvement (https://github.com/davisking/dlib/blob/6d29e0c7d41dfdffb8c0a0f83995caae84df5083/dlib/global_optimization/global_function_search.cpp#L754). For that, this value is potentially rather high. I am not sure if we should override dlib's default at all (their default is `0`), but at least it should be configurable by the user.

PaulJonasJost:
Change the hardcoded solver_epsilon value (0.002) to a configurable
option with a default of 0.0 (dlib's original default). Users can
now customize this parameter via the options dictionary.

Fixes #1492

### DIFF
--- a/pypesto/optimize/optimizer.py
+++ b/pypesto/optimize/optimizer.py
@@ -750,6 +750,30 @@ class DlibOptimizer(Optimizer):
         """Create default options specific for the optimizer."""
         return {"maxiter": 10000, "solver_epsilon": 0.0}
 
+    def _validate_options(self, options: dict) -> None:
+        """
+        Validate that provided options are valid for DlibOptimizer.
+
+        Parameters
+        ----------
+        options
+            Options dictionary to validate.
+
+        Raises
+        ------
+        ValueError
+            If invalid options are provided.
+        """
+        valid_options = set(self.get_default_options().keys())
+        provided_options = set(options.keys())
+        invalid_options = provided_options - valid_options
+
+        if invalid_options:
+            raise ValueError(
+                f"Invalid options for DlibOptimizer: {sorted(invalid_options)}. "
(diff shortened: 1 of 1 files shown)


dweindl (approved):
Thanks, looks good.

pypesto/optimize/optimizer.py
```diff
@@ -691,11 +691,7 @@ class DlibOptimizer(Optimizer):
     def __init__(self, options: dict = None):
         super().__init__()
 
-        self.options = options
-        if self.options is None:
-            self.options = DlibOptimizer.get_default_options(self)
-        elif "maxiter" not in self.options:
-            raise KeyError("Dlib options are missing the keyword maxiter.")
+        self.options = self.extend_options_with_defaults(options)
```
dweindl: Hm, it wasn't there before either - but maybe raise on any unsupported options? Currently, exactly two are supported. Would be good to alert the user if they, for example, misspell `maxiter` as `max_iter`.

vwiela: Would agree and also in favor for raising errors, if users missspecify anything.

vwiela (approved):

PR #1652, opened 2025-12-04
add low level 2d profile plot function and 2d profile function

### LINKED ISSUE #1651: Profiles do not show if dropoff due to bound hitting
When profiling it is not currently shown whether a dropoff in the likelihood is due to the data constraining the parameter value or whether there really is some structural identifiability but the dropoff happens because the other parameter hits a bound and is not able to progress along the non-identifiability valley.

For instance
If there is a structural non-identifiability that constrains a ratio of two parameters `k1` and `k2` to be constant, then the profile should be completely flat.
But if, as `k1` is increasing, `k2` hits its upper bound, the profiling is not able to keep the ratio optimally constant.  Thus the likelihood drops off. 

This would be fixed by either reporting whether the profiling hits a bound in any other parameter and when plotting the profile using a different marker for these points.

Alternatively, one can look at 2d plots of the profiling to see whether other parameters hit a bound as they progress through the value of the profiling parameter.

EchoRLiu:
This PR is to address #1651. 

To do so, I added two new visualization functions: 

1. `profile_lowlevel_2d()` - Visualizes the relationship between two parameters during profiling
- Shows one profiled parameter on the x-axis and tracks a second parameter on the y-axis
- Uses color mapping (via scatter plot) to indicate objective ratio or function values along the profile path

2. `visualize_2d_profile()` - High-level n by n profile grid:
- Diagonal plots: Shows traditional 1D profiles for each parameter
- Off-diagonal plots: Shows 2D parameter relationships during profiling using `profile_lowlevel_2d()`

This would provide a complete overview of how parameters interact during profile likelihood computation


Doresic:
Remaining TODO list from @EchoRLiu 

> TODO List for still needed improvements:
> 
> * [ ]  there is some weird jumps when doing profile

This will be taken care of in the another PR dealing with profiling enhancements.

> * [ ]  some range can not be profiled

This is weird, I saw it a couple of times. We'll need to check what happens, why does profiling stop prematurely sometimes. It might be the weird conditions in adaptive searches that doesn't have a else, so profiling can fail silently. Needs to be investigated.

> * [ ]  color scheme can be improved

Good as initial visualization, can be changed via user.

> * [ ]  original profile function can also be improved with threshold

For another PR as it deals with general profiling.

Doresic (approved):

dweindl:
Thanks, looks mostly good to me. 

Two minor things looking at https://pypesto--1652.org.readthedocs.build/en/1652/example/getting_started.html:
image

* I find the top-left y-label confusing. how about showing all y-labels, or at least always for the first, the diagonal, and the first off-diagonal subplot?
* the markers in the off-diagonal plots look all black. I guess this is just due to the overlap here. it might be preferable to plot them from low to high log-posterior ratio or setting the z-order accordingly.

PaulJonasJost (approved):
One thing that is bugging me a bit is that if we have a lot of points in a small intervall, the black borders of the point overshadow the coloring.

Doresic:
Ok some updates
- some separation of the plots
- y and x labels on each of the plots, bolded so it's easy to read
- fixed the black border (removed) and plotting the better points on top on the worse ones, so we see the better areas nicer
Some examples:
The same example from getting started:
image
And another example (Boehm). It has huge parameter names so they kinda overlap a bit, but I wouldn't reduce the plot font sizes, as usually parameters aren't so large.
boehm_2d_profile

PR #1654, opened 2025-12-08
Unified Interface: x0

PaulJonasJost:
Removing x0 as necessity from minimize.


dilpath (approved):
Fine for me except for Daniel's breaking change point.

PR #1658, opened 2025-12-15
Minor Code refactoring on pymc Sampler

PaulJonasJost:
slightly cleaner check for import of pymc and arviz. Left over  from #1643


arrjon (approved):
looks good to me

dilpath (approved):

PR #1660, opened 2025-12-15
Unified Interface: Absolute Tolerance

PaulJonasJost:
Add tolerance to the unified interface. Some optimizers not supported, especially Dlib with epsilon, though epsilon is not strictly a tolerance!

### DIFF
--- a/pypesto/optimize/optimizer.py
+++ b/pypesto/optimize/optimizer.py
@@ -1246,8 +1246,17 @@ class CmaOptimizer(Optimizer):
         ----------
         tol
             Absolute tolerance for termination.
+
+        Raises
+        ------
+        ValueError
+            If tolerance is not positive.
         """
-        self._set_option_tol(tol, "tolfun")
+        if tol < 0:
+            raise ValueError(f"Tolerance must be positive, got {tol}")
+        if self.options is None:
+            self.options = {}
+        self.options["tolfun"] = tol
 
 
 class CmaesOptimizer(CmaOptimizer):


pypesto/optimize/optimizer.py
```diff
@@ -413,6 +413,58 @@ def set_maxeval(self, evaluations: int) -> None:
...
+        ------
+        NotImplementedError
+            If the optimizer does not support absolute function tolerance.
+        """
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not support absolute function tolerance. "
+            f"Check supports_f_abs_tol() before calling set_f_abs_tol()."
+        )
+
+    def _set_option_tol(self, tol: float, option_key: str) -> None:
```
dweindl: `class Optimizer` does not have an `options` attribute, we should not try accessing it here. I'd leave validation to the optimizer and just set the options directly in each optimizer without this extra method.

pypesto/optimize/optimizer.py
```diff
@@ -822,9 +892,7 @@ def set_maxtime(self, seconds: float) -> None:
             )
         if self.options is None:
             self.options = {}
-        # We explicitly cast to float, as the IpoptOptimizer requires
-        # the provision of a float for the max_wall_time option.
-        self.options["max_wall_time"] = float(seconds)
```
dweindl, line 827: Why removed?

PaulJonasJost: that should not have happened. I think something got mixed up when i merged develop...

pypesto/optimize/optimizer.py
```diff
@@ -789,6 +859,17 @@ def set_maxiter(self, iterations: int) -> None:
...
+    def set_tol(self, tol: float) -> None:
+        """
+        Set the absolute tolerance for optimization.
+
+        Parameters
+        ----------
+        tol
+            Absolute tolerance for termination.
+        """
+        self._set_option_tol(tol, "tol")
```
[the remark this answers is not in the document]
PaulJonasJost: True, it says so in the IPOPT documentation with a quite lengthy passage. I would in general be fine with removing it, but i am not sure how to handle the `supports_rel_tol` later. I think we can say it does not directly support `f_abs_tol` but this one is a bit harder. Removed it for now!

PR #1664, opened 2025-12-16
Change from Pool to ProcessPoolExecutor

PaulJonasJost:
Changed to ProcessPoolExecutor according to "concurrent.futures" documentation.

I have the indexing in since we are not using imap anymore. Do we need a tests that sees we are returning them in the right oder? Probably not necessary to me, as we are anyways sorting them shortly after.


dweindl (commented):
> I have the indexing in since we are not using imap anymore. Do we need a tests that sees we are returning them in the right oder? Probably not necessary to me, as we are anyways sorting them shortly after.

The `MultiProcessEngine` is not restricted to multi-start optimization, so it shouldn't rely on that and we should ensure that result ordering corresponds to task ordering. So yes, having a test would be good. Doesn't have to be any kind of optimization, can be some trivial task.

vwiela (approved):

PR #1668, opened 2025-12-18
Fix Gradient Check Test

### LINKED ISSUE #1495: Test for gradient check asserts that gradients are wrong
The test for [`PetabImportTest.test_check_gradients`](https://github.com/ICB-DCM/pyPESTO/blob/58c7b5fcdcc54690f80f8d4e3e407e39420560b9/test/petab/test_petab_import.py#L117-L136) asserts that the gradients are `False`, in contrast to the description in L119.

PaulJonasJost:
Closes #1495

Cause for this was running the gradient check in `mode_res` with a noise parameter. This always ran into a runtime error and was evaluated as False. Fixed with `amici_model.setAddSigmaResiduals(True)` and correct assertion. Not sure how this got in.


dweindl (commented):
The `test/base/test_engine.py` changes belong somewhere else, but the rest looks fine to me.

vwiela (approved):

PR #1671, opened 2026-01-16
Support walltime limit in `ScipyOptimizer`

dweindl:
SciPy optimizers don't support wall time limits directly. However, this can be achieved through callbacks. This is done here.


PaulJonasJost (approved):
Thats an amazing workaround, never thought about it 🙈 thanks 🙏🏼

PR #1679, opened 2026-02-12
pyproject.toml / pyupgrade

dweindl:
* Fully migrate to pyproject.toml: Replace the old setup.cfg / setup.py files by pyproject.toml to have everything in one place. This also ensures that ruff picks up the current python requirement
* precommit autoupdate
* ruff: pyupgrade for Python 3.11
  * `zip(..., strict=True)`
  * type annotations: `Union` -> `|`
  * ...


dilpath (approved):
👍

PR #1685, opened 2026-02-19
Select: set optimal point to fixed vals if no parameters are estimated

dilpath:
Models with no estimated parameters can appear during model selection. This ensures the fixed values are assigned as the optimal point in this case.


dweindl (approved):
Thanks! Looks good to me.

PR #1690, opened 2026-02-27
Add deprecation warning in SacessOptimizer and ESSOptimizer

### LINKED ISSUE #1641: Sacess usage example or simplification of usage
It is a bit hard to start using sacess in pyPESTO. There is an example of usage, but it's only in the docstring of the `SacessOptimizer`. It is a place I would search for, but I don't think that any user would find it quickly. Having an example of usage, a notebook, could help with this. To also give some idea of what it does and which settings one should choose.

It could be a part of "optimizer_choice" notebook that would showcase some of the most common optimizers we supply and which are recommended (according to some preliminary optimizer benchmarking results).

A separate, but connected point:
I believe a user's first idea of how they would select sacess would be
```
sacess_optimizer = SacessOptimizer(options)
pypesto.optimize.minimize(problem, optimizer=sacess_optimizer)
```
I see issues with this approach and why it's separate, but it would be the most user-friendly approach imo.

dweindl:
Now available from https://github.com/ICB-DCM/pyscat/. Closes #1641.

### DIFF
--- a/pypesto/optimize/ess/ess.py
+++ b/pypesto/optimize/ess/ess.py
@@ -199,6 +199,15 @@ class ESSOptimizer:
             RefSet, or just the local search results and the overall best
             parameters.
         """
+        warn(
+            "Using pypesto.optimize.ess.ESSOptimizer is deprecated "
+            "and will be removed in a future release. "
+            "Use pyscat.ESSOptimizer instead "
+            "(https://github.com/ICB-DCM/pyscat/, `pip install pyscat`).",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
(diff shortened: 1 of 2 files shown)


PaulJonasJost (approved):
Looks fine by me. A question thouh:
All other optimizers are interfaced in pypesto again (e.g. pypesto.FidesOptimizer). Won't we want something like this here long term as well?

dweindl:
> All other optimizers are interfaced in pypesto again (e.g. pypesto.FidesOptimizer). Won't we want something like this here long term as well?

Not sure. One problem is that, unlike other interfaced optimizers, pyscat depends on pypesto. Cyclic dependencies are usually a pain in the neck.

PR #1691, opened 2026-03-03
doc: More readable figures in examples

dweindl:
* Update figsize/fontsize.
* Less verbose profiles
* Fix typos
* Fix type annotation
* Fix cropped waterfall markers

before: https://pypesto.readthedocs.io/en/latest/example/custom_objective_function.html
after: https://pypesto--1691.org.readthedocs.build/en/1691/example/custom_objective_function.html


PaulJonasJost (approved):
True looks much nicer!

PR #1692, opened 2026-03-03
Make `id` optional in `Optimizer.minimize`

dweindl:
If no history filename is set, the ID is not used, so we shouldn't require it.

### DIFF
--- a/pypesto/optimize/optimizer.py
+++ b/pypesto/optimize/optimizer.py
@@ -95,7 +95,7 @@ def history_decorator(minimize):
     ):
         if history_options is None:
             history_options = HistoryOptions()
-        if id is None:
+        if history_options.storage_file is not None and id is None:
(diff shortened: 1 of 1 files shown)


PaulJonasJost (approved):
Fine by me. Though to be fair, I kept this in since IDs should always be assigned by the time we get to this pont from pypesto.minimize (where it already is optional). But fine by me to merge

dweindl:
> I kept this in since IDs should always be assigned by the time we get to this pont from pypesto.minimize (where it already is optional).

The problem is:
```python
from pypesto.optimize import FidesOptimizer
from pypesto import Objective, Problem

problem = Problem(Objective(fun=lambda x: x), lb=[-1], ub=[1])
FidesOptimizer().minimize(problem)
```
fails with `ValueError: id must be provided for history tracking.` which makes absolutely no sense.

Doresic (approved):

PR #1694, opened 2026-03-09
Fix deployment workflow

dweindl:
https://github.com/ICB-DCM/pyPESTO/pull/1679 broke the deployment workflow. pypesto 0.5.9 was not deployed to PyPI (https://github.com/ICB-DCM/pyPESTO/actions/runs/22439382257), I just uploaded it manually.

This updates the deployment workflow to work with the current `pyproject.toml`.


vwiela (approved):
Thanks for fixing this.

PR #1695, opened 2026-03-11
Make parameters_correlation_matrix more customizable

dweindl:
* Forward heatmap kwargs
* Use non-zero linewidth (for old behavior, pass `heatmap_kwargs={"linewidth": 0}`)
* enable documentation cross references to seaborn
* add  `size` argument

Turns 
image
into
image

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -102,6 +102,7 @@ intersphinx_mapping = {
     "petab_select": ("https://petab-select.readthedocs.io/en/develop/", None),
     "python": ("https://docs.python.org/3", None),
     "scipy": ("https://docs.scipy.org/doc/scipy/", None),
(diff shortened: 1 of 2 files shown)


vwiela (approved):

PaulJonasJost (approved):
Could we add the classical `ax` and `size` arguments to it as well (just as in e.g. `waterfall`)?

dweindl:
> Could we add the classical `ax` and `size` arguments to it as well (just as in e.g. `waterfall`)?

`ax`: no, not supported by clustermap.

PaulJonasJost:
> > Could we add the classical `ax` and `size` arguments to it as well (just as in e.g. `waterfall`)?
> 
> `ax`: no, not supported by clustermap.

Yeah, sorry, was looking at sns.heatmap 🙈

PR #1696, opened 2026-03-13
Make validate_inner_parameter_pairings pandas 3 compatible

dweindl:
For reasons I don't fully understand, `None` gets converted to `nan` in `iterrows` in pandas 3.0.1 [breaking](https://github.com/ICB-DCM/pyPESTO/actions/runs/23039704731/job/66915208725) this function. This is not the case with `itertuples`.


PaulJonasJost (approved):
Thanks

PR #1705, opened 2026-03-30
Paused Julia Tests

PaulJonasJost:
Temporarily paused Petab.jl tests. Needs further investigation and discussion


vwiela (approved):
Fine for now, but we should try to get that fixed soon.

PR #1706, opened 2026-03-31
Fides optimizer: clarify use of Hessian and FIM

dilpath:
Currently it looks like a user needs to specify `pypesto.optimize.FidesOptimizer(..., hessian_update=None)` to ensure Fides uses the provided Hessian. It appears that default behavior does not use the Hessian from the objective when possible.

### DIFF
--- a/pypesto/optimize/optimizer.py
+++ b/pypesto/optimize/optimizer.py
@@ -1779,7 +1779,7 @@ class FidesOptimizer(Optimizer):
                 warnings.warn(
                     "A hybrid Hessian approximation strategy will be "
                     "employed. See the docstring for ``hessian_update`` in "
-                    "the class constructor for more details."
+                    "the class constructor for more details.",
                     stacklevel=1,
                 )
                 _hessian_update = fides.HybridFixed()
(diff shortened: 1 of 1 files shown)


FFroehlich:
from what I recall `fides.HybridFixed()` initially uses the provided Hessian approximation and then switches to BFGS after a fixed number of iterations. Based on the benchmarking I did back then, this is a reasonable default for FIM approximations, which is likely what most users would work with.

dweindl:
At least in the current (`develop`) version, the docstring doesn't match the implementation:

https://github.com/ICB-DCM/pyPESTO/blob/86a0076bcc6bdb28fa43906be6bdb8c39efb08d4/pypesto/optimize/optimizer.py#L1702-L1705

`hessian_update=None` would use the provided Hessian, not the hybrid scheme.

dilpath:
Thanks for the feedback -- I tried to resolve this by changing some docstrings.

pypesto/optimize/optimizer.py
```diff
@@ -1771,6 +1776,12 @@ def minimize(
                 )
                 _hessian_update = fides.BFGS()
             else:
+                warnings.warn(
```
PaulJonasJost: I find it puzzling that we print a warning in a default setting. Rather a debug? (unless we expect the user to set a value usually)

PaulJonasJost (approved):
Thanks for this 🙏🏼

pypesto/optimize/optimizer.py
dilpath: Sure, done now (also for the old warning). I also updated the docstring again to make explicit the assumption that the problem objective is actually the FIM.

FFroehlich (approved):

PR #1708, opened 2026-04-21
Fix remaining AMICI 1.0 calls and extend tests

Doresic:
Some small changes to fix some bugs that occurred recently.

I updated the remaining AMICI calls in the model-fit visualization code (get_timepoints, set_sensitivity_order, set_timepoints, get_state_names) and aligns related docstrings and error messages with the same naming.

Also I expanded visualization test coverage by:
- directly covering _get_simulation_rdatas(..., simulation_timepoints=None) to exercise default AMICI timepoint handling 
- extending time_trajectory_model(...) coverage to include hierarchical problems and named-state selection

### DIFF
--- a/pypesto/hierarchical/ordinal/solver.py
+++ b/pypesto/hierarchical/ordinal/solver.py
@@ -236,7 +236,7 @@ class OrdinalInnerSolver(InnerSolver):
         par_sim_ids:
             Ids of outer simulation parameters, includes fixed parameters.
         par_edata_indices:
-            Indices of parameters from `amici_model.getParameterIds()` that are needed for
+            Indices of parameters from `amici_model.get_parameter_ids()` that are needed for
             sensitivity calculation. Comes from `edata.plist` for each condition.
         snllh:
             A zero-initialized vector of the same length as ``par_opt_ids`` to store the
--- a/pypesto/hierarchical/ordinal/solver.py
+++ b/pypesto/hierarchical/ordinal/solver.py
@@ -236,7 +236,7 @@ class OrdinalInnerSolver(InnerSolver):
         par_sim_ids:
             Ids of outer simulation parameters, includes fixed parameters.
         par_edata_indices:
-            Indices of parameters from `amici_model.get_parameter_ids()` that are needed for
+            Indices of parameters from `amici_model.get_free_parameter_ids()` that are needed for
             sensitivity calculation. Comes from `edata.plist` for each condition.
         snllh:
             A zero-initialized vector of the same length as ``par_opt_ids`` to store the
--- a/test/visualize/test_visualize.py
+++ b/test/visualize/test_visualize.py
@@ -1191,6 +1191,9 @@ def test_get_simulation_rdatas_default_timepoints():
     result = optimize.minimize(
         problem=problem,
         n_starts=1,
+        optimizer=optimize.ScipyOptimizer(
+            method="L-BFGS-B", options={"maxiter": 1}
+        ),
         progress_bar=False,
     )
 
--- a/test/visualize/test_visualize.py
+++ b/test/visualize/test_visualize.py
@@ -25,6 +25,7 @@ from pypesto.testing.examples import (
     get_Boehm_JProteomeRes2014_hierarchical_petab_corrected_bounds,
 )
 from pypesto.visualize.model_fit import (
+    _get_simulation_rdatas,
     time_trajectory_model,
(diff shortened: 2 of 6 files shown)


FFroehlich (approved):

pypesto/hierarchical/ordinal/solver.py
```diff
@@ -236,7 +236,7 @@ def calculate_gradients(
         par_sim_ids:
             Ids of outer simulation parameters, includes fixed parameters.
         par_edata_indices:
-            Indices of parameters from `amici_model.getParameterIds()` that are needed for
+            Indices of parameters from `amici_model.get_parameter_ids()` that are needed for
```
dweindl:
```suggestion
            Indices of parameters from `amici_model.get_free_parameter_ids()` that are needed for
```

test/visualize/test_visualize.py
```diff
@@ -1182,6 +1183,22 @@ def test_sampling_prediction_trajectories():
     )
 
 
+@close_fig
+def test_get_simulation_rdatas_default_timepoints():
+    """Regression test for default AMICI timepoint access in model-fit plots."""
+    problem = create_petab_problem()
+
+    result = optimize.minimize(
```
dweindl, line 1240: Limit to 1 iteration as done below?

dweindl (approved):
👍 thanks

vwiela (approved):

test/visualize/test_visualize.py
```diff
@@ -1291,7 +1311,41 @@ def test_time_trajectory_model():
...
+    )
+    importer = pypesto.petab.PetabImporter(
+        hierarchical_petab_problem, hierarchical=True
+    )
+    helper_problem = importer.create_problem()
+
+    x_guesses = np.asarray(hierarchical_petab_problem.x_nominal_scaled)[
+        helper_problem.x_free_indices
+    ]
+    problem = importer.create_problem(x_guesses=[x_guesses])
```
PaulJonasJost: why add x_guesses here?

test/visualize/test_visualize.py
```diff
@@ -1291,7 +1311,41 @@ def test_time_trajectory_model():
...
+
+    result = optimize.minimize(
+        problem=problem,
+        n_starts=1,
+        optimizer=optimize.ScipyOptimizer(
+            method="L-BFGS-B", options={"maxiter": 1}
+        ),
```
PaulJonasJost, line 1392:
```suggestion
```
Don't think a specific optimizer is necessary right?

test/visualize/test_visualize.py
```diff
@@ -1291,7 +1311,41 @@ def test_time_trajectory_model():
...
 
     # test call of time_trajectory_model
-    time_trajectory_model(result=result)
+    axes = time_trajectory_model(result=result)
+    assert axes is not None
+
+    # test call of time_trajectory_model for hierarchical problems
+    hierarchical_petab_problem = (
+        get_Boehm_JProteomeRes2014_hierarchical_petab_corrected_bounds()
+    )
```
PaulJonasJost, line 1369: In general I Question whether we shouldn't create a problem and a corresponding hierachical problem and test all figures/pipelines with both (or only where we think it's gonna play a role) as we do with e.g. our optimizers. Would make the code cleaner i think.

PaulJonasJost (approved):
Looks good to me. thanks for adding a test for hierachical.

test/visualize/test_visualize.py
Doresic: Yeah good idea, we should scan testing and see whether we need to include specifically hierarchical tests of varying things. Will write down/make an issue on it.

test/visualize/test_visualize.py
Doresic: Yeah probably not
Just chose L-BFGS-B since it's fast and we don't really need to optimise since we have nominal points, so I put maxiter to 1 to basically completely remove optimisation time
Just to get a quick result through the pipeline. So I would keep this.

test/visualize/test_visualize.py
Doresic: We're not testing optimisation but need a optimised result object, so I'm just using x_guesses to pass nominal parameters as the startpoint to reduce optimisation time and get a reproducible result.
However there is a small problem here that I'll fix.
Using this helper_problem is ok, but not the nicest. It could brake if the order of parameters in the petab file is not mehcanistic first -- hierarchical last

PR #1709, opened 2026-05-04
Fix: harmonise visualize style -- axes, returns, attributes

Doresic:
This PR lays the foundation for a coherent pyPESTO visualization API. It primarily normalizes the *contracts* that callers depend on, so that downstream style work can be applied uniformly, with a few small review-driven follow-up changes to helper placement, layout handling, and observable-mapping axes handling.

### What changed

**Shared helpers in `pypesto/visualize/misc.py`**
Shared helpers used by all plotters:
- `get_ax(ax, size)` — returns the provided axes or creates a new figure/axes pair
- `get_axes_array(axes, nrows, ncols, size)` — returns the provided 2-D array or creates a new subplot grid; validates shape if an existing array is passed
- `plot_diagonal_marginal(ax, values, diag_kind)` — shared KDE/histogram for scatter-matrix diagonals
- `process_deprecated_kwarg(...)` — generic rename shim with `DeprecationWarning`
- `make_grid_shape(n_panels)` — shared near-square subplot-grid helper

**Type hints** (all visualize files)
- `tuple` → `tuple[float, float]` for `size` parameters
- `ax: Axes = None` → `ax: Axes | None = None` throughout
- Added `-> matplotlib.axes.Axes` and `-> np.ndarray` return annotations where missing

**Single-panel ax contract**
All single-panel public functions now accept `ax: matplotlib.axes.Axes | None` and return `matplotlib.axes.Axes`. 

**Multi-panel axes grid contract**
All multi-panel public functions now accept `axes: np.ndarray | None` and return a 2-D `np.ndarray` of `Axes`. `optimization_scatter` rewritten from seaborn PairGrid to pure matplotlib: per-panel KDE diagonals, fval-encoded viridis_r colorbar, correct `lb_full/ub_full[parameter_indices[i]]` bound indexing.

**`sampling_parameter_cis`: `alpha` → `confidence_levels`**
- Old: `alpha: Sequence[int]` (integer percentages, e.g. `[95, 68]`)
- New: `confidence_levels: Sequence[float]` (fractions, e.g. `[0.95, 0.68]`)
- `alpha` kept as a deprecated kwarg; auto-converts with `DeprecationWarning`

**`profiles`: `confidence_level` convenience parameter**
New optional `confidence_level: float | None = None` alongside `ratio_min`; converts via `chi2_quantile_to_ratio`.

**Layout handling**
Where figures are created internally, they now use constrained layout instead of forcing `tight_layout()` afterward.

**Tests**
- helper tests moved from `test/visualize/test_style.py` to `test/visualize/test_misc.py`
- `close_fig` moved to `test/conftest.py` so it can be reused across visualization tests

### What this PR does NOT touch
- Default figure sizes, font sizes, or any `rcParams`
- Spine removal, colorbar style unification (planned for follow-up)
- `model_fit.py`, `ordinal_categories.py` internals


pypesto/visualize/ensemble.py
```diff
@@ -6,13 +6,14 @@
...
+from ._style import get_ax
 
 
 def ensemble_identifiability(
     ensemble: Ensemble,
     ax: plt.Axes | None = None,
-    size: tuple[float] | None = (12, 6),
-):
+    size: tuple[float, float] | None = (12, 6),
+) -> plt.Axes:
```
vwiela: Should this rather be `matplotlib.axes.Axes` as in the other files?

pypesto/visualize/ensemble.py
```diff
@@ -56,8 +57,8 @@ def ensemble_identifiability_lowlevel(
     ub_hit: np.ndarray,
     both_hit: np.ndarray,
     ax: plt.Axes | None = None,
-    size: tuple[float] | None = (16, 10),
-):
+    size: tuple[float, float] | None = (16, 10),
+) -> plt.Axes:
```
vwiela: Same here

pypesto/visualize/sampling.py
```diff
@@ -707,7 +709,7 @@ def _handle_colors(
 
     # define colormap
     variable_colors = [
-        list(cmap(v))[:LEN_RGB]
+        list(matplotlib.cm.viridis(v))[:LEN_RGB]
         for v in np.linspace(cmap_min, cmap_max, n_variables)
     ]
```
vwiela, line 717: This whole function is pretty tailored to this one specific colormap. 
Since we are touching visualizations anyways, we could think about making this more verstaile. So a user could use different colormaps and cut-offs (0.3, 0.85) and we leave the current one as defaults.

pypesto/visualize/dimension_reduction.py
```diff
@@ -71,8 +73,9 @@ def projection_scatter_umap_original(
     umap_object: UmapTypeObject,
     color_by: Sequence[float] = None,
     components: Sequence[int] = (0, 1),
+    ax: plt.Axes | None = None,
     **kwargs,
-):
+) -> plt.Axes:
```
vwiela: Shouldn't this be `matplotlib.axes.Axes`?

vwiela (changes requested):

pypesto/visualize/_style.py
```diff
@@ -0,0 +1,184 @@
+"""
+Shared styling helpers for ``pypesto.visualize``.
+
+Cross-cutting pieces used by multiple plotters live here so each plotter
+does not reinvent them.
```
PaulJonasJost:
```suggestion
```
Superfluous

pypesto/visualize/optimization_stats.py
```diff
@@ -195,41 +197,46 @@ def optimization_run_properties_per_multistart(
...
     opt_run_property: str,
-    axes: matplotlib.axes.Axes | None = None,
+    axes: matplotlib.axes.Axes | np.ndarray | None = None,
     size: tuple[float, float] = (18.5, 10.5),
     start_indices: int | Iterable[int] | None = None,
     colors: COLOR | list[COLOR] | np.ndarray | None = None,
     legends: str | list[str] | None = None,
     plot_type: str = "line",
-) -> matplotlib.axes.Axes:
+) -> np.ndarray:
```
PaulJonasJost, line 239: Yes it is an array, yes it says that the array consist of axes, yet i find the previous `matplotlib.axes.Axematplotlib.axes.Axes` still more informative

pypesto/visualize/parameters.py
```diff
@@ -663,72 +656,175 @@ def optimization_scatter(
...
-        result.problem.x_names[parameter_index]
-        for parameter_index in parameter_indices
-    ]
-    df = pd.DataFrame(parameters, columns=x_labels)
 
-    sns.set(style="ticks")
+    n = len(parameter_indices)
+    x_labels = [result.problem.x_names[i] for i in parameter_indices]
 
-    ax = sns.pairplot(
```
PaulJonasJost, line 718: If I see this correctly, we have changed something like 10 lines of code for 120, for an improvement that to me is not immediately visible.

If the goal is to move to purely matplotlib, that might be a reason, but I don't see why we would push for that necessarily.

Summary of this specific change (my opinion, to be discussed): This does not improve the code, rather the opposite.

pypesto/visualize/profiles.py
```diff
@@ -408,10 +424,10 @@ def profile_lowlevel(
     ax: plt.Axes | None = None,
```
PaulJonasJost: No changes here to the `Axes`? If we make a point of harmonising it, please check all functions.

pypesto/visualize/_style.py
```diff
@@ -0,0 +1,184 @@
...
+        else:
+            raise ValueError(f"Pass `axes` with shape ({nrows}, {ncols}).")
+
+    if axes_array.shape != (nrows, ncols):
+        raise ValueError(f"Pass `axes` with shape ({nrows}, {ncols}).")
+
+    return axes_array
+
+
+def plot_diagonal_marginal(
```
PaulJonasJost: This is not really a style thing right? In general, why isn't this whole file put into misc.py or in a utility.py which we rather use?

pypesto/visualize/observable_mapping.py
```diff
@@ -41,13 +42,90 @@
...
+    if n_panels < 1:
+        raise ValueError("At least one panel is required.")
+    n_rows = int(np.ceil(np.sqrt(n_panels)))
+    n_cols = int(np.ceil(n_panels / n_rows))
+    return n_rows, n_cols
+
+
+def _normalize_axes_array(
+    axes: matplotlib.axes.Axes | np.ndarray,
+) -> np.ndarray:
```
PaulJonasJost: This seems to be similar to identical to the _style.py function and should be merged IMO

pypesto/visualize/observable_mapping.py
```diff
@@ -41,13 +42,90 @@
...
+    if created_axes:
+        _, axes = plt.subplots(n_rows, n_cols, squeeze=False, **kwargs)
+    else:
+        axes = get_axes_array(axes=axes, nrows=n_rows, ncols=n_cols)
+
+    for ax in axes.flat:
+        ax.clear()
+        ax.set_visible(True)
+    for ax in axes.flat[n_panels:]:
+        ax.set_visible(False)
```
PaulJonasJost: This came up in multiple other changes to function, please check as well and actually harmonize.

pypesto/visualize/observable_mapping.py
```diff
@@ -41,13 +42,90 @@
...
+    measurements: np.ndarray,
+) -> None:
+    """Plot measurement points with the shared visualization styling."""
+    ax.scatter(
+        simulation,
+        measurements,
+        color="C0",
+        s=40,
+        alpha=0.9,
+        linewidths=0.6,
```
PaulJonasJost, line 74: Actually not quite sure, but hardcoding these values will result in rc.Params not overwriting this right? In that case this would be kind of a downside if we wanna move towards more flexible plotting

pypesto/visualize/observable_mapping.py
```diff
@@ -625,45 +710,59 @@ def plot_splines_from_inner_result(
...
-        axes[ax_index].plot(
+        flat_axes[ax_index].plot(
             spline_bases,
             spline_knots,
             linestyle="-",
-            color="g",
+            color="C2",
+            linewidth=1.8,
             label="Spline function",
+            zorder=2,
```
PaulJonasJost, line 705: I again feel like with more hardcoded style parameters, we move away from a flexible visualisation approach.

pypesto/visualize/sampling.py
```diff
@@ -1117,22 +1146,25 @@ def sampling_parameter_traces(
         i_chain=i_chain,
         stepsize=stepsize,
         full_trace=full_trace,
-        par_indices=par_indices,
+        parameter_indices=parameter_indices,
```
PaulJonasJost, line 1152: Needs a depracation hint as in lines 1359

pypesto/visualize/sampling.py
```diff
@@ -1212,50 +1245,86 @@ def sampling_scatter(
...
         result=result, i_chain=i_chain, stepsize=stepsize
     )
 
-    sns.set(style="ticks")
+    if size is None and axes is None:
+        size = (2.5 * nr_params + 0.5, 2.5 * nr_params + 0.5)
 
-    # TODO: Think this throws the axis errors in seaborn.
-    ax = sns.pairplot(
-        params_fval.drop(["logPosterior", "iteration"], axis=1),
```
PaulJonasJost, line 1229: Same argument as in the other scatterplot. Is this change useful?!

pypesto/visualize/sampling.py
```diff
@@ -994,31 +998,46 @@ def sampling_parameter_cis(
...
+            "Note: units have changed — pass fractions in (0, 1) "
+            "(e.g. `confidence_levels=[0.95]`) instead of integer percentages "
+            "(e.g. `alpha=[95]`). Your values have been divided by 100 automatically.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        confidence_levels = [a / 100 for a in alpha]
+
+    if confidence_levels is None:
+        confidence_levels = [0.95]
```
PaulJonasJost, line 1018: This seems to be a target that should also be covered by `process_deprecated_kwarg` right?

pypesto/visualize/_style.py
```diff
@@ -0,0 +1,184 @@
...
+            ax.set_ylabel("Density")
+            return
+        except np.linalg.LinAlgError:
+            pass
+
+    ax.hist(values, bins="auto", color=color, alpha=0.6)
+    ax.set_ylabel("Count")
+
+
+def process_deprecated_kwarg(
```
PaulJonasJost: I feel this is an amazing idea. So amazing, this should not be confined to visualization alone and also perhaps not to specific instances, rather have this as some kind of decorator on top of a function. Instead of passing the value we can have some kind of transformation function (either identity or in this case for example y = x/100), this would be great for maintaining this and can be easily removed at the next major release as discussed in our versioning policy.

PaulJonasJost (changes requested):
Some great ideas among this. Not entirely sure what the big picture is, you want to move towards with this initial PR as the AI summary hints here. Would perhaps be important to discuss this once in the near future.

pypesto/visualize/optimization_stats.py
Doresic: Hmmm yeah I see the point. `matplotlib.axes.Axes` is more informative, but in this case it was wrong to use it.
The code defines axes, a np.ndarray of axes. In both cases.
- if plot_type!="both" it creates a (1,1) array of Axes objects
- if plot_type=="both" it creates a (1,2) array of Axes objects
so with the current code of the visualisation, `matplotlib.axes.Axes` would always have been wrong.

What we can do is make it return really just a `matplotlib.axes.Axes` object in the first case, and then say the return type is either this or that depending on the settings. But I would vote for current situation instead, to make each function always have one type of output instead of the output type changing depending on input parameters.

pypesto/visualize/parameters.py
Doresic: Yeah, true
Initial reason for changing this code was to go away from the different return object (in the idea of harmonisation) -- a Seaborn grid object -- compared to the object we return in most cases -- either Axes or a np.ndarray of them.
Furthermore, a second PR later on would focus on harmonisation of general looks of the figures across the module. In the sense that the examples we have in docs have figures with similar fonts, color schemes, and general looks. So moving only to matplotlib would make that simpler. For instance, it's nice to have all plots of this type, a scatter plot with diagonal distributions, have a colorbar on the side to map objective function value to the scatter plot dots. This is a bit tough or would be very different from other ones if we stuck with seaborn.

So I would vote for this. I do see it's more lines just since matplotlib doesn't have an integrated visualisation of this type of a plot, but this allows us a nicer freedom of playing with it to make "our style" across the module.

pypesto/visualize/profiles.py
Doresic: ty, missed some

pypesto/visualize/_style.py
Doresic: True, was thinking about that
feels nicer in misc.py

pypesto/visualize/observable_mapping.py
Doresic: Hmmm true
I just wanted to make the observable mapping plotting a bit nicer, as it looked really outdated compared to the other visualisations.
In the next PR, I would generally target visualisation style across all visualisations in pyPESTO to unify a real style in a way.
What we could do as part of that next PR is allow for passing of kwargs to all plotting routines to allow changes to scatter plot sizes or similar things. But I would keep that for next PR.
How does that sound?

pypesto/visualize/observable_mapping.py
Doresic: similar to above

pypesto/visualize/sampling.py
Doresic: I think so yeah
Reasoning above

pypesto/visualize/sampling.py
Doresic: Yeah, but this one is a bit different. The kwarg deprecation is only changing the naming of some argument.
This is much more specific deprecation -- the value users provide has to be changed as well. So I made a custom one for here to point this out to not cause very plotting outcomes for people without warning.

pypesto/visualize/ensemble.py
Doresic: Yepp will change

pypesto/visualize/ensemble.py
Doresic: Yepp will change

pypesto/visualize/_style.py
Doresic: Yeah, good idea
For now I made it only in this module, so I kept it at this point.
But definitely a good idea for a separate whole-package-wide PR introducing this concept for further use
Apart from a transformation function, then also a custom message should also be possible to add. Cause without it, the warning would be too vague in case of an added transformation. Or maybe just messages depending on the transformation.
Definitely a good addition for whole pyPESTO

pypesto/visualize/_style.py
```diff
@@ -0,0 +1,184 @@
...
+        Forwarded to :func:`warnings.warn`. Default 3 attributes the
+        warning to the caller of the public function that invoked this
+        helper.
+
+    Returns
+    -------
+    value:
+        The resolved value, or ``None`` if neither was given.
+    """
+    if deprecated_value is None:
```
dweindl: This will miss `f(deprecated_kwarg=None)`, right?

pypesto/visualize/dimension_reduction.py
dweindl: Doc incompatible with return type annotation.

pypesto/visualize/dimension_reduction.py
dweindl: Either is fine.

pypesto/visualize/dimension_reduction.py
dweindl: Doc incompatible with return type annotation.

pypesto/visualize/sampling.py
Doresic: Good idea
I agree
Paul suggested below some similar additions of allowing the user pass settings to the functions.
I would tackle all of these together in the next PR that'll focus on style of all plots + the possibility of used adding style changes to the functions or even whole module. There could be a very cool module-wide-changes thing one can do and define.
A simple thing could be just allowing the user to pass a dictionary of kwargs that will get distributed acorss all plotting functions to change sizes shapes etc. -- up to some limit, adding all would probably be too much.
And then having a specification that's general across all visualisation functions in _style.py of general sizes/font sizes etc that will be used by default and imported from the _style as constants.

pypesto/visualize/observable_mapping.py
```diff
@@ -41,13 +42,90 @@
     pass
 
 
+def _observable_mapping_grid_shape(n_panels: int) -> tuple[int, int]:
```
dweindl: This isn't really related to observable mappings. Rename and move to general plotting utils?

pypesto/visualize/_style.py
Doresic: Yeah I think all of these functions fit better to misc.py
There will be a purpose to the style file but it's not there yet -- next PR stuff that actually deal with style.
In the sense that it will contain constants of sizes, fonts etc that would be used as style defaults across the vis. module.
Makes more sense yup

pypesto/visualize/optimization_stats.py
dweindl: I wouldn't consider an incorrect type more informative 🤔

pypesto/visualize/observable_mapping.py
Doresic: Yep true, makes sense, we use this at many places. Will add to misc and use it anywhere we want to define a grid with a certain number of plots

pypesto/visualize/dimension_reduction.py
Doresic: I'll put it to matplotlib.axes.Axes just for consistency.

pypesto/visualize/sampling.py
```diff
@@ -1178,18 +1210,19 @@ def sampling_parameter_traces(
     fig.tight_layout()
```
dweindl: If a user is allowed to pass fig/ax, we should remove all calls to `tight_layout()`. The user may have chosen a different layout engine which would be replaced here - that's annoying.

pypesto/visualize/_style.py
Doresic: Hmmm that's true
I didn't think about this, in case a user wanted to pass None, this would not catch it.
Currently I don't think I have examples of where I use this deprecation in which a user would ever want to pass None. But there might be

Should we then in this case
- have a specific _UNSET value that tracks whether something is set or not for variables that are set or not that would then be different from None to catch None's too?
- just have a dosctring acknowledging this limitation for future usage of this function (mostly for us) to make sure the case where we use it is not one of these cases?

test/visualize/test_style.py
```diff
@@ -0,0 +1,49 @@
...
+
+def test_get_ax():
+    """Returns the given Axes; otherwise creates one with ``size``."""
+    _, given = plt.subplots()
+    assert get_ax(given) is given
+
+    custom = get_ax(size=(4.0, 3.0))
+    assert tuple(custom.get_figure().get_size_inches()) == (4.0, 3.0)
+
+    plt.close("all")
```
dweindl: There's `close_fig` to be used as decorator in `test_visualizy.py`. Move that to `conftest.py` so it can also be used here. Replace all direct `plt.close("all")` calls to ensure figures are closed also in case of test failure.

pypesto/visualize/_style.py
dweindl: I'd go for the first option, which is probably even faster...

pypesto/visualize/sampling.py
Doresic: very true, didn't have this in mind
hmmm should this be "no tight layout in case an ax is passed" or in general?

dweindl: No experience with this specific instance, but overall, I think, tight_layout will disappear.
I'd suggest using `plt.subplots(..., layout="constrained")` in `get_ax` et al., and just removing all `tight_layout()` calls.

pypesto/visualize/observable_mapping.py
Doresic: Good point. This needs changes.

Hmm it's kinda different. the get_axes things make axes. This part of the code checks the inputed axes and adjusts them. But we shouldn't do that. We should rather check just whether the axes are what they should be.

I changed it now, this was in some sense a problem of the observable mapping function. It allowed for users to provide very various axes objects. However, the axes object should only ever be a 2D ndarray. So now we check this.

pypesto/visualize/observable_mapping.py
Doresic: True
There are different examples of hiding panels across the module, but main ones are doing exactly thins -- hiding extra created panels we don't really need
So I'll make another function in misc.py that does that `hide_unused_axes(axes, n_used)`

Doresic:
Thank you a lot for all the comments!
They should be addressed now
Feel free to take another look @dweindl @PaulJonasJost @vwiela

dweindl:
> They should be addressed now

Didn't check everything in all details, but my points have been addressed, thanks.

vwiela (approved):
Looks better for me now and with the style and further changes incoming, good for me to merge this.

PaulJonasJost (approved):
Also did not check every detail. Have you considered the pair plot in matplotlib that I sent you? Might be much less code and not require specific code to plot the diagonal.

Doresic:
> Also did not check every detail. Have you considered the pair plot in matplotlib that I sent you? Might be much less code and not require specific code to plot the diagonal.

I've taken a close look at this. It would be possible to use the `scatter_matrix` method. But with the amount of changes we're making afterwards, I don't think it would be a good change. In this case the `scatter_matrix` would replace only a part of the current re-writing. Specifically 23 lines that do

```
for row in range(n):
    for col in range(n):
        if row == col:
            plot_diagonal_marginal(...)
        else:
            ax.scatter(...)
```

All of the rest of the code that I've added are additions to the visualisation itself like
- showing bounds
- coloring scatter points per cmap
- removing spines and unnecessary axes
- better "parameter-amount-aware" sizing
- sharing x- and y-limits per row with padding
and some other things that will be added in the next PR.

So with this in mind, I wouldn't use "scatter_matrix" as it would just introduce an external function we would need to rely on for a simple part of the code we can have there manually allowing easier bug-fixing or changing.

PR #1711, opened 2026-05-07
Profile: Improve step-size defaults and step-size robustness

Doresic:
Absolute profiling step sizes can be too small for parameters with very wide bounds. This PR updates profiling step-size handling so that pyPESTO can choose between absolute and relative step sizes per profiled parameter.

Relative step sizes are computed as fractions of the parameter span `ub - lb` on the optimization scale. pyPESTO then uses either the full absolute or the full relative step-size family, based on the resolved default step size.

Also:
- rename the absolute step-size options to make them explicit
- keep the old option names with deprecation warnings
- allow disabling a step-size family by setting its default step size to `0`
- require finite profiling bounds
- resolve profile step sizes once before profiling and pass them downstream
- add a small precheck warning for setups that may require many profile steps
- adjust the related tests


pypesto/profile/options.py
```diff
@@ -11,10 +11,22 @@ class ProfileOptions(dict):
         Default step size of the profiling routine along the profile path
         (adaptive step lengths algorithms will only use this as a first guess
         and then refine the update).
+    default_step_size_relative:
+        Relative default step size for wide `lin`-scale parameters, expressed
```
dweindl: I think I wouldn't distinguish between lin/log here. If it's always applied to the scaled bounds, I don't see an advantage of treating those differntly.

pypesto/profile/options.py
```diff
@@ -11,10 +11,22 @@ class ProfileOptions(dict):
...
+        as a fraction of the full parameter span `ub - lb`. The effective
+        default step size is the maximum of the absolute and relative values.
     min_step_size:
         Lower bound for the step size in adaptive methods.
```
dweindl: Maybe add an explicit "absolute" here, so that it's clearer what "absolute" in the previous sentence is referring to?

pypesto/profile/options.py
```diff
@@ -38,33 +50,46 @@ class ProfileOptions(dict):
     whole_path:
         Whether to profile the whole bounds or only till we get below the
         ratio.
+    step_size_precheck_mode:
+        Behavior of the profile step-size precheck.
```
dweindl: What *is* the profile step-size precheck? 😅

pypesto/profile/options.py
```diff
@@ -111,6 +136,25 @@ def validate(self):
             raise ValueError("default_step_size must be <= max_step_size.")
         if self.default_step_size < self.min_step_size:
             raise ValueError("default_step_size must be >= min_step_size.")
+        if self.default_step_size_relative <= 0:
+            raise ValueError("default_step_size_relative must be > 0.")
```
dweindl: Why not 0 (i.e., use only absolute)?

pypesto/profile/options.py
```diff
@@ -111,6 +136,25 @@ def validate(self):
...
+        if self.max_step_size_relative <= 0:
+            raise ValueError("max_step_size_relative must be > 0.")
+        if self.min_step_size_relative > self.default_step_size_relative:
+            raise ValueError(
+                "min_step_size_relative must be <= default_step_size_relative."
+            )
+        if self.default_step_size_relative > self.max_step_size_relative:
+            raise ValueError(
+                "default_step_size_relative must be <= max_step_size_relative."
+            )
```
dweindl: I understood that the different step sizes are `effective_stepsize = max(abs_stepsize, rel_stepsize * (ub-lb))`. Is that correct? 
In that case, do those checks in isolation really make sense? Shouldn't this rather be applied to `effective_stepsize`, i.e., at a later stage where the bounds are available?

pypesto/profile/profile_next_guess.py
```diff
@@ -557,16 +579,14 @@ def do_line_search(
         next_x = clip_to_bounds(par_extrapol(step_size_guess))
 
         # Check if we hit the bounds
-        if (
-            direction == "decrease"
-            and step_size_guess
-            == options.min_step_size * min_step_increase_factor
+        if direction == "decrease" and np.isclose(
```
dweindl: Is that safe with `isclose` default tolerances? I think this could become a problem for small parameter values.

pypesto/profile/util.py
```diff
@@ -80,6 +125,173 @@ def calculate_approximate_ci(
...
+    default_step_size = options.default_step_size
+    min_step_size = options.min_step_size
+    max_step_size = options.max_step_size
+    span = None
+    uses_relative_min = False
+    uses_relative_default = False
+    uses_relative_max = False
+
+    scale = str(problem.x_scales[i_par]).lower()
+    if scale == "lin":
```
dweindl: `pypesto.C.LIN`?

pypesto/profile/options.py
Doresic: Yeah sounds good
It the major question of this PR. I was not sure whether I wanted to do this because of this

> I left that out here to avoid introducing a less obvious behavior change for log-scaled parameters -- like in the case when a user really wants to have dense profiles, but the step sizes are just never dense enough as it uses relative step sizes then.

But it does make sense to introduce relative-step-size-choice for `log` and `log10` parameters too.
I'll make the changes

pypesto/profile/options.py
Doresic: Yeah true, I'll change the naming, and make it more explicit.

pypesto/profile/options.py
Doresic: Hahah true, will need to edit to say what it does.
It's basically just checking step sizes and seeing if some of them would cause too many steps taking place. Now this will change if we make it default to swap to relative step sizes in `log` and `log10` scales too

pypesto/profile/options.py
Doresic: Hmmmm that feels a bit of a weird unintended manipulation of step sizes, no?
Should we allow for this?...
I would rather in that case allow for None? To disable it?

dweindl: For me, using 0 to disable it would feel quite natural :man_shrugging: 
`None` just makes it unnecessarily more complex?

pypesto/profile/options.py
dweindl: After reading the respective code, it was quite obvious, but the description here alone left me a bit puzzled.

pypesto/profile/options.py
Doresic: So I would rather do the following logic that makes sure the following is true - We're either using all relative or all absolute step sizes. A mix of both feels a bit weird or feels like it would cause unusual unintended behavior. So that's why these checks are checking only relative against relative and same for absolute. 

So the logic of choosing whether a parameter is using a relative or absolute step sizes would be based on which has a higher step size default. Meaning 
- if resolved_relative_default_step_size >= resolved_absolute_default_step_size --> We use relative step sizes
- we use abosolute step sizes otherwise

I would probably want to circumvent a weird mix of relative and absolute step sizes depending on the chosen step sizes. The information would be easier to deal with imo 
--> warnings of "this par is using relative step sizes" are simpler than "this parameter is using this relative, this absolute etc..." so the effects of managing and changing them are "simpler to understand for the user"?

pypesto/profile/options.py
Doresic: Will describe it for sure ^^

pypesto/profile/options.py
Doresic: Hmmm yeah I see... I would just need to make explicit how it's chosen which are used -- max of both. Then users can conclude what to do to disable.
Yeah, I'll do that but have nice documentation around it

pypesto/profile/profile_next_guess.py
Doresic: Hmmm true I'll take a look at this...

pypesto/profile/options.py
Doresic: Just a note
-- this is different to what I initially suggested in this PR -- so this logic is not present rn.

pypesto/profile/options.py
Doresic: A bit of a weird thing about this -- some ValueError check might jump in this case.
What if the default is set to 0? But then it's lower than the min step size.
Should we
- make the user set the min to 0 too?
- assume the user is trying to turn this whole thing off so we make min and max 0 as well if default is 0?

I would actually vote for the second one and make that explicit in the docstring? That default is the step size that controls if it's turned of by making it 0. Either absolute or relative.

Doresic:
Thanks @dweindl for the comments, I updated the step-size handling quite a bit so here's a short summary:

- relative step sizes are now applied on the optimization scale for all parameter scales
- step-size options are split into explicit absolute/relative names
- old absolute option names are kept with deprecation warnings
- a step-size family can be disabled by setting its default step size to `0`
- pyPESTO now resolves one full step-size family per parameter instead of mixing absolute/relative min/default/max values
- step sizes are resolved once before profile walking and passed downstream
- the precheck docs/message were shortened and made more explicit
- the `np.isclose` bound checks in the line search were replaced by directional `<=` / `>=` checks
- tests were trimmed to cover the new behavior

pypesto/profile/options.py
```diff
@@ -38,38 +60,90 @@ class ProfileOptions(dict):
...
     def __init__(
         self,
-        default_step_size: float = 0.01,
-        min_step_size: float = 0.001,
-        max_step_size: float = 0.1,
+        default_step_size_absolute: float = 0.02,
+        default_step_size_relative: float = 0.01,
+        min_step_size_absolute: float = 0.01,
+        min_step_size_relative: float = 0.005,
+        max_step_size_absolute: float = 0.2,
```
dweindl, line 77: As the goal here is improving robustness, I am wondering if the absolute step sizes shouldn't be much smaller. Or disabled by default?! For a parameter in [1e-4, 1e-3], the current defaults wouldn't be too helpful. Do you see a case where just relying on relative tolerances would be problematic?

pypesto/profile/options.py
```diff
@@ -99,18 +173,45 @@ def validate(self):
...
+            max_step_size = self[f"max_step_size_{family}"]
+
+            if default_step_size < 0:
+                raise ValueError(f"default_step_size_{family} must be >= 0.")
+            if default_step_size == 0:
+                return False
+            if min_step_size <= 0:
+                raise ValueError(f"min_step_size_{family} must be > 0.")
+            if max_step_size <= 0:
+                raise ValueError(f"max_step_size_{family} must be > 0.")
```
dweindl, line 189:
```suggestion
            if min_step_size < 0:
                raise ValueError(f"min_step_size_{family} must be >= 0.")
            if max_step_size < 0:
                raise ValueError(f"max_step_size_{family} must be >= 0.")
```

dweindl: Ah, this was intended. Now I get it.

pypesto/profile/options.py
```diff
@@ -1,20 +1,42 @@
...
-        and then refine the update).
-    min_step_size:
-        Lower bound for the step size in adaptive methods.
-    max_step_size:
-        Upper bound for the step size in adaptive methods.
+    default_step_size_absolute:
+        Default absolute profile step size. Set to `0` to disable.
+    default_step_size_relative:
+        Default relative profile step size, as fraction of `ub - lb`. Set to
+        `0` to disable.
```
PaulJonasJost, line 31: Wouldn't set to `None` be better to disable?

PaulJonasJost: also for other option

PaulJonasJost (approved):
I am not fully clear on the usefulness of the Resolved Step size, i.e. adding data class, multiple checks at multiple points etc, for it to check two simple calculations of "how many steps are expected" and "How many in a worst case?". Have I missed a crucial functionality for them?

pypesto/profile/options.py
Doresic: The only downside I see of using only relative stepping is that it would take unnecessary many steps for very tightly constrained parameters, e.g., if there is a parameter with range 0-2 (some exponent). 

The absolute step sizes I chose here were chosen based on the profiling benchmark -- as a tradeoff between robustness and speed. So I would keep them as such. But you do raise a good point --  actually the relative defaults should be different, better thought through. Generally, the median range of parameters the benchmark spanned 8 orders of magnitude.

So making the relative method kick in when the span exceeds 8 would be appropriate. And then from then on it would make as many steps as it would make for parameters with span 8. That would make the relative defaults be:

```
min_step_size_relative = 0.00125
default_step_size_relative = 0.0025
max_step_size_relative = 0.025
```

So for very small ranges the absolute method would cause the step size to be appropriate and finish the profiling quickly. For ones exceeding 8 orders of magnitude, the relative stepping would kick in which would cause all profiling to have per default step size 400 steps, and per min step size (worse case scenario) 800 steps.

Doresic:
> I am not fully clear on the usefulness of the Resolved Step size, i.e. adding data class, multiple checks at multiple points etc, for it to check two simple calculations of "how many steps are expected" and "How many in a worst case?". Have I missed a crucial functionality for them?

It's more just an implementational thing. It causes for very simple downstream coding as we can call `resolved_steps.min_step_size` and we can pass only one object for all step sizes, instead of many defaults and flags and if statements.

PR #1712, opened 2026-05-07
Profile: add multistart optimization for profile points

Doresic:
I added multistart optimization at each profile point.

Additional starts are sampled around the point suggested by `profile_next_guess`, and the original suggested start is always included as well. Failed starts are tolerated, and the best finite result is kept.

I also added warnings when profiling finds lower function values than the supplied optimum, and when non-profiled parameters hit bounds during profiling.

Also:
- add `profile_n_starts` and `profile_sampling_sigma` to `ProfileOptions`
- use multistart optimization during profile walking
- extend the related profiling tests

Open question: when non-profiled parameters hit bounds during profiling, should we eventually support relaxing those bounds (of the parameters we're not profiling), or is warning-only the better behavior here? Allowing a relaxation of the bounds would make the plots actually show non-identifiable parameters when they are such. But we might risk some weird optimisation failures in some unusual cases.

Should be merged only after #1711 -- contains all changes of that PR currently, too.


pypesto/profile/options.py
```diff
@@ -83,6 +91,8 @@ def __init__(
         reg_order: int = 4,
         adaptive_target_scaling_factor: float = 1.5,
         whole_path: bool = False,
+        profile_n_starts: int = 6,
```
PaulJonasJost, line 94: Wouldn`t keeping the default at one or two be better? In case everything works it is 1/6  (1/3)of the overall work and otherwise depending on how fatally it fails, we can add starts?

pypesto/profile/profile.py
```diff
@@ -210,6 +210,34 @@ def create_next_guess(
...
+        logger.warning(
+            "Profiling found lower function values than the supplied "
+            f"optimization result (fval = {global_opt:.6g}) for:\n"
+            f"{par_lines}\n"
+            "This means profiling was started from a suboptimal point, so the "
+            "profile ratios and confidence thresholds are not anchored to the "
+            "true global optimum. Re-optimization of the model is suggested. "
+            "The better profile points can be included as startpoints "
+            "through `x_guesses` of the optimization problem."
+        )
```
PaulJonasJost, line 239: If I get the message right it is notifying me when I find a point actually better then the best found point from optimization? Like the message, but I think it should also include how to get the best points out if it includes the x_guesses hint

pypesto/profile/walk_along_profile.py
```diff
@@ -14,6 +14,83 @@
...
+    -------
+    The best finite optimization result across all attempted starts, or the
+    result from the original startpoint if all sampled starts fail.
+    """
+    if options.profile_n_starts == 1:
+        return optimizer.minimize(
+            problem=problem,
+            x0=startpoint,
+            id=str(0),
+            optimize_options=OptimizeOptions(allow_failed_starts=True),
```
PaulJonasJost, line 56: why allow failed starts?

PaulJonasJost: Actually in this one I tend to understand (even though it should be impossible. But In 76 it seems more weird.

pypesto/profile/walk_along_profile.py
```diff
@@ -271,6 +348,24 @@ def walk_along_profile(
...
+                warned_at_bound.add(j_par)
+                bound_val = lb_j if at_lb else ub_j
+                logger.warning(
+                    f"Parameter '{problem.x_names[j_par]}' hit its "
+                    f"{'lower' if at_lb else 'upper'} bound "
+                    f"({bound_val:.4g}) while profiling "
+                    f"'{problem.x_names[i_par]}'. "
+                    "The profile may be constrained near this region."
+                )
+
```
PaulJonasJost, line 368: Not sure how informative that actually is, as it should become very apparent in the 2d plots right? Would potentially remove.

test/profile/test_profile.py
```diff
@@ -628,6 +631,70 @@ def test_profile_step_size_precheck_modes(mode, expect_warning, expect_raise):
         assert not precheck_warnings
 
 
+def test_profile_multistart_optimize_uses_best_start(monkeypatch):
```
PaulJonasJost, line 634: what does monkeyxpatch do?

PaulJonasJost (approved):
Regarding open question: I am strongly against loosening bounds. They are there for a reason, even if they are set uninformed. Thus they should not be just changed, especially not mid workflow. Other view on this is that if I loosen the bounds, I do not guarantee at all anymore, that my optimum is actually an optimum, e.g. if in the optimisation we already hit the same bound.

pypesto/profile/options.py
```diff
@@ -60,6 +60,14 @@ class ProfileOptions(dict):
...
         Whether to profile the whole bounds or only till we get below the
         ratio.
+    profile_n_starts:
+        Number of optimization starts attempted at each profile point.
+    profile_sampling_sigma:
+        Standard deviation used for each free coordinate when sampling
+        additional optimization startpoints from a Gaussian centered at the
+        point suggested by `profile_next_guess`, in the reduced parameter
+        space. The profiled parameter itself is already fixed and is therefore
+        not perturbed.
```
dweindl, line 70:
```suggestion
        not perturbed. Only relevant if `profile_n_starts > 1`.
```

Or similar.

pypesto/profile/options.py
```diff
@@ -83,6 +91,8 @@ def __init__(
         reg_order: int = 4,
         adaptive_target_scaling_factor: float = 1.5,
         whole_path: bool = False,
+        profile_n_starts: int = 6,
+        profile_sampling_sigma: float = 0.01,
```
dweindl, line 95: Since this is an absolute sigma, it might be nice to support per-parameter settings, but I guess this can also easily be extended later.

pypesto/profile/profile.py
```diff
@@ -210,6 +210,34 @@ def create_next_guess(
...
+            continue
+        best_fval = float(np.min(profiler_result.fval_path))
+        if best_fval > global_opt or np.isclose(
+            best_fval, global_opt, rtol=1e-10, atol=1e-8
+        ):
+            continue
+        better_optima.append((problem.x_names[i_par], best_fval))
+    if better_optima:
+        par_lines = "\n".join(
+            f"  {name}: best fval = {fval:.6g}" for name, fval in better_optima
```
dweindl, line 228: With the `isclose` thresholds above, this may display identical values, which might be confusing.

pypesto/profile/walk_along_profile.py
dweindl: I'd also stick to False. Maybe make it configurable. 
In my opinion, `Objective.__call__` should never raise when called with admissible parameters. If it does, this is most likely a problem with the Objective and we should not cover that up.

dweindl (commented):
Nice!

pypesto/profile/options.py
Doresic: True
This is the design choice of the default being either
- robust according to the benchmark we ran
- initial non-robust run but faster --> in models with non-identifiability will probably show many jumps.

Having less starts had a large impact on the amount of jumps the profiles showed for the benchmark models I tested.

Based on previous discussions we had in team meetings, it seemed that the default we would like to have would be the robust one -- so all settings are following that -> order 0 (next PR), medium step size (previous PR), 6 multistarts.

Would you vote for the efficient version rather the robust one?

pypesto/profile/walk_along_profile.py
Doresic: True
I put it there with something else in mind that doesn't apply. Will change to False

pypesto/profile/walk_along_profile.py
Doresic: True
Still wanted to add it as Jan mentioned it was in MATLAB previously, and it might just indicate something "might be happening" to users.
I don't think it hurts, as we also send it only once ever per parameter. So logs won't get cluttered with it.

pypesto/profile/options.py
Doresic: Will put 4 for now
And then if there are future robustness problems, we can adjust

PR #1716, opened 2026-05-26
Viz 01: Unify cluster palette and add discrete palette

Doresic:
The cluster-coloured plotters (`waterfall`, `parameters`, `optimization_stats`, `optimizer_history`, `profiles`) all had their cluster palette hardcoded with no user override path. That default sometimes produced indistinguishable cluster colours. As @PaulJonasJost  suggested, a default categorical palette works better. One can change it using a `style_kwargs` override path. I also added a guarantee the palette stays far from the MLE (default red) and isolated point (default grey) colors.

Main changes:

- new `pypesto/visualize/_style.py` with the reserved colours (`MLE_COLOR = tab:red` for the best cluster, `OUTLIER_COLOR` for singletons) and the cycled palette source (`CMAP_DISCRETE`, default `"tab10"`), plus a small `resolve_style` helper
- `_build_cluster_palette` samples non-best cluster colours from `cmap_discrete` with a small distance filter — that way the reserved best/outlier colours stay visually distinct from the cycled ones even when someone passes a cmap that happens to contain red or grey
- `style: dict | None` threaded through `assign_clustered_colors`, `assign_colors`, `assign_colors_for_list`, `process_result_list`
- the 5 top-level plotters get a new `style_kwargs: dict | None = None` kwarg; their lowlevel counterparts (`waterfall_lowlevel`, `parameters_lowlevel`, `optimizer_history_lowlevel`, `stats_lowlevel`) take an already-resolved `style: dict | None = None`. So `resolve_style` runs exactly once per call chain at the user-facing boundary.

This is the first in a small series of visualisation PRs. The rest of `_style.py` (constants for bounds, scatter, colorbars, …) grows incrementally as later PRs polish each visualisation family.

### DIFF
--- a/pypesto/visualize/_style.py
+++ b/pypesto/visualize/_style.py
@@ -1,29 +1,13 @@
 """
-Shared visualization constants and helpers for ``pypesto.visualize``.
+Visual style for ``pypesto.visualize``.
 
-TODO remove too much info after all PRs have been merged.
+Default constants, the ``style_kwargs`` registry, and small cross-module
+helpers.
 
-Grown incrementally across the PR 1.5 follow-on series: each per-viz PR
-adds the constants and helpers its consumers need, in the same diff as
-those consumers. The final PR in the series adds :func:`apply_style`
-(an opt-in rcParams preset).
-
-Style keys are surfaced to users via the ``style_kwargs`` parameter on
-every public plotter, validated against :data:`_DEFAULTS`::
+Users override any default per call via ``style_kwargs``, validated against
+:data:`_DEFAULTS`::
 
     waterfall(result, style_kwargs={"mle_color": "tab:purple"})
-
-How to extend
--------------
-- **Add a constant**: ``UPPER_SNAKE`` name with a 1-line comment on its
-  semantic role, under the appropriate section header. Add a new
-  section header (``# ===`` block) when the purpose is genuinely new.
-- **Add a registry key**: lowercase entry in :data:`_DEFAULTS` referencing
-  the constant. Unknown keys passed to :func:`resolve_style` raise
-  ``UserWarning`` so typos surface immediately.
-- **Add a helper**: under an existing section, or a new one. Helpers that
-  cross module boundaries belong here. Module-local helpers stay in
-  their module.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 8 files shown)


pypesto/visualize/_style.py
```diff
@@ -0,0 +1,62 @@
...
+# ---------
+CMAP_DISCRETE = "tab10"  # qualitative: cluster + per-variable colours
+
+# Style registry
+# --------------
+
+_DEFAULTS: dict[str, object] = {
+    "mle_color": MLE_COLOR,
+    "outlier_color": OUTLIER_COLOR,
+    "cmap_discrete": CMAP_DISCRETE,
```
PaulJonasJost, line 32: some of these seem doubled. I don't know how to feel about this code wise. I would probably say if this is the only place `MLE_COLOR` etc are used, define them immodestly in the default dict. But also just started review.

pypesto/visualize/clust_color.py
```diff
@@ -36,20 +89,25 @@ def assign_clustered_colors(
     # assign clusters
     clusters, cluster_size = assign_clusters(vals)
```
PaulJonasJost, line 90: Not sure whether at other points we also create clusters, but could be a good idea to save the "assign_clusters" somewhere such that they can readily be reused (e.g. I assign clusters based on eval and then color parameter vectors by it). Though not sure this is not already done. might be good to check as assign_clusters takes some time for larger number of starts

PaulJonasJost (approved):
thanks 🙏🏼

pypesto/visualize/_style.py
Doresic: Yeah true
I can only define them in the defaults dict, will do, no need for the duplication
But they are used in many other places tho that will come later on

pypesto/visualize/clust_color.py
Doresic: I think assign_clusters should be quite fast now, it was slow beforehand but Vincent fixed it no?
And now it works quite fast/well and is used across different places like parameter plot, waterfall plot, history plot etc... 
But I will check that too, and whether it makes sense to move the clustering thing somewhere.

PR #1718, opened 2026-05-28
Viz 02: Harmonise 1D distribution plots (histogram + KDE)

Doresic:
It makes sense to make `parameter_hist` and `sampling_1d_marginals` (the two 1D distribution plotters) to behave the same: same bounds handling, same axis labels, same legend layout, same argument vocabulary, as they are very similar plotters.

Main changes:

- both functions now take `show_bounds: bool = True` — when true, the plot frames to the parameter bounds and draws bound lines; when false, it frames tightly to the data. xlim is set explicitly from bounds + actual data, so the KDE tail extension never dominates the visible range
- scale-aware x-labels (log parameters get a log-scale axis label)
- per-panel legend with KDE / Samples / Bounds entries
- rug uses the darker `dash_color` (matching the bar-edge family) so individual data points stay readable on the lighter histogram
- new shared helper `plot_density_panel` in `misc.py` doing the histogram / KDE / rug / bounds rendering — takes a resolved `style` dict
- harmonised the two siblings' arguments: same shared-kwarg order (`plot_type, bins, bw_method, show_bounds, title, size, ax/axes, style_kwargs`), and `sampling_1d_marginals` gained `bins` (was hardcoded). Tidied up some confusing docstrings along the way (`stepsize` is a thinning factor; `start_indices` subsets starts while `parameter_indices` subsets parameters)
- new `apply_style()` in `_style.py` — a small opt-in rcParams preset (top/right spines off, framed legends, constrained layout). General so later PRs in the series just call it
- new generic style keys: `LINE_*` for any line (KDE, simulation, …), `DASH_*` for any tick marker (rug, CI endpoints, …). The naming principle is documented in `_style.py` so future PRs follow it
- `parameter_hist` gains `bw_method` and `title`; its legacy `color` kwarg is deprecated — use `style_kwargs={"rectangle_color": ..., "line_color": ..., "dash_color": ...}` instead
- `process_deprecated_kwarg` in `misc.py` gains a removal mode (`canonical_name=None`) and a `note=` parameter for custom messages — used here for the `color` deprecation; PR 09b will reuse it
- `GRID_SIZE_PER_COL` / `GRID_SIZE_PER_ROW` in `_style.py` wire into `get_axes_array`'s `size=None` default so every multi-panel plotter inherits consistent panel sizing

This is the second in the small series of visualisation PRs. Before/after PNGs below.

### DIFF
--- a/pypesto/visualize/_style.py
+++ b/pypesto/visualize/_style.py
@@ -1,29 +1,13 @@
 """
-Shared visualization constants and helpers for ``pypesto.visualize``.
+Visual style for ``pypesto.visualize``.
 
-TODO remove too much info after all PRs have been merged.
+Default constants, the ``style_kwargs`` registry, and small cross-module
+helpers.
 
-Grown incrementally across the PR 1.5 follow-on series: each per-viz PR
-adds the constants and helpers its consumers need, in the same diff as
-those consumers. The final PR in the series adds :func:`apply_style`
-(an opt-in rcParams preset).
-
-Style keys are surfaced to users via the ``style_kwargs`` parameter on
-every public plotter, validated against :data:`_DEFAULTS`::
+Users override any default per call via ``style_kwargs``, validated against
+:data:`_DEFAULTS`::
 
     waterfall(result, style_kwargs={"mle_color": "tab:purple"})
-
-How to extend
--------------
-- **Add a constant**: ``UPPER_SNAKE`` name with a 1-line comment on its
-  semantic role, under the appropriate section header. Add a new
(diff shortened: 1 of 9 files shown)


Doresic:
Here are the plot comparisons
before and after
Histograms before:
parameter_hist__before
Histograms after:
parameter_hist__after
1d-sampling-marginals before:
sampling_1d_marginals__before
1d-sampling-marginals after no bounds:
sampling_1d_marginals__after_no_bounds
1d-sampling-marginals after with bounds:
sampling_1d_marginals__after

PR #1634, opened 2025-10-29
PEtab v2 import via amici

dweindl:
Support for [PEtab v2](https://petab.readthedocs.io/en/latest/v2/documentation_data_format.html) problems via amici.

**WIP**

PR #1730, opened 2026-07-11
PEtab v2 import via AMICI (finish #1634)

FFroehlich:
This picks up @dweindl's WIP "PEtab v2 import via AMICI" (#1634) and completes it against the current AMICI 1.0 PEtab importer API (`amici.importers.petab` / `amici.sim.sundials.petab`), so PEtab v2 problems can be imported and optimized end-to-end.

### Commits

1. **Import PEtab v2 via AMICI support from #1634** — the WIP work, rebased onto `develop`.
2. **Complete PEtab v2 via AMICI support** — finishes the remaining pieces and fixes several correctness bugs.
3. **Simplify the PEtab v2 calculator and share model/solver with the simulator** — removes the accidental complexity introduced along the way.
4. **Derive `AmiciPetabV2ObjectiveCreator` from `AmiciObjectiveCreator`** — deduplication.

### Objective / calculator (`pypesto/objective/amici/`)

`AmiciCalculatorPetabV2` translates an `amici.sim.sundials.petab.PetabSimulator` result into pyPESTO's return dict:

- use `result.llh` (was an invalid `result["llh"]` subscript);
- fix the Hessian — `result.s2llh` is the (positive semi-definite) FIM, i.e. the Hessian of the *negative* log-likelihood, so it must not be negated;
- implement residual mode (`MODE_RES`): residuals from `result.res`, residual sensitivities from `result.sres`, with the least-squares-safe check for parameter-dependent sigma. Requires an AMICI build with residual-sensitivity aggregation (currently `main`);
- return full-dimension gradients / Hessians / `sres`, as `calculate_function_values` does, rather than pre-reduced arrays that happened to pass through `FixedParametersProcessor.postprocess`;
- log the individual simulations and return the error output for non-finite objective values, matching the v1 path;
- raise a clear error when a parameter is free in the pyPESTO problem but not estimated in the PEtab problem (it is a constant in the AMICI model, so no sensitivities exist — only `sensi_orders=(0,)` works).

`AmiciPetabV2Objective` now hands its own model and solver to the PEtab simulator instead of operating on clones the simulation never sees. Consequences:

- the reporting mode chosen by `AmiciObjective.call_unprocessed` takes effect, so the calculator no longer has to request `RDataReporting.full` on every call — residuals and the FIM are computed only when requested;
- `objective.amici_model` / `objective.amici_solver` are the instances actually simulated with (e.g. `set_add_sigma_residuals`);
- `n_threads` reaches the simulator (it was silently 1);
- `guess_steadystate` is disabled — the simulator builds fresh `ExpData` objects per call, so guesses stored on `self.edatas` could never be applied.

Fixed parameters no longer need separate plumbing. The vector reaching `call_unprocessed` already carries the pyPESTO-fixed values; they were only stripped again by the v1-specific bookkeeping in `AmiciObjective.update_from_problem`, which is now bypassed (there is no parameter mapping to update for v2). This affects profile likelihoods and any non-nominal `fix_parameters`.

Also: `check_gradients_match_finite_differences` no longer uses `x_nominal_scaled`, which does not exist for PEtab v2.

### Objective creator (`pypesto/petab/objective_creator.py`)

`AmiciPetabV2ObjectiveCreator` now derives from `AmiciObjectiveCreator` (−147 lines). Model import and compilation do not depend on the PEtab version; only `compile_model`, `_create_amici_importer`, `create_edatas`, `create_objective` and the `rdatas_to_*` conversions are version-specific. `create_edatas` builds the `ExperimentManager` on the PEtab problem preprocessed by the AMICI importer — only that one matches the compiled model. Stale `amici.Model` / `amici.Solver` annotations updated to `amici.sim.sundials`.

### Importer / startpoints (`pypesto/petab/`)

- `_create_prior_v2`: index priors by the parameter's position in `x_ids` (was the running prior count, which mis-assigned priors unless they were on the leading estimated parameters).
- `PetabStartpoints` for v2: sample from the parameter's own prior distribution (`Parameter.prior_dist`), falling back to a uniform over the bounds — the v1 `sample_from_prior` uses different distribution names (e.g. `logNormal` vs. `log-normal`) and crashed for `log-normal` / `log-laplace` / `log-uniform` initialization priors.

### Why a separate calculator

`AmiciCalculator` is built around PEtab v1's `fill_in_parameters` and `calculate_function_values(parameter_mapping=…)`. PEtab v2 has no equivalent `ParameterMapping`: parameter application (`ExperimentManager.apply_parameters`) and the aggregation of sensitivities across experiments (`PetabSimulator._aggregate_sllh` / `_s2llh` / `_sres`, including placeholder summation) live in AMICI. Reusing the v1 calculator would mean reimplementing that mapping in pyPESTO. `AmiciCalculatorPetabV2` is the fifth `AmiciCalculator` subclass to override `__call__` wholesale, alongside `InnerCalculatorCollector`, `RelativeAmiciCalculator`, `OrdinalCalculator` and `SemiquantCalculator`.

### Not yet supported

Hierarchical optimization, non-quantitative data types, predictors and the prediction-to-PEtab-dataframe conversions raise `NotImplementedError` for v2 problems.

Known limitation: sensitivities are computed for every parameter estimated in the PEtab problem, including those fixed in the pyPESTO problem. Unlike for v1, `plist` cannot be narrowed from pyPESTO — it is set inside `ExperimentManager.apply_parameters`. This costs performance for profile likelihoods.

### Validation

Against AMICI `main` and libpetab 0.8.2:

- the PEtab v2 import tests (Boehm, Schwen) and the SBML and PySB PEtab test-suite v2 cases pass, unsupported cases skipped;
- no regression in the v1 path;
- Boehm checked in detail — nominal function value `138.2219980`, gradient against relative-step finite differences at tight solver tolerances (max rel. err. 3.3e-4), FIM Hessian positive semi-definite with the correct shape, residuals and residual sensitivities against finite differences (rel. 5e-5), fixing a parameter to a non-nominal value, and `deepcopy` / `pickle` preserving the model↔simulator sharing.

Co-authored with @dweindl (see commit trailers). Supersedes the WIP in #1634.

### DIFF
--- a/pypesto/objective/amici/amici.py
+++ b/pypesto/objective/amici/amici.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Union
 
 import numpy as np
+import pandas as pd
 
 from ...C import (
     FVAL,
@@ -36,6 +37,8 @@ from .amici_util import (
 )
 
 if TYPE_CHECKING:
+    import amici.importers.petab
+
     from ...hierarchical import InnerCalculatorCollector
 
 try:
@@ -718,3 +721,67 @@ class AmiciObjective(ObjectiveBase):
             ) in condition_mapping.map_preeq_fix.items():
                 if (val := id_to_val.get(mapped_to_par)) is not None:
                     condition_mapping.map_preeq_fix[model_par] = val
+
+
+class AmiciPetabV2Objective(AmiciObjective):
+    """An AMICI objective constructed from a PEtab v2 problem."""
+
+    def __init__(
+        self,
+        petab_importer: amici.importers.petab.PetabImporter,
+        **kwargs,
+    ) -> None:
+        from .amici_calculator import AmiciCalculatorPetabV2
+
+        self._petab_simulator: amici.petab.petab_importer.PetabSimulator = (
+            petab_importer.create_simulator()
+        )
+        self.petab_problem = petab_importer.petab_problem
+        amici_model = self._petab_simulator.model
+        amici_solver = self._petab_simulator.solver
+        edatas = self._petab_simulator.exp_man.create_edatas()
+
+        super().__init__(
+            amici_model=amici_model,
+            amici_solver=amici_solver,
+            edatas=edatas,
+            calculator=AmiciCalculatorPetabV2(self._petab_simulator),
+            **kwargs,
+        )
+
+    def __deepcopy__(self, memo=None):
+        """Override AmiciObjective.__deepcopy__."""
+        if memo is None:
+            memo = {}
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            setattr(result, k, copy.deepcopy(v, memo))
+        return result
+
+    def __getstate__(self) -> dict:
+        """Use Python's default pickling semantics (shallow copy of instance dict)."""
+        return dict(self.__dict__)
+
+    def __setstate__(self, state: dict) -> None:
+        """Restore state using the instance dict (default unpickling behaviour)."""
+        self.__dict__.update(state)
+
+    def rdatas_to_simulation_df(
+        self,
+        rdatas: Sequence[amici.ReturnData],
+    ) -> pd.DataFrame:
+        """
+        See :meth:`rdatas_to_measurement_df`.
+
+        Except a petab simulation dataframe is created, i.e. the measurement
+        column label is adjusted.
+        """
+        from amici.importers.petab import rdatas_to_simulation_df
+
+        return rdatas_to_simulation_df(
+            rdatas,
+            self._petab_simulator._model,
+            self._petab_simulator._petab_problem,
+        )
--- a/pypesto/objective/amici/amici.py
+++ b/pypesto/objective/amici/amici.py
@@ -725,7 +725,33 @@ class AmiciObjective(ObjectiveBase):
 
 
 class AmiciPetabV2Objective(AmiciObjective):
-    """An AMICI objective constructed from a PEtab v2 problem."""
+    """An AMICI objective constructed from a PEtab v2 problem.
+
+    The objective value is the negative log-likelihood, or the unnormalized
+    negative log-posterior if the PEtab problem defines parameter priors (see
+    :meth:`PetabImporter.create_problem`, which adds the prior terms).
+
+    .. warning::
+
+        PEtab v2 support is still under development, and this interface may
+        change while the following limitations are addressed:
+
+        * Sensitivities are computed for every parameter that is estimated in
+          the PEtab problem, including those fixed in the
+          :class:`pypesto.Problem`. Unlike for PEtab v1, ``plist`` cannot be
+          narrowed down from pyPESTO, since it is set by
+          :meth:`amici.sim.sundials.petab.ExperimentManager.apply_parameters`.
+          This costs performance, in particular for profile likelihoods.
+        * Steady-state guessing (``guess_steadystate``) is not available: the
+          PEtab simulator creates its own :class:`amici.ExpData` objects for
+          every simulation, so guesses cannot be passed on to it.
+        * Hierarchical optimization and non-quantitative data types (ordinal,
+          censored, semiquantitative) are not supported.
+        * Predictors (:class:`pypesto.predict.AmiciPredictor`) and the
+          conversion of predictions to PEtab dataframes are not supported.
+        * Custom timepoints (:meth:`AmiciObjective.set_custom_timepoints`)
+          have no effect, for the same reason as steady-state guessing.
+    """
 
     def __init__(
         self,
--- a/pypesto/objective/amici/amici.py
+++ b/pypesto/objective/amici/amici.py
@@ -751,17 +751,27 @@ class AmiciPetabV2Objective(AmiciObjective):
             petab_importer.create_simulator(force_import=force_compile)
         )
         self.petab_problem = petab_importer.petab_problem
-        amici_model = self._petab_simulator.model
-        amici_solver = self._petab_simulator.solver
-        edatas = self._petab_simulator.exp_man.create_edatas()
+
+        # the simulator creates its own ExpData objects for every simulation,
+        #  so steady-state guesses cannot be passed on to it
+        kwargs.setdefault("guess_steadystate", False)
 
         super().__init__(
-            amici_model=amici_model,
-            amici_solver=amici_solver,
-            edatas=edatas,
+            amici_model=self._petab_simulator.model,
+            amici_solver=self._petab_simulator.solver,
+            edatas=self._petab_simulator.exp_man.create_edatas(),
             calculator=AmiciCalculatorPetabV2(self._petab_simulator),
             **kwargs,
         )
+        # `AmiciObjective` works on clones of the model and the solver, but the
+        #  simulation is run by the simulator. Discard the clones, so that
+        #  settings applied to `self.amici_model` / `self.amici_solver`
+        #  (sensitivity order, reporting mode, sigma residuals, ...) take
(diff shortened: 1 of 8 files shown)


FFroehlich:
## Benchmark collection sweep (PEtab v2)

Ran the whole [Benchmark-Models-PEtab](https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab) collection through this branch, against AMICI `main`. All 35 problems load and upgrade to PEtab v2 without error, so nothing below is an upgrade *failure*.

Five models were skipped up front — AMICI compilation dominates and the point here is coverage of the v2 code path, not benchmarking: `Froehlich_CellSystems2018` (2686 reactions), `Chen_MSB2009` (827), `Lang_PLOSComputBiol2024` (669), `SalazarCavazos_MBoC2020` (618), `Smith_BMCSystBiol2013` (367).

Of the remaining 30, **27 completed** import → compile → objective → gradient → optimization. Three failed at import, none inside pyPESTO:

| model | raised in | cause |
| --- | --- | --- |
| `Fiedler_BMCSystBiol2016` | `amici` `_check_placeholders` | timepoint-specific observable overrides; needs `flatten_timepoint_specific_output_overrides`. **Not a v2 regression** — the v1 path rejects it too |
| `Liu_IFACPapersOnLine2025` | `amici` `_v2.py:get_k` | `KeyError: '_petab_experiment_indicator_experiment____ctrl'` on a converter-generated id |
| `Raimundez_PCB2020` | `petab/v2/converters.py` → `sbmlmath` → `lxml` | `XMLSyntaxError: Excessive depth in document: 256` |

## v1 vs. v2 objective agreement

The likelihood at the nominal parameters does not depend on the parameter scale, so importing a problem as PEtab v1 or as upgraded PEtab v2 has to give the same objective value. Comparing all 30:

**17 agree** to integrator tolerance (relative differences `2e-16` … `7e-7`). Every one of them uses `observableTransformation = lin` throughout.

**9 disagreed**, and the split was exact: 8 of them were precisely the models with a non-linear `observableTransformation`, which libpetab 0.8.2 dropped during the upgrade. With libpetab `main` (`36467063`) that is fixed for `log` — `Blasi_CellSystems2016` and `Laske_PLOSComputBiol2019` now agree to `5e-15` and `2.9e-12`. `log10` is still mapped to `LOG_NORMAL`, i.e. a natural-log distribution, so `Perelson`, `Borghans`, `Elowitz`, `Lucarelli`, `Bachmann` and `Schwen` still differ; that is tracked upstream.

**The 9th was ours**, and is fixed in 4c252fd. `Zhao_QuantBiol2020` has no objective priors in v1, but the upgrade synthesises a `uniform` prior for every parameter, and `_create_prior_v2` added all of them to the objective. The gap was exactly `sum(log(ub - lb)) = 112.83` (`777.34` v1 vs. `890.17` v2) — a constant, since the bounds are enforced by the pyPESTO problem anyway. The v1 path skips `parameterScaleUniform` for this reason; v2 now does the same, and the two agree to `6e-9`.

One further discrepancy, upstream: `Isensee_JCB2018` gives `3953.82` on v1 and `inf` on v2, with 122 of 123 experiments failing as `CVode returned a root after reinitialization at t=55` — AMICI's v2 period/event handling.

## Note for reviewers

pyPESTO's CI installs the benchmark collection in the `base`, `petab` and `hierarchical` tox environments, but no test iterates over it — only five models are referenced by name (`Boehm`, `Brannmark`, `Bruno`, `Weber`, `Zheng`), and nothing compares v1 against v2. None of the above would have been caught by CI.

A parametrized v1-vs-v2 nominal-objective test over five `lin` models (`Boehm`, `Armistead`, `Fujita`, `Brannmark`, `Zhao`) passes on this branch in ~3.5 min, but it needs both the v1 and v2 model compiled, so it is not included here — it would want its own opt-in environment rather than going into `base`.

pypesto/objective/amici/amici.py
```diff
@@ -718,3 +722,164 @@ def update_from_problem(
             ) in condition_mapping.map_preeq_fix.items():
                 if (val := id_to_val.get(mapped_to_par)) is not None:
                     condition_mapping.map_preeq_fix[model_par] = val
+
+
+class AmiciPetabV2Objective(AmiciObjective):
+    """An AMICI objective constructed from a PEtab v2 problem."""
```
dweindl: > Known limitation: sensitivities are computed for every parameter estimated in the PEtab problem, including those fixed in the pyPESTO problem. Unlike for v1, plist cannot be narrowed from pyPESTO — it is set inside ExperimentManager.apply_parameters. This costs performance for profile likelihoods.

If this is still true, please add a note here. As well as any additional noteworthy limitations.

pypesto/petab/importer.py
```diff
@@ -277,6 +305,57 @@ def create_prior(self) -> NegLogParameterPriors | None:
...
+
+            if str(parameter.prior_distribution) == petab_c.UNIFORM:
+                # A uniform prior only adds a constant to the objective inside
+                #  the parameter bounds, which are enforced by the pyPESTO
+                #  problem anyway. Skipping it keeps the objective comparable
+                #  to the PEtab v1 path, which skips `parameterScaleUniform`
+                #  for the same reason. Upgraded v1 problems can carry such
+                #  priors for every parameter, which would otherwise offset
+                #  the objective by `sum(log(ub - lb))`.
+                continue
```
dweindl: I wasn't aware that this was the case for v1, and I would be in favor including all prior terms (at least) for v2. In the PEtab context, I don't think we need to worry about its cost.

This is also what we encourage in the PEtab v2 specs:

> Different tools may use different objective function formulations internally, as long as they preserve the optima. However, tools should provide the final objective function value, as the negative log-likelihood or unnormalized negative log-posterior, to facilitate comparison and reproducibility.

The Objective subclass should then clearly document what the objective value represents.

test/petab/test_petab_suite.py
```diff
@@ -69,29 +68,33 @@ def _execute_case(case, model_type, version):
     model_name = (
         f"petab_test_case_{case}_{model_type}_{version.replace('.', '_')}"
     )
-    output_folder = f"amici_models/{model_name}"
 
     # import and create objective function
     if case.startswith("0006"):
+        if version == "v2.0.0":
+            pytest.skip("TODO")
```
dweindl: Remove this `skip`. 

0006:
> This case tests support for time-point specific overrides in the measurement
table.

Either this should raise a NotImplementedError which should be caught and skipped outside `_execute_case`, or it should preprocess the Problem with `amici.importers.petab._petab_importer.flatten_timepoint_specific_output_overrides`.

dweindl (commented):
Thanks for pushing this forward. A couple of comments above. I'd prefer adding a note that the whole PEtab v2 interface might still be subject to changes while we address the current limitations.

test/petab/test_petab_suite.py
FFroehlich: removed skip and added timepoint flattening

pypesto/petab/importer.py
FFroehlich: now also includes priors & note in the class description

pypesto/objective/amici/amici.py
FFroehlich: added

dweindl (approved):
Great! Thanks!

pypesto/petab/importer.py
```diff
@@ -20,6 +20,8 @@
 except ImportError:
     roadrunner = None
 
+from petab import v2
```
PaulJonasJost, line 23: If we support v2, should we start importing petab.v1 as v1 as well?

PaulJonasJost (approved):
Thanks for this. Only comment I had is on streamline of the petab import.

pypesto/petab/importer.py
dweindl: I wouldn't touch the existing code now, but any new code should make it clear whether it targets petab v1 or v2.

PR #1733, opened 2026-08-04
CI: widen pytest terminal width so failure summaries aren't truncated

dweindl:
## Summary
- GitHub Actions runners have no tty, so pytest falls back to an 80-column width and truncates failure lines in the "short test summary info" section (e.g. `FAILED ... - AssertionError: ...` gets cut to `As...`).
- Set `COLUMNS: '200'` in the workflow `env` and add `COLUMNS` to tox's `passenv` so the wider width actually reaches the pytest subprocess.

Verified locally that pytest honors `COLUMNS` for this truncation, and that tox forwards it through to the pytest subprocess given this `passenv` entry.

## Test plan
- [ ] CI runs green on this PR
- [ ] A failing-test run (or an intentionally long assertion message) shows the full failure text in the summary instead of being cut off at ~80 chars


PaulJonasJost (approved):
thanks 👍🏼

PR #1735, opened 2026-08-10
CI: Use uv (via tox-uv) for tox virtual environments

dweindl:
## Summary
- Switch tox to create/manage virtual environments and install dependencies via [uv](https://docs.astral.sh/uv/) through the [tox-uv](https://github.com/tox-dev/tox-uv) plugin, instead of virtualenv/pip.
- `tox.ini`: auto-provision `tox-uv` (`requires`), and `uv_seed = true` so `pip` stays available inside venvs for the environments (e.g. `petab`) that call `pip install`/`pip uninstall` directly.
- CI (`install_deps.sh`, `ci.yml`) installs `tox-uv` alongside `tox` and caches `~/.cache/uv` next to the existing pip/`.tox` caches.
- `requirements-dev.txt`/`CONTRIBUTE.rst` updated for local contributor setup.
- No test/build behavior changes — only how environments are created.

Measured locally (repo's `windows` tox env, warm caches): env setup ~7.5x faster (36.0s → 4.8s), overall tox run ~1.9x faster (49.0s → 26.0s). Repeat/warm-cache package installs measured up to ~100x faster wall time than pip.

## Test plan
- [x] `tox -e clean` runs successfully with the auto-provisioned `tox-uv` plugin
- [x] `tox -e windows` passes (99 tests) with the new uv-based runner, matching legacy virtualenv+pip runner results
- [ ] CI green across all jobs (base, mac, windows, petab, julia, optimize, hierarchical, select, quality, docs, notebooks1/2)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,7 @@ jobs:
       with:
         path: |
           ~/.cache/pip
+          ~/.cache/uv
           .tox/
         key: "${{ runner.os }}-${{ runner.arch }}-py${{ matrix.python-version }}-ci-${{ github.job }}"
 
@@ -80,6 +81,7 @@ jobs:
       with:
         path: |
           ~/.cache/pip
+          ~/.cache/uv
           .tox/
         key: "${{ runner.os }}-${{ runner.arch }}-py${{ matrix.python-version }}-ci-${{ github.job }}"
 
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -41,6 +41,7 @@ jobs:
       with:
         path: |
           ~/.cache/pip
```
PaulJonasJost, line 43: If we change to uv, do we need pip, I though uv also handles that?

PaulJonasJost (approved):
Thanks, happy to change if it is faster 👍🏼

.github/workflows/ci.yml
dweindl: One thing is that we still install tox and tox-uv in the outer Python environment via pip, and then inside some tox environments, we still install a couple of packages via pip. I think they can be replaced eventually, but I didn't want to touch that for now. The current changes already give a noticeable improvement at low cost.

PR #1737, opened 2026-09-02
Fix Hessian check in check_gradients_match_finite_differences

FFroehlich:
## What

`ObjectiveBase.check_gradients_match_finite_differences` accepted an `order` argument (documented as "0 for gradient, 1 for hessian") but never forwarded it, so `order=1` silently re-ran the order-0 gradient check. Forwarding it alone was not enough — `check_grad` had the same class of bug underneath.

Three changes in `pypesto/objective/base.py`:

1. **`check_grad`**: the perturbed points now evaluate sensitivity order `(0 + order,)` instead of always `(0,)`. Since 074078e the base point has been evaluated at `(0 + order, 1 + order)` while the perturbed points stayed at `(0,)`, so an `order=1` check compared Hessian columns against finite differences of the *function value* rather than the gradient. For a `normal(1, 1)` prior at x = 0.5 it compared `hess = 1.0` against `fd_c = -0.5`.
2. **`check_gradients_match_finite_differences`**: forwards `order=order` to `check_grad_multi_eps`.
3. Same function: with `mode=None` and `order=1`, only `MODE_FUN` is checked. There are no second-order residual sensitivities, so `MODE_RES` would raise, get swallowed by the existing `try/except`, and be reported as a derivative mismatch.

## Why

`test/base/test_prior.py::test_derivatives` asserts the `order=1` check passes, believing it validates the priors' second derivatives. It did not — it re-ran the gradient check. The Hessian check now genuinely runs.

## The priors were fine

Forwarding `order` initially failed 23 `test_derivatives` parametrizations, which looked like broken prior second derivatives. It wasn't. Probing all 21 prior-type x scale combinations directly — analytic `density_ddx` against central differences of `density_dx`, bypassing `check_grad` entirely — the worst deviation is 4e-5 (logNormal/lin). The chain-rule transforms for log and log10 scales are correct. The failures came entirely from the broken `order=1` comparison.

## Verification

- `test/base/test_prior.py`: 90 passed, with the Hessian check actually running.
- `test/base/test_objective.py` (the direct `check_grad` / `check_grad_multi_eps` consumers, all `order=0`): 62 passed, 8 skipped.
- The check now has teeth: a prior with `density_ddx` corrupted to -5 instead of -1 is rejected at `order=1` while still passing at `order=0`.
- `order=0` — every existing caller in the repo — is byte-identical. No caller passes `order` explicitly.

## For the reviewer

Making `order=1` real changes behavior for two categories of objective that previously got a free pass, and neither is covered by the test suite:

- **Objectives without Hessian support** (adjoint AMICI, JAX, `PetabSimulatorObjective`). `order=1` now requests sensi orders `(1, 2)`, `__call__` raises `ValueError`, and the blanket `except (RuntimeError, ValueError): return False` reports it as "derivatives don't match" rather than "unsupported". Previously such a call silently ran an order-0 check and returned `True`.
- **`AmiciObjective` with the default `fim_for_hess=True`**, where sensi order 2 in `MODE_FUN` returns the Fisher information matrix (`rdata["FIM"]`), not the true Hessian. An `order=1` check compares finite differences of the gradient against the FIM, which agree only at a residual-free fit.

Neither is introduced by this PR — both were latent behind the dropped argument — but `order=1` should not be adopted for those objectives until the swallow at `base.py:714` distinguishes "unsupported" from "mismatched". Happy to address that here if preferred.

Also worth knowing: with `order=1` the per-index errors are whole Hessian columns collapsed with `np.mean` before the tolerance test, so a single wrong entry is diluted by the correct ones. At 400 parameters a Hessian entry that is 200% wrong still passes; at 60 it is caught. Fine for the diagonal prior Hessians here, weaker than it looks for large models.

## Base branch

Targeting `develop`, matching every recent merged PR. `main` is 26 commits behind and would have produced a 27-commit diff.

### DIFF
--- a/pypesto/objective/base.py
+++ b/pypesto/objective/base.py
@@ -552,12 +552,12 @@ class ObjectiveBase(ABC):
             # forward (plus) point
             x_p = copy.deepcopy(x)
             x_p[ix] += eps
-            fval_p = self(x_p, (0,), mode)
+            fval_p = self(x_p, (0 + order,), mode)
 
             # backward (minus) point
             x_m = copy.deepcopy(x)
             x_m[ix] -= eps
-            fval_m = self(x_m, (0,), mode)
+            fval_m = self(x_m, (0 + order,), mode)
 
             # finite differences
             fd_f_ix = (fval_p - fval) / eps
@@ -690,7 +690,9 @@ class ObjectiveBase(ABC):
         dfs = []
 
         if mode is None:
-            modes = [MODE_FUN, MODE_RES]
+            # second order derivatives (order == 1) are only defined for
+            # MODE_FUN (there are no second order residual sensitivities)
+            modes = [MODE_FUN] if order == 1 else [MODE_FUN, MODE_RES]
         else:
             modes = [mode]
 
@@ -705,6 +707,7 @@ class ObjectiveBase(ABC):
                         *args,
                         **kwargs,
                         mode=mode,
+                        order=order,
                         multi_eps=multi_eps,
                     )
                 )
--- a/pypesto/objective/base.py
+++ b/pypesto/objective/base.py
@@ -493,17 +493,20 @@ class ObjectiveBase(ABC):
         detailed: bool = False,
     ) -> pd.DataFrame:
         """
-        Compare gradient evaluation.
+        Compare an analytic derivative against finite differences.
 
-        Firstly approximate via finite differences, and secondly use the
-        objective gradient.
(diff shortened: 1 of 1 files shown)


pypesto/objective/base.py
PaulJonasJost: I feel this is specifically under the clarification of the bug misleading. Should probably be adjusted appropriately as well.

PaulJonasJost (approved):

PR #1703, opened 2026-03-19
Update to adhere to NEP29

PaulJonasJost:
Updated to python3.12. Due date: 26.04

PR #1738, opened 2026-09-02
Fix UMAP and pymc/ArviZ import failures in the base test env

FFroehlich:
Fixes two import failures in the `base` test environment. Both were masked for a long time by a stale cached `.tox/` env (see *Root cause* below).

## 1. `umap.plot` unimportable — the four UMAP test failures

```
FAILED test/base/test_ensemble.py::test_umap_representation_parameters
FAILED test/base/test_ensemble.py::test_umap_representation_predictions
FAILED test/visualize/test_visualize.py::test_projection_scatter_umap_parameters
FAILED test/visualize/test_visualize.py::test_projection_scatter_umap_predictions
ImportError: umap.plot requires pandas matplotlib datashader bokeh holoviews scikit-image and colorcet to be installed
```

The message is misleading — all of those packages *are* installed. The real error, reproduced locally on Python 3.14:

```
holoviews -> panel 0.10.3 -> from bokeh.models import Box
ImportError: cannot import name 'Box' from 'bokeh.models'
```

`umap-learn[plot]` requires `bokeh` unpinned, so the resolver picks bokeh 3.10.0 and then satisfies `holoviews` by backtracking to holoviews 1.14.9 / panel 0.10.3 — old enough that their bokeh requirement is loose enough to "fit", but whose code does not work with bokeh 3.x. `umap/plot.py` wraps its whole import block in a bare `except ImportError` and re-raises the "requires pandas matplotlib datashader ..." message, hiding the cause.

- `pypesto/ensemble/dimension_reduction.py` — drop `import umap.plot` from `_get_umap_representation_lowlevel`. It was never used there; the low-level UMAP fit has no business requiring the plotting stack. This alone fixes all four failures.
- `pyproject.toml` — add `holoviews>=1.18` to the `mltools` extra. Without it, `projection_scatter_umap_original` (the one function that really does need `umap.plot`) stays broken for users. Its test mocks `umap.plot`, so CI would have gone green while the shipped feature was dead.

## 2. `arviz.concat` gone — pymc 5.x paired with ArviZ 1.x

```
.tox/base/lib/python3.14/site-packages/pymc/backends/arviz.py:31: in <module>
    from arviz import InferenceData, concat, rcParams
E   ImportError: cannot import name 'concat' from 'arviz'
```

ArviZ 1.x is a rewrite that dropped `arviz.concat`. Only pymc>=6 supports it (pymc 6.3.1 requires `arviz>=1.1.0,<2.0`), while pymc<6 declares *no* upper bound on ArviZ — so a resolver may pair pymc 5.x with the `arviz>=1.1.0` we require on Python >=3.12. The CI env did exactly that: pymc 5.28.0 + arviz 1.3.0.

- `pyproject.toml` — bound pymc alongside ArviZ so the broken pairing is unrepresentable: `pymc<6` with ArviZ 0.x on Python <3.12, `pymc>=6` with ArviZ 1.x above.

`_get_posterior_dataset` in `pypesto/sample/pymc.py` already handles both ArviZ generations, so no source change was needed — but note that **pymc 6 has never actually run in CI** (the env was stuck on 5.28), so this bump exercises it for the first time. I verified it locally, see below.

## Root cause behind both

The `.tox/` directory is cached under a key with no dependency hash:

```
key: "${{ runner.os }}-${{ runner.arch }}-py${{ matrix.python-version }}-ci-${{ github.job }}"
```

With a constant key the cache is written once and never refreshed, so each run installs incrementally into a long-frozen environment and only upgrades what is strictly forced. That is why the umap breakage looked mac-specific (different jobs froze different dependency sets), and why the pymc/ArviZ mismatch surfaced the moment a `pyproject.toml` edit triggered a reinstall pass. A **fresh** resolve of the `base` extras on Python 3.12 and 3.14 gives a fully working set: bokeh 3.9.2 / panel 1.9.4 / holoviews 1.23.2 / pymc 6.3.1 / arviz 1.3.0 / pytensor 3.3.0.

Both fixes here are correct independently of the cache, and the pymc bound makes the environment self-heal rather than silently keeping the broken pin. But I have deliberately **not** touched the cache configuration — adding `hashFiles('pyproject.toml')` to the `.tox` cache key (or splitting the `~/.cache/uv` download cache from the `.tox` env cache) would fix the systemic issue, and that is a maintainer policy call affecting all ~10 jobs and their runtimes. Worth doing separately, otherwise CI will keep testing months-old dependency sets.

## Verification

- In a fresh Python 3.14 venv where `umap.plot` is deliberately unimportable, `get_umap_representation_parameters` + `visualize.projection_scatter_umap` both succeed (previously the `ImportError` above).
- With `holoviews>=1.18`, resolving the full `base` extras set for Python 3.14 yields bokeh 3.9.2 / panel 1.9.4 / holoviews 1.23.2, and `import umap.plot` succeeds.
- Resolving the `base` extras set with the new pymc bound (linux, Python 3.12 and 3.14) yields pymc 6.3.1 + arviz 1.3.0 + pytensor 3.3.0. Python 3.11 is unchanged: arviz 0.23.4 + pymc 5.28.5.
- `PymcSampler` end-to-end on pymc 6.3.1 / arviz 1.3.0: NUTS sampling of a 2-D Rosenbrock, correct `trace_x` / `trace_neglogpost` shapes via the ArviZ-1.x `_get_posterior_dataset` path, followed by `geweke_test`. All pass.

## Note for reviewers

I did not run the affected test modules themselves locally — `test/base/test_ensemble.py` and `test/sample` pull in `create_petab_problem` / a full AMICI toolchain. The CI run on this PR is the check for that. Because of the cache behaviour described above, if the job still reports the old pins in its `pip list`, its cache needs clearing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,10 +106,16 @@ mpi = [
 ]
 
 pymc = [
-  # TODO: once Python 3.11 support is dropped, require only ArviZ >=1.1.0.
+  # ArviZ 1.x is a rewrite that, among others, dropped `arviz.concat`. Only
+  #  pymc>=6 supports it; pymc<6 declares no upper bound on ArviZ, so without
+  #  the bounds below a resolver pairs pymc 5.x with ArviZ 1.x, which fails on
+  #  `import pymc`.
+  # TODO: once Python 3.11 support is dropped, require only ArviZ >=1.1.0 and
+  #  pymc >=6.
   "arviz>=0.12.1,<1.0; python_version < '3.12'",
   "arviz>=1.1.0; python_version >= '3.12'",
-  "pymc>=4.2.1",
+  "pymc>=4.2.1,<6; python_version < '3.12'",
+  "pymc>=6; python_version >= '3.12'",
 ]
 
 jax = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,6 +127,11 @@ dynesty = [
 
 mltools = [
   "umap-learn[plot]>=0.5.3",
+  # `umap-learn[plot]` requires an unpinned bokeh; without a lower bound here,
+  #  resolvers satisfy `holoviews` with an ancient release whose `panel`
+  #  version is incompatible with recent bokeh, which makes `umap.plot`
+  #  unimportable.
+  "holoviews>=1.18",
   "scikit-learn>=0.24.1",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -106,16 +106,7 @@ mpi = [
 ]
 
 pymc = [
-  # ArviZ 1.x is a rewrite that, among others, dropped `arviz.concat`. Only
-  #  pymc>=6 supports it; pymc<6 declares no upper bound on ArviZ, so without
-  #  the bounds below a resolver pairs pymc 5.x with ArviZ 1.x, which fails on
-  #  `import pymc`.
-  # TODO: once Python 3.11 support is dropped, require only ArviZ >=1.1.0 and
-  #  pymc >=6.
-  "arviz>=0.12.1,<1.0; python_version < '3.12'",
-  "arviz>=1.1.0; python_version >= '3.12'",
-  "pymc>=4.2.1,<6; python_version < '3.12'",
-  "pymc>=6; python_version >= '3.12'",
+  "pymc>=6",
 ]
 
 jax = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,10 +133,6 @@ dynesty = [
 
 mltools = [
   "umap-learn[plot]>=0.5.3",
-  # `umap-learn[plot]` requires an unpinned bokeh; without a lower bound here,
-  #  resolvers satisfy `holoviews` with an ancient release whose `panel`
-  #  version is incompatible with recent bokeh, which makes `umap.plot`
-  #  unimportable.
   "holoviews>=1.18",
   "scikit-learn>=0.24.1",
 ]
--- a/pypesto/ensemble/dimension_reduction.py
+++ b/pypesto/ensemble/dimension_reduction.py
@@ -228,7 +228,6 @@ def _get_umap_representation_lowlevel(
         returned fitted umap object from umap.UMAP()
     """
     import umap
-    import umap.plot
     from sklearn.preprocessing import StandardScaler
 
     # create a umap object


pyproject.toml
```diff
@@ -127,6 +133,11 @@ dynesty = [
 
 mltools = [
   "umap-learn[plot]>=0.5.3",
+  # `umap-learn[plot]` requires an unpinned bokeh; without a lower bound here,
+  #  resolvers satisfy `holoviews` with an ancient release whose `panel`
+  #  version is incompatible with recent bokeh, which makes `umap.plot`
+  #  unimportable.
```
dilpath: Since this is about the usual practice of setting a minimum required version, I think a justification is not necessary
```suggestion
```

pyproject.toml
```diff
@@ -106,10 +106,16 @@ mpi = [
...
+  #  pymc>=6 supports it; pymc<6 declares no upper bound on ArviZ, so without
+  #  the bounds below a resolver pairs pymc 5.x with ArviZ 1.x, which fails on
+  #  `import pymc`.
+  # TODO: once Python 3.11 support is dropped, require only ArviZ >=1.1.0 and
+  #  pymc >=6.
   "arviz>=0.12.1,<1.0; python_version < '3.12'",
   "arviz>=1.1.0; python_version >= '3.12'",
-  "pymc>=4.2.1",
+  "pymc>=4.2.1,<6; python_version < '3.12'",
+  "pymc>=6; python_version >= '3.12'",
```
dilpath: NumPy already requires Python 3.12, and we follow NumPy's policy. @ICB-DCM/pypesto-maintainers if you drop Python 3.11 support in a separate PR, we can simplify this now. It looks like pymc 6 also already requires arviz > 1.1.0 so I dropped that here too.

```suggestion
  "pymc>=6",
```

pyproject.toml
```diff
@@ -127,6 +133,11 @@ dynesty = [
 
 mltools = [
   "umap-learn[plot]>=0.5.3",
+  # `umap-learn[plot]` requires an unpinned bokeh; without a lower bound here,
+  #  resolvers satisfy `holoviews` with an ancient release whose `panel`
+  #  version is incompatible with recent bokeh, which makes `umap.plot`
+  #  unimportable.
+  "holoviews>=1.18",
```
dilpath, line 127: It looks like umap-learn is somewhat actively maintained so this could be submitted there too/instead. Fine as is.

dilpath (approved):
Thanks! I would prefer we drop Python 3.11 now since it simplifies this, but fine to merge if that doesn't happen quickly.

pyproject.toml
dilpath: @PaulJonasJost Ah, it looks like dropping Python 3.11 was already implemented here but not completely: https://github.com/ICB-DCM/pyPESTO/pull/1703

FFroehlich:
@PaulJonasJost @Doresic @MoR1chter blocked until your approval

dweindl:
> @PaulJonasJost @Doresic @MoR1chter blocked until your approval

:alarm_clock:

PR #1739, opened 2026-09-02
Fix `x_free` handling in `check_gradients_match_finite_differences`

FFroehlich:
## Problem

`ObjectiveBase.check_gradients_match_finite_differences` used `x_free` to slice the parameter vector and passed the result on as the full `x`:

```python
free_indices = par[x_free]
...
self.check_grad_multi_eps(free_indices, ...)
```

So whenever `x_free` was a strict subset of the parameters, the objective was called with a wrong-length vector. For `AmiciObjective` the strict `zip` in `par_arr_to_dct` raises `ValueError`, which is swallowed by

```python
except (RuntimeError, ValueError):
    return False
```

and the check silently reported a **gradient mismatch** instead of failing. Reproducible with the Boehm PEtab v2 problem (2 PEtab-fixed parameters), where `objective.check_gradients_match_finite_differences(mode="mode_fun")` returned `False` while `objective.check_grad(x, x_indices=free_indices)` showed matching gradients.

The defect dates back to c0df8d6 (#690).

## Change

- `x_free` is forwarded as `check_grad`'s `x_indices`: the objective is evaluated at the full `x`, and finite differences are computed only for the selected indices.
- The `AmiciObjective` / `AmiciPetabV2Objective` defaults are mapped into the objective's parameter space, so parameters fixed in the `Problem` are part of neither the nominal vector nor the indices. Without this the problem-normalized objectives would break — they happened to work before, because slicing produced exactly their reduced input.
- Regression test with a cheap non-AMICI objective whose `fun`/`grad` require the full-length vector.

## Verification

- `test/base/test_objective.py`, `test/base/test_prior.py`: 153 passed, 8 skipped.
- `test/petab/test_petab_import.py::PetabImportTest::test_check_gradients` and `::test_plist_mapping`: passed.
- Boehm PEtab v2, default eps sweep: raw objective and problem-normalized objective both `True` (was `False` for the raw one).

## Not addressed here

Found while working on this, each a separate silent-wrong-answer bug in the same method:

- `order` is documented as selecting the derivative order but is never forwarded, so `order=1` checks the gradient and reports it as a Hessian check (`test_prior.py::test_derivatives` asserts this). Forwarding it makes 23 parametrizations fail, so it needs its own investigation.
- An integer-valued `x` truncates the finite-difference step, giving a false mismatch.
- The default `mode=None` returns `False` for any objective without residuals.
- `multi_eps=[]` raises `AttributeError` on a `None` result.

### DIFF
--- a/pypesto/objective/amici/amici.py
+++ b/pypesto/objective/amici/amici.py
@@ -667,9 +667,23 @@ class AmiciObjective(ObjectiveBase):
             Indicates whether gradients match (True) FDs or not (False)
         """
         if x is None and "petab_problem" in dir(self.amici_object_builder):
-            x = self.amici_object_builder.petab_problem.x_nominal_scaled
-            kwargs["x_free"] = (
-                self.amici_object_builder.petab_problem.x_free_indices
+            petab_problem = self.amici_object_builder.petab_problem
+            # parameters fixed in the `Problem` are not part of this
+            #  objective's parameter vector, so drop them from the nominal
+            #  values and map the free-parameter indices accordingly
+            x_full = np.asarray(petab_problem.x_nominal_scaled)
+            x = self.pre_post_processor.reduce(x_full)
+            input_to_full = self.pre_post_processor.reduce(
+                np.arange(len(x_full))
+            )
+            free = set(petab_problem.x_free_indices)
+            kwargs.setdefault(
+                "x_free",
+                [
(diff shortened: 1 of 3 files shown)


PaulJonasJost (approved):
Looks good. thanks for noticing the problem.

dweindl (approved):

PR #1740, opened 2026-09-02
Hierarchical optimization for the PEtab v2 importer

FFroehlich:
## What

Adds hierarchical optimization (analytically solved scaling, offset and sigma inner parameters for relative data) to the PEtab v2 import path, which previously raised `NotImplementedError` for `hierarchical=True`.

* `InnerCalculatorCollectorPetabV2` — v2 counterpart of `InnerCalculatorCollector`. Simulation is delegated to AMICI's `PetabSimulator` through `AmiciCalculatorPetabV2`; the inner problems are solved from the returned observables, sigmas and sensitivities. Simulation sensitivities are mapped onto the optimization parameters via per-experiment index slices derived from `ExpData.plist` (no PEtab-v1-style parameter mapping is constructed for v2). Adjoint sensitivities, the FIM and residual mode go through the same twice-call scheme as v1.
* `RelativeInnerProblem.from_petab_v2_amici` and the v2 inner-problem construction in `hierarchical/relative/problem.py`, reusing the v1 per-condition measurement indexing on the preprocessed problem's `measurement_df`.
* `validate_hierarchical_petab_problem_v2` — formula-form validation for v2 (explicit placeholders, `real=True` symbols), sharing the checks with v1 via extracted helpers.
* The pyPESTO annotations (`parameterType`, `measurementType`) are read from the PEtab v2 `model_extra` fields, which survive `petab1to2` and AMICI's experiment conversion.

Shared v1 code changes are minimal: `RelativeInnerSolver.calculate_gradients` and `calculate_quantitative_result` gain one trailing `index_slices=None` argument; `_get_quantitative_data_mask` uses `any` instead of `all` (with `all`, one condition without quantitative data dropped the quantitative contribution of every condition).

## Scope / limitations

* Only relative (and quantitative) data. Ordinal, censored and semiquantitative data are detected and rejected with `NotImplementedError`.
* Observable-mapping plots (`visualize.observable_mapping`) raise `NotImplementedError` for v2 objectives; they re-simulate via the v1 parameter mapping.
* Measurement-specific placeholder overrides that differ within one experiment are rejected: AMICI applies one value per placeholder and experiment, and its own guard for this never fires.
* Sigma bounds are forced to `INNER_PARAMETER_BOUNDS` (as `correct_parameter_df_bounds` does for v1); a scaling/offset pair that is only partially estimated is rejected rather than silently treated as uncoupled.

## Verification

* `test/hierarchical/test_hierarchical_petab_v2.py`: structure, v1/v2 equivalence of fval, gradient and inner parameters on Boehm, finite-difference gradient check, adjoint/FIM/residual consistency, agreement with the non-hierarchical objective at the inner optimum, a short optimization run, validation and coupled-pair cases.
* Out-of-tree probes (Boehm, and a 2-experiment synthetic problem with replicates and per-condition overrides) agree between v1 and v2 to ~1e-16 relative gradient error.
* `test_hierarchical_optimization_pipeline` fails locally on `develop` as well; unrelated.

## Notes for review

* `PetabSimulationResult.res` is a method in amici ≤ 1.0.1 and a property on AMICI main; `AmiciCalculatorPetabV2` handles both. The `petab` tox env installs AMICI main, the `hierarchical` env the PyPI release.
* History: one feature commit, four review-fix commits, then a slimming commit that removed most of what the review rounds had added. Happy to squash.

## Stack

8. **this PR** — the PEtab v2 hierarchical calculator

### DIFF
--- a/pypesto/hierarchical/inner_calculator_collector.py
+++ b/pypesto/hierarchical/inner_calculator_collector.py
@@ -9,9 +9,13 @@ from __future__ import annotations
 import copy
 import warnings
 from collections.abc import Sequence
-from typing import Union
+from typing import TYPE_CHECKING, Union
 
 import numpy as np
+
+if TYPE_CHECKING:
+    import amici.sim.sundials.petab
+    from petab import v2
 
 from ..C import (
     AMICI_SIGMAY,
@@ -43,6 +47,7 @@ from ..objective.amici.amici_util import (
     filter_return_dict,
     init_return_values,
     par_index_slices,
+    petab_v2_index_slices,
 )
 
 try:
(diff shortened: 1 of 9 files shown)


pypesto/hierarchical/relative/problem.py
```diff
@@ -20,8 +21,10 @@
 try:
     import amici.sim.sundials as asd
     import petab.v1 as petab
+    from petab import v2
     from petab.v1.C import (
         ESTIMATE,
+        LIN,
```
dweindl, line 27: Shadows the previous one. Currently no effect, but I think we should be explicit about which one we mean.

pypesto/hierarchical/inner_calculator_collector.py
```diff
@@ -666,6 +671,204 @@ def __call__(
...
+        sensi_orders: tuple[int],
+        mode: ModeType,
+        amici_model: AmiciModel,
+        amici_solver: AmiciSolver,
+        edatas: list[asd.ExpData],
+        n_threads: int,
+        x_ids: Sequence[str],
+        parameter_mapping: ParameterMapping,
+        fim_for_hess: bool,
+    ) -> dict | None:
```
Doresic, line 776: I like the separation of the inner calculator collector, it's nice especially for the v2 so it can be changed.
Can we name this function differently though? That part of code was more for directing or delegating the calculation to the relative calculator in cases where only it could handle the setting (adjoints, 2nd order, RES). So I would prefer it called something like `_direct_to_relative_calculator`

pypesto/hierarchical/inner_calculator_collector.py
Doresic, line 419: Would rename, see v2 function call comment.

pypesto/hierarchical/inner_calculator_collector.py
```diff
@@ -666,6 +671,204 @@ def __call__(
...
+        """See :meth:`InnerCalculatorCollector._inner_only_result`.
+
+        Unlike the PEtab v1 dispatch, this one also requires a gradient
+        before it takes the adjoint route.
+        """
```
Doresic, line 781: taking a closer look, I'm not sure this override is needed, since #1753 was merged.
Adding the `1 in sensi_orders` can be done in all cases too. We can go through the `calculate_directly` routing if we need to calculate order 0 only in all cases. So adding `1 in sensi_orders` in this `if clause` can be done generally

- The stated reason for calling call_amici_twice directly (~L799: "whose own dispatch does not route residual mode here") no longer holds — RelativeAmiciCalculator.__call__ now routes MODE_RES to call_amici_twice (relative/calculator.py ~L176), and its dispatch condition is identical to this hook's predicate.
- The dummy values applied here (~L795) look redundant: for v2, necessary_par_dummy_values is exactly relative_inner_problem.get_dummy_values(scaled=True) (collector __init__, ~L146), and call_amici_twice applies that itself (~L261).
- The INNER_PARAMETERS packaging (~L813) duplicates what __call__ already does (~L205).
- Going through __call__ would also bring back the check_edatas guard (~L158) that the direct call currently skips — that should be fine for v2, since the inner problem is built from the same edatas. Or was this the call that had to be circumvented?

I might be missing something though

Doresic (commented):
Some initial comments specific to the `_only_inner_result` method, continuing the review

pypesto/hierarchical/inner_calculator_collector.py
Doresic: Not this PR's code, but noticed while reviewing: the _known_least_squares_safe block in the base _simulate is dead now — MODE_RES never reaches _simulate since the template routes it to the hook first, and the same check lives in AmiciCalculator anyway. Happy to open a small issue/PR to remove it.

pypesto/hierarchical/relative/problem.py
```diff
@@ -326,3 +381,121 @@ def _ixs_for_condition(
...
+    ix_matrices = ix_matrices_from_arrays(ixs, data)
+
+    # assign matrices, observable indices and ids to inner parameters
+    for par in inner_parameters:
+        par.ixs = ix_matrices[par.inner_parameter_id]
+        par.observable_indices = [
+            meas_indices[2] for meas_indices in ixs[par.inner_parameter_id]
+        ]
+        par.observable_ids = [
+            amici_model.get_observable_ids()[obs_idx]
```
dweindl, line 421: store observable ids once outside the nested loop

test/hierarchical/test_hierarchical_petab_v2.py
```diff
@@ -2,20 +2,328 @@
...
+def objective_v2(importer_v2):
+    """A v2 hierarchical objective that is not part of a problem.
+
+    Takes the full vector of outer parameters, including the ones fixed in
+    the pypesto problem.
+    """
+    return importer_v2.create_objective_creator().create_objective()
+
+
+def _to_linear(x_scaled, scales):
```
dweindl, line 79: There is `pypesto.hierarchical.base_problem.scale_back_value` (needs `map`)

test/hierarchical/test_hierarchical_petab_v2.py
```diff
@@ -2,20 +2,328 @@
...
+    ) == sorted(problem_v1.objective.calculator.get_inner_par_ids())
+    assert problem_v2.x_names == problem_v1.x_names
+    assert problem_v2.x_fixed_indices == problem_v1.x_fixed_indices
+    assert sorted(problem_v2.inner_x_names) == sorted(problem_v1.inner_x_names)
+
+    # inner parameters are the parameters with a parameterType annotation
+    expected_inner_ids = [
+        parameter.id
+        for parameter in petab_problem_v2.parameters
+        if (parameter.model_extra or {}).get("parameterType")
```
dweindl, line 119: pypesto.C.PARAMETER_TYPE

dweindl (approved):
Thanks!

Just minor things. Mostly relying on @Doresic's expertise here.

pypesto/hierarchical/inner_calculator_collector.py
Doresic, line 545: Not this PR's code, but noticed while reviewing: the _known_least_squares_safe block in the base _simulate is dead now — MODE_RES never reaches _simulate since the template routes it to the hook first, and the same check lives in AmiciCalculator anyway. We can address this is a later PR.

pypesto/hierarchical/inner_calculator_collector.py
```diff
@@ -666,6 +671,204 @@ def __call__(
...
+        n_threads: int,
+        x_ids: Sequence[str],
+        parameter_mapping: ParameterMapping,
+        fim_for_hess: bool,
+    ) -> tuple[list[asd.ReturnDataView], dict | None]:
+        """See :meth:`InnerCalculatorCollector._simulate`.
+
+        The PEtab simulator handles the parameter mapping, so
+        ``parameter_mapping`` is unused.
+        """
```
Doresic, line 842: I like the separation/addition of simulate and evaluate. I think it might even be possible to define an `evaluator` for v1 that will be faster (doing index_slices only once) such that we can have only one `_simulate` method, to work exactly in the same way in both v1 and v2. Also can be done in a subsequent PR

pypesto/hierarchical/relative/calculator.py
```diff
@@ -294,30 +294,40 @@ def call_amici_twice(
...
-        x_dct = {
-            par_id: value
-            for par_id, value in x_dct.items()
-            if par_id not in inner_values
-        }
+        if self.evaluator is not None:
+            # The evaluator simulates from `x_dct` and ignores the PEtab v1
+            #  parameter mapping, so the optimal values have to go there. Its
+            #  plist comes from the problem's free parameters, which exclude
+            #  the inner ones, so the mapping rewrite below is unnecessary.
```
Doresic, line 301: A small change, I believe plist still contains the inner parameters. They are only removed from par_opt_ids, from pypesto's problem.

pypesto/objective/amici/amici.py
```diff
@@ -796,11 +808,34 @@ def __init__(
...
+            )
+
+            calculator = InnerCalculatorCollectorPetabV2(
+                data_types=non_quantitative_data_types,
+                petab_simulator=self._petab_simulator,
+                inner_options=inner_options or {},
+            )
+            # the inner problems are solved from the observables, sigmas and
+            #  their sensitivities
+            kwargs["amici_reporting"] = asd.RDataReporting.full
```
Doresic, line 823: Does it need to be full? That can be very slow in some cases? Is .likelihood sufficient here?

pypesto/objective/amici/amici.py
```diff
@@ -796,11 +808,34 @@ def __init__(
...
+            # parameters estimated in the inner subproblems are removed from
+            #  the objective parameters
+            inner_parameter_ids = set(calculator.get_inner_par_ids())
+            x_ids = kwargs.get("x_ids") or self.petab_problem.x_ids
+            kwargs["x_ids"] = [
+                x_id for x_id in x_ids if x_id not in inner_parameter_ids
+            ]
+        else:
+            calculator = AmiciCalculatorPetabV2(self._petab_simulator)
+
```
Doresic, line 833: To make things work and choose things in the same places, can we move this part to the `create_objective` method of the `AmiciPetabV2ObjectiveCreator` instead? So it works exactly as v1 version works -- same things done in same places?
The the chosen calculator would just be passed to this class and then up to the parent as it already is.

pypesto/visualize/observable_mapping.py
```diff
@@ -92,6 +95,25 @@ def _finalize_observable_mapping_axes(
...
+    """
+    if not isinstance(calculator, InnerCalculatorCollector):
+        raise ValueError(
+            "The calculator must be an instance of the InnerCalculatorCollector."
+        )
+    if isinstance(calculator, InnerCalculatorCollectorPetabV2):
+        raise NotImplementedError(
+            "Visualizing the observable mapping is not yet supported for "
+            "PEtab v2 problems."
+        )
```
Doresic, line 114: Yeah true, all of this will also be touched significantly with the visaulisation module changes, probably potentially the whole visualisation, as it's definitely not in a good shape right now.

Doresic (approved):
Thank you for the implementation! I added some comments, they are not necessary to address in this PR tho. Most of them are just about renaming or moving code around.

PR #1742, opened 2026-09-07
Fix two pre-existing hierarchical/PEtab v2 bugs

FFroehlich:
Two independent bug fixes found while implementing hierarchical optimization for the PEtab v2 importer. Split out of that work so they can be reviewed and merged on their own. **1 of 4 in a stack** (see below).

## 1. Quantitative data silently dropped when one condition has none

`InnerCalculatorCollector._get_quantitative_data_mask` returned `None` unless *every* condition had quantitative data:

```python
if not all(mask.any() for mask in quantitative_data_mask):
    return None
```

`None` means "no quantitative contribution at all", so a single condition without quantitative data — common when one experiment measures only relative observables — removed the quantitative contribution of *all* conditions from the objective. Silently: no error, just a wrong objective value and gradient.

An all-False mask contributes nothing on its own, so `any` is the correct quantifier.

## 2. `AmiciCalculatorPetabV2` returned a bound method as the residuals

```python
if (res := result.res) is None:
```

`PetabSimulationResult.res` is a *method* in amici <= 1.0.1 and only became a property later. With the released amici, `result.res` is therefore a bound method: the `is None` check never fires, and the method object is returned in the result dict as `RES`.

Verified against the amici in the dev environment: `PetabSimulationResult.res` is a `function`.

## Review notes

* Both bugs are on `develop` today; neither is introduced by the PEtab v2 work.
* Bug 1 affects the PEtab v1 hierarchical path, bug 2 the PEtab v2 residual path. They are unrelated to each other and only grouped because both are small and were found together — happy to split further.
* Neither has a regression test here. Bug 1 needs a multi-condition fixture where one condition is purely relative; bug 2 needs `MODE_RES` on a v2 problem, which the stack's last PR exercises.

## Stack

2. **this PR** — pre-existing bug fixes

### DIFF
--- a/pypesto/hierarchical/inner_calculator_collector.py
+++ b/pypesto/hierarchical/inner_calculator_collector.py
@@ -246,8 +246,10 @@ class InnerCalculatorCollector(AmiciCalculator):
         ):
             condition_mask[np.isnan(edata)] = False
 
-        # If there is no quantitative data, return None
-        if not all(mask.any() for mask in quantitative_data_mask):
+        # If there is no quantitative data at all, return None. Individual
+        #  conditions without quantitative data are fine -- their (all-False)
+        #  mask simply contributes nothing.
+        if not any(mask.any() for mask in quantitative_data_mask):
             return None
 
         return quantitative_data_mask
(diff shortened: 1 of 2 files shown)


Doresic (approved):
Looks good to me, thank you for noticing and fixing this
We can add a test for the hierarchical issue, might not be necessary though. Open to both.

dweindl (approved):
👍

PR #1743, opened 2026-09-07
Index slices: PEtab v2 support, and v1 refactors for hierarchical

FFroehlich:
Pure refactoring ahead of PEtab v2 hierarchical support. **No behaviour change for PEtab v1**; the v1 hierarchical test suite passes unchanged (43 passed, plus the `test_hierarchical_optimization_pipeline` failure that is already present on `develop`).

Split out so the PEtab v2 PRs that build on it only *add* code. **2 of 4 in a stack** (see below).

## `hierarchical/petab.py` — split the formula checks

`_validate_measurement_specific_observable_formula` and `..._noise_formula` did two things at once: work out which inner parameters appear in a measurement's formula, and check the shape of the result (`y = scaling * (...) + offset`, sigma constituting the whole noise formula).

The second half is version-independent, so it moves into `_validate_observable_formula_form`, `_validate_noise_formula_form` and `_get_formula_inner_parameters`, which take an already-substituted formula plus its inner parameters. The v1 entry points still infer those from the `observableParameter{n}_{id}` naming convention; PEtab v2 will pass its explicit placeholders instead.

One behavioural subtlety inside the refactor: inner parameters are now matched by **symbol name** rather than by `sp.Symbol(id) in free_symbols`. PEtab v2 formula symbols carry assumptions (`real=True`) and so never compare equal to a plain `sp.Symbol`. For v1 the two are equivalent.

## `relative/problem.py` — extract `_ixs_for_condition`

The per-condition body of `ixs_for_measurement_specific_parameters` (replicate-aware timepoint indexing) becomes a helper, so it can be reused for a PEtab v2 problem's measurement table rather than reimplemented.

## `relative/solver.py`, `relative/calculator.py`, `inner_calculator_collector.py` — index slices

`calculate_gradients` and `calculate_quantitative_result` gain a trailing `index_slices=None`: pre-computed `(par_sim_slice, par_opt_slice)` pairs per condition. When not given they are derived from the parameter mapping exactly as before, via the existing `par_index_slices`.

The accumulation loop is now a single `np.add.at`, replacing the `np.unique`-plus-fixup-loop. Equivalent for repeated indices, and measurably faster (~3.7x, ~85x when indices repeat), verified over 200 randomized cases.

## `amici_util.py` — where PEtab v2 gets its slices

PEtab v2 problems are already supported, just not hierarchically, and there was no v2 equivalent of `par_index_slices`. There is no parameter mapping to derive one from: model parameters carry the PEtab parameter IDs, except the observable and noise placeholders, which are overridden per experiment, and sensitivities come back in `ExpData.plist` order.

`petab_v2_index_slices` resolves that into the same pairs `par_index_slices` produces; `petab_v2_placeholder_mapping` resolves the placeholders and rejects overrides that differ between measurements of one experiment. AMICI applies a single value per placeholder and experiment, so a conflict would be silently mis-simulated and its gradient misattributed — and AMICI's own guard never fires, because the dict it checks is never written.

**This is the part worth checking closely, and it is tested on its own.** `test_petab_v2_index_slices.py` uses the plain, non-hierarchical v2 Boehm problem: AMICI aggregates the per-experiment sensitivities into a gradient itself, and assembling the same gradient through these slices must reproduce it. That pins the correspondence end to end with no hierarchical code involved.

## Review notes

* `pypesto/objective/amici/amici_util.py` is **unchanged** — `par_index_slices` and `add_sim_grad_to_opt_grad` keep their signatures and their callers.
* The reviewer's question for this PR is only "is this equivalent to what was there?".

## Stack

3. **this PR** — index slices: PEtab v2 support, and v1 refactors

### DIFF
--- a/pypesto/hierarchical/petab.py
+++ b/pypesto/hierarchical/petab.py
@@ -329,7 +329,32 @@ def _validate_measurement_specific_observable_formula(
         petab_problem=petab_problem,
         inner_parameters=inner_parameters,
     )
+    return _validate_observable_formula_form(
+        formula=formula, formula_inner_parameters=formula_inner_parameters
+    )
 
+
+def _validate_observable_formula_form(
+    formula: sp.Expr,
+    formula_inner_parameters: dict[sp.Symbol, InnerParameterType],
+) -> tuple[sp.Symbol, sp.Symbol]:
+    """Check that inner parameters enter an observable formula as expected.
+
+    The observable formula should take the form
+    ``y = scaling * (...) + offset``.
+
+    Parameters
+    ----------
+    formula:
+        The symbolic observable formula, with any placeholders replaced by
+        the measurement-specific overrides.
+    formula_inner_parameters:
+        The inner parameters appearing in the formula, and their types.
+
+    Returns
+    -------
+    The offset and scaling parameters.
+    """
     offset = None
     scaling = None
 
@@ -414,7 +439,31 @@ def _validate_measurement_specific_noise_formula(
         petab_problem=petab_problem,
         inner_parameters=inner_parameters,
     )
+    return _validate_noise_formula_form(
+        formula=formula, formula_inner_parameters=formula_inner_parameters
+    )
 
+
+def _validate_noise_formula_form(
+    formula: sp.Expr,
+    formula_inner_parameters: dict[sp.Symbol, InnerParameterType],
+) -> sp.Symbol:
+    """Check that inner parameters enter a noise formula as expected.
+
+    A sigma inner parameter must constitute the full noise formula.
+
+    Parameters
+    ----------
+    formula:
+        The symbolic noise formula, with any placeholders replaced by the
+        measurement-specific overrides.
+    formula_inner_parameters:
+        The inner parameters appearing in the formula, and their types.
+
+    Returns
+    -------
+    The sigma parameter.
+    """
     sigma = None
 
     for (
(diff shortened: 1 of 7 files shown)


pypesto/hierarchical/relative/calculator.py
```diff
@@ -65,6 +65,10 @@ def __init__(
             inner_solver = AnalyticalInnerSolver()
         self.inner_solver = inner_solver
 
+        #: Per-condition ``(par_sim_slice, par_opt_slice)`` index pairs, set by
+        #: the PEtab v2 collector; ``None`` derives them from the mapping.
+        self.index_slices: list[tuple[np.ndarray, np.ndarray]] | None = None
```
dweindl: Slightly more extensive documentation as in pypesto/hierarchical/relative/solver.py would be nice.

pypesto/hierarchical/petab.py
```diff
@@ -517,10 +566,62 @@ def _get_symbolic_formula_from_measurement(
...
+                f"Observable transformation: `{observable_transformation}`. "
+                f"Measurement:\n{measurement}"
+            )
+
+    return symbolic_formula, symbolic_formula_inner_parameters
+
+
+def _get_formula_inner_parameters(
+    symbolic_formula: sp.Expr,
+    formula_type: Literal["observable", "noise"],
```
dweindl: Let's not use `Literal`, unless we really require a string literal (i.e. not some variable). Static type checkers will complain. (It's used in other places too, some of them added by me, but should be removed.)

pypesto/objective/amici/amici_util.py
```diff
@@ -213,6 +214,121 @@ def par_index_slices(
...
+                observable.observable_placeholders,
+                measurement.observable_parameters,
+            ),
+            (observable.noise_placeholders, measurement.noise_parameters),
+        ):
+            for placeholder, override in zip(
+                placeholders, overrides, strict=True
+            ):
+                if not override.is_Symbol:
+                    continue
```
dweindl, line 256: Can you remind me, and add a comment, when/where that happens?

test/petab/test_petab_v2_index_slices.py
```diff
@@ -0,0 +1,180 @@
...
+    assembled = np.zeros(len(objective.x_ids))
+    for rdata, (par_sim_slice, par_opt_slice) in zip(
+        rdatas, index_slices, strict=True
+    ):
+        np.add.at(assembled, par_opt_slice, -rdata.sllh[par_sim_slice])
+
+    assert np.allclose(assembled, ret["grad"], rtol=1e-8, atol=1e-8), (
+        f"assembled={assembled}\nexpected={ret['grad']}"
+    )
+    # the test is only meaningful if the slices actually select something
```
dweindl: ... and if most `ret["grad"] > 1e-8`. just to be sure 😅

test/petab/test_petab_v2_index_slices.py
```diff
@@ -0,0 +1,188 @@
...
+    observable.observable_placeholders = [
+        f"observableParameter1_{observable_id}"
+    ]
+    return petab_problem, experiment, siblings
+
+
+def test_petab_v2_placeholder_mapping_resolves_overrides(boehm_v2_objective):
+    """A placeholder overridden consistently maps to the overriding id."""
+    _, objective = boehm_v2_objective
+    petab_problem, experiment, siblings = _problem_with_one_placeholder(
```
dweindl: I don't find `siblings` very intuitive here. Maybe some version of `measurements` would work better?

dweindl (approved):

PR #1744, opened 2026-09-07
Detect and validate hierarchical PEtab v2 problems

FFroehlich:
Adds the PEtab v2 side of problem inspection: reading pyPESTO's hierarchical annotations from a v2 problem, and validating that such a problem can actually be optimized hierarchically.

Nothing consumes this yet — the importer still rejects hierarchical PEtab v2 problems — so this PR only makes the annotations readable and the problems checkable. **3 of 4 in a stack** (see below).

## Reading the annotations

`parameterType` and `measurementType` are pyPESTO extensions, not part of the PEtab format. In PEtab v2 they survive as pydantic **extra fields** on the table elements — through TSV loading, the v1-to-v2 upgrade, and AMICI's experiment conversion. `get_petab_v2_extra_field` reads them, treating absent and empty values alike via `petab.is_empty`.

`get_petab_non_quantitative_data_types` now dispatches to a v2 implementation that reads those extra fields instead of relying on the v1 dataframe columns. Ordinal, censored and semiquantitative data are detected so the importer can reject them, rather than silently treating them as quantitative.

## Validating

`validate_hierarchical_petab_problem` accepts a v2 problem. Placeholders are explicit in v2, so the measurement-specific overrides are substituted directly instead of being inferred from the `observableParameter{n}_{id}` convention, and the resulting formulae go through the validators extracted in #1743.

One check has no v1 counterpart: **non-Gaussian noise is rejected** for observables carrying inner parameters. The analytical inner solver assumes additive Gaussian noise, so hierarchically fitting an observable declared with a Laplace noise distribution silently optimizes the wrong objective. v1 has this hole; this only closes it for v2.

## Review notes

* `get_petab_non_quantitative_data_types` is called for **every** problem the importer builds, v1 and v2, hierarchical or not — this changes which code path v2 problems take through it. The two non-hierarchical v2 import tests still pass.
* The `dtype=object` on the inner-parameter dataframe is load-bearing on pandas 3: without it a mixed str/None column becomes a string dtype whose missing value is NaN, and the pairing check compares against `None` (`nan != nan`). The same one-line issue exists in the v1 function and is worth porting separately.
* Test coverage here is validation only: unknown parameter type, unsupported measurement type, non-additive offset, a sigma that is not the whole noise formula, and non-Gaussian noise.

## Stack

7. **this PR** — PEtab v2 detection and validation

### DIFF
--- a/pypesto/hierarchical/petab.py
+++ b/pypesto/hierarchical/petab.py
@@ -6,6 +6,7 @@ import pandas as pd
 import petab.v1 as petab
 import sympy as sp
 from more_itertools import one
+from petab import v2
 from petab.v1.C import (
     ESTIMATE,
     LIN,
@@ -67,7 +68,9 @@ def correct_parameter_df_bounds(parameter_df: pd.DataFrame) -> pd.DataFrame:
     return parameter_df.apply(correct_row, axis=1)
 
 
-def validate_hierarchical_petab_problem(petab_problem: petab.Problem) -> None:
+def validate_hierarchical_petab_problem(
+    petab_problem: petab.Problem | v2.Problem,
+) -> None:
     """Validate a PEtab problem for hierarchical optimization.
 
     Parameters
@@ -75,6 +78,10 @@ def validate_hierarchical_petab_problem(petab_problem: petab.Problem) -> None:
     petab_problem:
         The PEtab problem.
     """
+    if isinstance(petab_problem, v2.Problem):
+        validate_hierarchical_petab_problem_v2(petab_problem)
+        return
+
     if PARAMETER_TYPE in petab_problem.parameter_df:
         # ensure we only have linear parameter scale
         inner_parameter_table = petab_problem.parameter_df[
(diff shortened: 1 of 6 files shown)


pypesto/hierarchical/petab.py
```diff
@@ -659,6 +666,171 @@ def _get_formula_inner_parameters(
...
+                observable.noise_placeholders,
+                measurement.noise_parameters,
+                strict=True,
+            )
+        )
+
+        observable_formula = sp.sympify(observable.formula).subs(substitutions)
+        noise_formula = sp.sympify(observable.noise_formula).subs(
+            substitutions
+        )
```
dweindl: Is `sympify` required here? Might be a bug in libpetab then.

pypesto/petab/util.py
```diff
@@ -27,8 +27,19 @@
...
+    ``None`` if the field is absent or empty.
+    """
+    value = (element.model_extra or {}).get(field)
+    return None if value is None or petab.is_empty(value) else value
+
+
 def get_petab_non_quantitative_data_types(
-    petab_problem: petab.Problem,
+    petab_problem: petab.Problem | v2.Problem,
 ) -> set[str]:
```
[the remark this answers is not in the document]
dweindl: I don't think the extra complexity of returning None instead of an empty set is really helping here. (Or am I missing something?) The issue was present before, but I'd be happy to have it cleaned up instead of propagating it to the v2 function.

pypesto/petab/util.py
```diff
@@ -97,6 +111,45 @@ def get_petab_non_quantitative_data_types(
...
+    # Ordinal, censored and semiquantitative data are not supported for PEtab
+    # v2 yet, but they still have to be detected here so that they are
+    # rejected with a clear message rather than silently treated as
+    # quantitative data.
+    for measurement in petab_problem.measurements:
+        data_type = get_petab_v2_extra_field(measurement, MEASUREMENT_TYPE)
+        if data_type in [ORDINAL, SEMIQUANTITATIVE] + CENSORING_TYPES:
+            non_quantitative_data_types.add(
+                CENSORED if data_type in CENSORING_TYPES else data_type
+            )
```
dweindl: I think splitting that into an `elif` branch is clearer than first combining and then distinguishing those types.

test/hierarchical/test_hierarchical_petab_v2.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_hierarchical_petab_v2_validation(petab_problem_v2):
+    """Invalid hierarchical PEtab v2 problems are rejected."""
+    # the unmodified problem is valid
+    validate_hierarchical_petab_problem(petab_problem_v2)
+
+    # unknown parameter type
+    petab_problem = copy.deepcopy(petab_problem_v2)
+    petab_problem.parameters[-1].model_extra["parameterType"] = "pink"
```
dweindl:
```suggestion
    petab_problem.parameters[-1].model_extra[PARAMETER_TYPE] = "pink"
```

test/hierarchical/test_hierarchical_petab_v2.py
```diff
@@ -0,0 +1,59 @@
+"""Tests for hierarchical optimization with PEtab v2 problems."""
+
+import copy
+
+import pytest
+
+from pypesto.C import MEASUREMENT_TYPE, ORDINAL
```
dweindl:
```suggestion
from pypesto.C import MEASUREMENT_TYPE, ORDINAL, PARAMETER_TYPE
```

dweindl (approved):

pypesto/hierarchical/petab.py
```diff
@@ -67,14 +68,20 @@ def correct_row(row: pd.Series) -> pd.Series:
     return parameter_df.apply(correct_row, axis=1)
 
 
-def validate_hierarchical_petab_problem(petab_problem: petab.Problem) -> None:
+def validate_hierarchical_petab_problem(
+    petab_problem: petab.Problem | v2.Problem,
```
PaulJonasJost: I find it confusing to call one a `petab.Problem` and the other only a `v2.Problem`. Would probably prefer `petab.v2.Problem` and `petab.v1.Problem`

PaulJonasJost (approved):

pypesto/hierarchical/petab.py
```diff
@@ -659,7 +666,171 @@ def _get_formula_inner_parameters(
...
+    petab_problem:
+        The PEtab v2 problem.
+    inner_parameters:
+        See :func:`get_inner_parameters_v2`.
+
+    Returns
+    -------
+    A dataframe containing the inner parameters for each measurement.
+    """
+    import petab.v2.C as petab_v2_C
```
dilpath, line 748: Not already available as `v2.C`?

dilpath (approved):
Looks good, thanks.

> hierarchically fitting an observable declared with a Laplace noise distribution silently optimizes the wrong objective. v1 has this hole; this only closes it for v2.

Also fix or open issue?

> Ordinal, censored and semiquantitative data are not supported for PEtab
v2 yet, but they still have to be detected here so that they are
rejected with a clear message rather than silently treated as quantitative data.

Make explicit in the title? e.g. "Hierarchical: validate PEtab v2 problems with relative data".

Doresic (approved):
Thank you, looks good.

Yeah some v1 things need to be fixed, I can take care of it after this is in.
Also I'm really interested in the ordinal/semiquantitative integration, but those might change significantly in the following months ,so I would do it as they do.

PR #1749, opened 2026-09-07
Extract the inner-result combination from the collector

FFroehlich:
Pure refactoring of the PEtab v1 hierarchical collector. **No behaviour change**; the v1 hierarchical tests pass unchanged (44 passed, plus the `test_hierarchical_optimization_pipeline` failure already present on `develop`).

`InnerCalculatorCollector.__call__` does two separable things: it produces simulations, and it then combines them. The second half — run each inner calculator, accumulate the objective and gradient, collect the interpretable inner parameters and spline knots, add the quantitative contribution and assemble the return dict — moves into `_combine_inner_results`. How the simulations are produced stays where it is.

## Why now

Split out of the PEtab v2 hierarchical work. The v2 collector performs exactly this combination on simulations produced a different way, so without the extraction it duplicates all of it. Landing the refactor on its own keeps the v2 PR to code that is actually new, and lets this be reviewed for equivalence alone.

## Review notes

* The only judgement call is the interface: `call_kwargs` is passed as a dict rather than a dozen explicit parameters, because both callers forward the same arguments to the inner calculators verbatim. Happy to spell them out if you prefer.
* `dim`, and the `init_return_values` tuple, remain in `__call__` because the simulation-failure path still uses them.
* The extracted body is byte-identical to what it replaces apart from the argument plumbing.

> Replaces #1747, which GitHub auto-closed as merged when its head branch became an ancestor of its former base during the restacking. Same branch, same content, correct base.

## Stack

4. **this PR** — extract the inner-result combination

### DIFF
--- a/pypesto/hierarchical/inner_calculator_collector.py
+++ b/pypesto/hierarchical/inner_calculator_collector.py
@@ -301,6 +301,98 @@ class InnerCalculatorCollector(AmiciCalculator):
             for scale in inner_calculator.inner_problem.get_interpretable_x_scales()
         ]
 
+    def _combine_inner_results(
+        self,
+        rdatas: list[asd.ReturnDataView],
+        x_dct: dict,
+        sensi_orders: tuple[int],
+        mode: ModeType,
+        amici_model: AmiciModel,
+        amici_solver: AmiciSolver,
+        edatas: list[asd.ExpData],
+        n_threads: int,
+        x_ids: Sequence[str],
+        parameter_mapping: ParameterMapping,
+        fim_for_hess: bool,
+        index_slices: list[tuple[np.ndarray, np.ndarray]] | None = None,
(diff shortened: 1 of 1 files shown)


pypesto/hierarchical/inner_calculator_collector.py
```diff
@@ -301,6 +301,82 @@ def get_interpretable_inner_par_scales(self) -> list[str]:
             for scale in inner_calculator.inner_problem.get_interpretable_x_scales()
         ]
 
+    def _combine_inner_results(
+        self,
+        rdatas: list[asd.ReturnDataView],
+        call_kwargs: dict,
```
dweindl: I'd prefer passing the arguments separately. (Does "call" refer to `__call__`? I don't like that dependency inversion.)

Otherwise good.

PR #1751, opened 2026-09-07
Make the inner-calculator collector's __call__ a template

FFroehlich:
`InnerCalculatorCollector.__call__` does three things at once: it rejects unsupported data-type and mode combinations, it decides whether the relative calculator can produce the result on its own, and it runs the simulate-then-solve pass. Only the last two are backend-specific, so this splits them out as overridable hooks:

* `_inner_only_result(call_kwargs)` returns the inner calculator's own result when that suffices, `None` when the collector has to simulate.
* `_simulate(...)` runs the simulations with the inner parameters at their dummy values, returning the results plus a ready-made return dict for the failure case.

`__call__` keeps the checks, builds `call_kwargs` once, applies the dummy values and drives the two hooks. `_index_slices` moves onto the base class so a subclass that knows the sensitivity-to-parameter correspondence can supply it.

**PEtab v1 behaviour is unchanged**: 43 passed, plus the `test_hierarchical_optimization_pipeline` failure already on `develop` (`fides` missing in this environment).

## Why the dispatch is a hook rather than shared

The two versions disagree about when the relative calculator answers alone, and the difference is not cosmetic:

| | v1 | v2 |
|---|---|---|
| adjoint | always | only with a gradient requested |

Unifying them would change v1 behaviour, so the predicate stays inside the hook next to the action it guards. **This surfaced a latent v1 defect that I have deliberately not fixed here** — see below.

## A pre-existing v1 bug, fixed separately in #1753

In residual mode v1 routes to `RelativeAmiciCalculator.__call__`, whose own dispatch omits `MODE_RES` and therefore picks `calculate_directly`. That method takes `res` and `sres` from `init_return_values` and never fills them. On PEtab v1 hierarchical Boehm:

```
PEtab v1 hierarchical, MODE_RES:
  res shape: (0,)
  n measurements in problem: 48
```

So v1 hierarchical returns empty residuals. That is out of scope for a no-behaviour-change refactor, so it is fixed on its own in #1753, which routes v1 residual mode through the two-call scheme and adds a test. #1753 targets `develop` directly and is independent of this stack; once both have landed, the v1 and v2 dispatch predicates agree and one of the two hooks here can collapse back into the template.

## Stack

6. **this PR** — `__call__` as a template with two hooks

### DIFF
--- a/pypesto/hierarchical/inner_calculator_collector.py
+++ b/pypesto/hierarchical/inner_calculator_collector.py
@@ -114,6 +114,11 @@ class InnerCalculatorCollector(AmiciCalculator):
 
         self._known_least_squares_safe = False
 
+        #: per condition, the ``(par_sim_slice, par_opt_slice)`` pairs
+        #: mapping the sensitivities onto the optimization parameters.
+        #: ``None`` means they are derived from the parameter mapping.
+        self._index_slices = None
+
     def initialize(self):
         """Initialize."""
         for calculator in self.inner_calculators:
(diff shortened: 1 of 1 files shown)


FFroehlich:
wait for #1754 before reviewing

dweindl (approved):

PR #1753, opened 2026-09-07
Fix silently empty residuals from hierarchical optimization

FFroehlich:
Hierarchical optimization returned an **empty residual vector** in residual mode, without raising. On hierarchical Boehm, `res` had shape `(0,)` for a problem with 48 measurements.

`RelativeAmiciCalculator.__call__` sends residual mode to `calculate_directly`, which computes `FVAL` and `GRAD` from the inner solver and leaves `RES`/`SRES` at the empty arrays `init_return_values` hands it. Nothing ever fills them.

## Fix

Route residual mode to `call_amici_twice`, which is what the PEtab v2 collector already does. Its second simulation runs with the inner parameters at their optimum, so AMICI computes the residuals and their sensitivities itself.

That alone trips the least-squares check in `AmiciCalculator.__call__`:

```
RuntimeError: Cannot use least squares solver with parameter dependent sigma!
```

The optimal inner parameters were written into `x_dct` while the parameter mapping still mapped the simulation parameters to their **ids**, so AMICI kept them in the `plist` and differentiated with respect to them. Those sensitivities were then discarded in `par_index_slices`, which skips ids that are not optimization parameters. Writing the optimal values into the mapping instead is what the comment there already claimed the code did:

```python
# directly writing to parameter mapping ensures that plists do not
# include hierarchically computed parameters
```

`par_index_slices` derives each parameter's position in AMICI's sensitivity output from the cumulative count of id-mapped parameters, mirroring how the `plist` is built, so the shorter `plist` needs no other change. The second simulation also stops computing sensitivities that were only going to be thrown away.

## Hierarchical sigmas are not least-squares safe

Making the `plist` change without saying more would **silence** the check quoted above for exactly the case it should catch. At the analytically optimal sigma, σ\* is the RMS deviation, so Σrᵢ² ≡ n identically:

| Δ on `Epo_degradation_BaF3` | nllh | 0.5·‖res‖² |
|---|---|---|
| 0.00 | 133.571498 | 24.000000000 |
| 0.05 | 133.564625 | 24.000000000 |
| −0.10 | 138.616771 | 24.000000000 |
| 0.25 | 146.459417 | 24.000000000 |

A least-squares optimizer would be minimizing a constant.

The residuals themselves are fine — they are the standardized residuals at the fitted noise level, and are worth having for diagnostics — so they stay available. Only the residual **sensitivities** are refused, which is precisely what a least-squares solver needs and precisely what the existing parameter-dependent-sigma check blocks, for the same reason. Same `RuntimeError`, same `mode == MODE_RES and 1 in sensi_orders` condition. It lives in `call_amici_twice` rather than `__call__` so the PEtab v2 collector, which calls it directly, is covered too.

## What now works

With the sigmas fixed and only scalings and offsets solved analytically:

| | before | after |
|---|---|---|
| `res` | `(0,)` | `(48,)` — one per measurement |
| `sres` | `(0, 6)` | `(48, 6)` |

* `0.5 * res @ res` tracks the negative log-likelihood, differing from it by a constant 114.2337 across parameter values — the sigma term, now genuinely constant.
* `sres.T @ res` matches the function-mode gradient to a relative 3e-10.

Both are asserted by `test_hierarchical_residuals`; `test_hierarchical_residuals_with_inner_sigma` pins `res @ res == n_measurements` and the refusal of the Jacobian.

`test/hierarchical`: 45 passed. The one failure, `test_hierarchical_optimization_pipeline`, is `fides` not being installed in my environment and is unrelated. `test_hierarchical_calculator_and_objective` covers the adjoint path through `call_amici_twice` and is unchanged by the shorter `plist`.

## Note on the Jacobian

`sres` treats the inner parameters as constants at their optimum. That is the standard variable-projection choice: the implicit dependence of the inner optimum on the outer parameters contributes nothing to `J^T r`, which is why the Gauss-Newton gradient matches exactly. It is not the full derivative of `res`, so finite-differencing `res` against `sres` will not agree.

Found while refactoring the collector in #1751, which documents the defect and deliberately preserved it rather than changing behaviour inside a refactor.

### DIFF
--- a/pypesto/hierarchical/relative/calculator.py
+++ b/pypesto/hierarchical/relative/calculator.py
@@ -327,10 +327,18 @@ class RelativeAmiciCalculator(AmiciCalculator):
         Hessian nor residuals are requested. In this case, the objective
         function and gradient are computed directly using the solver methods.
 
-        Only ``FVAL`` and ``GRAD`` are filled in; ``RES`` and ``SRES`` keep
-        the empty arrays from :func:`init_return_values`, so this must not be
-        called in :obj:`MODE_RES`.
+        Only ``FVAL`` and ``GRAD`` are filled in; ``RES`` and ``SRES`` would
+        keep the empty arrays from :func:`init_return_values`, so residual
+        mode raises rather than returning residuals that are silently empty.
         """
+        if mode == MODE_RES:
+            raise ValueError(
+                "`calculate_directly` computes the objective and its "
+                "gradient from the inner solver and produces no residuals. "
+                "Residual mode goes through `call_amici_twice`, where AMICI "
+                "computes them itself."
+            )
+
         dim = len(x_ids)
         # compute optimal inner parameters
         x_dct = copy.deepcopy(x_dct)
--- a/pypesto/hierarchical/relative/calculator.py
+++ b/pypesto/hierarchical/relative/calculator.py
@@ -21,9 +21,12 @@ from ...C import (
     GRAD,
     HESS,
     INNER_PARAMETERS,
+    LIN,
+    MODE_RES,
     RDATAS,
     RES,
     SRES,
+    InnerParameterType,
     ModeType,
 )
 from ...objective.amici.amici_calculator import (
@@ -35,6 +38,7 @@ from ...objective.amici.amici_util import (
     filter_return_dict,
     init_return_values,
 )
+from ..base_problem import scale_back_value_dict
 from .problem import AmiciInnerProblem
 from .solver import AnalyticalInnerSolver, InnerSolver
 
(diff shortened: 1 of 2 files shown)


pypesto/hierarchical/relative/calculator.py
```diff
@@ -272,9 +323,13 @@ def calculate_directly(
...
-        This is possible if the forward method is used, and the Hessian is not
-        requested. In this case, the objective function and gradient are computed
-        directly using the solver methods.
+        This is possible if the forward method is used, and neither the
+        Hessian nor residuals are requested. In this case, the objective
+        function and gradient are computed directly using the solver methods.
+
+        Only ``FVAL`` and ``GRAD`` are filled in; ``RES`` and ``SRES`` keep
+        the empty arrays from :func:`init_return_values`, so this must not be
+        called in :obj:`MODE_RES`.
```
dweindl: Then fail on `mode == MODE_RES`?

dweindl (approved):

